PR #12, opened 2015-03-08
BIGTOP-1734.  support solr in puppet components.

jayunit100:


c0s:
It doesn't apply anymore, it seems.

PR #25, opened 2015-08-05
BIGTOP-1888. Upgrade Flume to 1.6.0

youngwookim:


jayunit100:
Looks like there are some kite changes in here? Maybe accidental?

youngwookim:
Jay, Kite in the Bigtop stack! 

https://issues.apache.org/jira/browse/BIGTOP-1149

jayunit100:
Ah ok so flume is reading a kite version.  That makes sense. +1

youngwookim:
Thanks Jay for reviewing the patch! I'll commit it shortly.

PR #29, opened 2015-08-20
BIGTOP-1971. Support Spark SQL CLI with Apache Hive out of the box

youngwookim:


jayunit100:
Looks good to me if tested.  +1

PR #33, opened 2015-08-25
BIGTOP-1983. Move BigPetStore data generator to bigtop-data-generators

rnowling:
BIGTOP-1983. Move BigPetStore data generator to bigtop-data-generators


jayunit100:
+1 .. i think this is just a pure mv operation.

rnowling:
Don't have GitHub permissions, so I merged manually.

PR #34, opened 2015-08-25
BIGTOP-1984. Extract samplers library from BigPetStore data generator

rnowling:
BIGTOP-1984. Extract samplers library from BigPetStore data generator


rnowling:
Closing due to conflicts.  Will resubmit as patch.

PR #35, opened 2015-08-25
BIGTOP-1985. Extract name generator from BigPetStore data generator

rnowling:
BIGTOP-1985. Extract name generator from BigPetStore data generator


rnowling:
Closing due to conflicts.  Will resubmit as patch.

PR #38, opened 2015-09-03
BIGTOP-2003. Bigtop puppet fails to deploy on Ubuntu due to a hiera 1…

evans-ye:
BIGTOP-2003. Bigtop puppet fails to deploy on Ubuntu due to a hiera 1…


jayunit100:
+1

PR #40, opened 2015-09-21
add src param so rpm build works

aelent:
gradle bigtop-utils-rpm task fails with:

Copy  to /opt/bigtop/build/bigtop-utils/tar/bigtop-utils-1.1.0-SNAPSHOT.tar.gz
:bigtop-utils-tar FAILED

* Where:
Script '/opt/bigtop/packages.gradle' line: 250


youngwookim:
@aelent Please point out corresponding JIRA issue to reviewers. If the JIRA does not exist, file a appropriate one first. https://issues.apache.org/jira/browse/BIGTOP

Thanks!

aelent:
@youngwookim Created https://issues.apache.org/jira/browse/BIGTOP-2068

Thanks

PR #51, opened 2015-11-14
BIGTOP-2143. [Puppet] Automatically generate default repo

evans-ye:


evans-ye:
Sure. It works for every supported OS in Bigtop except OpenSuSE. ;)

PR #63, opened 2015-12-05
BIGTOP-2106. tez does not build on opensuse

evans-ye:


evans-ye:
Sorry. Wrong JIRA number.

PR #66, opened 2015-12-16
BIGTOP-2179: Add packaging scripts for Apache Tajo

yeongeon:


jayunit100:
This looks promising.
In general puppet deployment and smoke tests are also good to have.  

Are you building those as well?   You can usually just copy examples from something like pig or spark puppet recipes in our repo to get started .

Smoke tests are also quite easy just a groovy file which you can dump in smoke-tests.

The vagrant recipes can then be used to build and test from scratch

yeongeon:
I just have tested rpm/deb as a standalone on centos/ubuntu. 
The smoke-tests and the deploy will be added step by step.

youngwookim:
The patch does not apply. Please rebase it.

bigtop-packages/src/common/tajo/do-component-build
```diff
@@ -0,0 +1,28 @@
...
+
+set -ex
+
+# recognize for TAJO_VERSION, HADOOP_VERSION
+. `dirname $0`/bigtop.bom
+
+rm -rf build
+mkdir build
+
+mvn clean package -DskipTests -Pdist -Dtar -Dhadoop.version=$HADOOP_VERSION "$@"
```
youngwookim: I believe, you should provide the appropriate hive.version for Hive integration

youngwookim:
'compat' file is missing for deb? See https://cwiki.apache.org/confluence/display/BIGTOP/Bigtop+Packaging

Like this, https://github.com/apache/bigtop/blob/master/bigtop-packages/src/deb/hive/compat

bigtop-packages/src/common/tajo/do-component-build
yeongeon: @youngwookim Your mention seems to be related with TAJO-1442 which remove hcatalog dependency. I think we first should make a new issue and then deal with it in this.

yeongeon:
Thanks guys. I have pushed three fixed.

bigtop-packages/src/common/tajo/do-component-build
yeongeon: @youngwookim I had some problems while rebase but solved it using merge now.

yeongeon:
Let me know if this PR have any issues more for merging. 
(Also, PR/67 was related with this.)

yeongeon:
I think this PR was updated as you guys are mentioned.
Also, patch file was added on the issue.

PR #71, opened 2016-01-17
BIGTOP-2254. Replace the Docker orchestration tool from Vagrant to Do…

evans-ye:
…cker Compose


c0s:
I will start looking at it in a day or two as soon as I am done with a couple of things in the woodworks right now.

evans-ye:
Thanks for the review!

Yes you should have docker-compose installed first. I plan to add support to download and install it via docker-hadoop as well. After all it just need one line to get it done. :)

config/hieradata will be generated in generate-config function. In the future we should allow users to specify customized config instead of always generate new ones.

The new directory is a very good idea. I was digging to deep so that I can't see the name is so misleading... With that we can just merge it into master w/o breaking anything.

In fact, I'm thinking that maybe we can put a top level directory for provisioners. That things currently sit inside bigtop-deploy is hard to find. What do you think?

BTW, any shell script styling comment is welcome.

PR #76, opened 2016-01-27
BIGTOP-2180: Add smoke-test scripts using gradle for Apache Tajo.

yeongeon:
It's a new PR that is separated from BIGTOP-2179.


c0s:
This seems to be fine. Will commit this to the master in a bit. Thanks!

PR #77, opened 2016-01-29
BIGTOP-2285. add qfs rpm and debian packaging code

fsareshwala:
This change adds code to bigtop to package qfs for debian and rpm systems. The
packages produced are specific to their functions, e.g. a separate package for
the metaserver, chunkserver, webui, client tools, development headers and
libraries, etc. This way, users can select which packages they wish to install
on which system in their infrastructure.

In order to build qfs and its respective packages, various dependencies were
added to the toolchain puppet tasks (e.g. boost-devel). These will be installed
along with the other dependencies that get installed during the toolchain gradle
task. In doing so, users should be presented with a seamless setup process to
get qfs packages building.

I have tested to make sure that packaging is correct by using the qfs-rpm and
qfs-deb grandle tasks to generate the respective rpm and debian packages. I then
manually inspected package contents to ensure that files were being installed
into the correct location.


fsareshwala:
Closing this pull request out so that I can fix the source branch.

fsareshwala:
New packaging pull request: https://github.com/apache/bigtop/pull/83

PR #78, opened 2016-02-02
BIGTOP-2302. Use apt instead of yum in setup-env-debian.sh

fsareshwala:


fsareshwala:
This was committed.

https://issues.apache.org/jira/browse/BIGTOP-2302

PR #79, opened 2016-02-02
BIGTOP-2303. Fix the indentation in docker-hadoop.sh

fsareshwala:


fsareshwala:
This was committed.

https://issues.apache.org/jira/browse/BIGTOP-2303

PR #83, opened 2016-02-09
BIGTOP-2285. add qfs rpm and debian packaging code

fsareshwala:
This change adds code to bigtop to package qfs for debian and rpm systems. The
packages produced are specific to their functions, e.g. a separate package for
the metaserver, chunkserver, webui, client tools, development headers and
libraries, etc. This way, users can select which packages they wish to install
on which system in their infrastructure. One caveat worth mentioning is that the
packages themselves don't contain configuration files for the services in
question. The configuration files are provided by the puppet deploy scripts that
come with the bigtop repository. Those will be submitted in a separate change.

In order to build qfs and its respective packages, various dependencies were
added to the toolchain puppet tasks (e.g. boost-devel). These will be installed
along with the other dependencies that get installed during the toolchain gradle
task. In doing so, users should be presented with a seamless setup process to
get qfs packages building.

I have tested to make sure that packaging is correct by using the qfs-rpm and
qfs-deb grandle tasks to generate the respective rpm and debian packages. I then
manually inspected package contents to ensure that files were being installed
into the correct location.


bigtop-packages/src/common/qfs/do-component-build
```diff
@@ -0,0 +1,133 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# For debugging:
+# set -ex
+
+# Autodetect JAVA_HOME if not defined
+source /usr/lib/bigtop-utils/bigtop-detect-javahome
```
youngwookim: I believe JAVA_HOME detection is not necessary at build stage because we have bigtop toolchain. So, please remove this bits.

bigtop-packages/src/deb/qfs/rules
```diff
@@ -0,0 +1,81 @@
...
+
+# Uncomment this to turn on verbose mode.
+export DH_VERBOSE=1
+
+# This has to be exported to make some magic below work.
+export DH_OPTIONS
+
+# Build configuration -- override with environment variables
+SHELL := /bin/bash
+HADOOP_VERSION ?= 1.0.4
```
youngwookim: Bigtop defines HADOOP_VERSION in bom so, just pass the variable ${HADOOP_VERSION} or ${HADOOP_BASE_VERSION}. ref. https://github.com/apache/bigtop/blob/master/bigtop-packages/src/deb/zookeeper/rules#L39

bigtop-packages/src/rpm/qfs/SPECS/qfs.spec
```diff
@@ -0,0 +1,360 @@
...
+%define lib_dir /usr/lib/qfs
+%define etc_dir %{_sysconfdir}/qfs
+%define include_dir /usr/include
+%define data_dir /usr/share/qfs
+%define var_dir /var
+%define qfs_services chunkserver metaserver webui
+
+%global HADOOP_VERSION %{hadoop_version}
+
+%if %{?!HADOOP_VERSION:1}0
```
youngwookim: Would be good to have default Hadoop version from bigtop bom. Bigtop does not support Hadoop 1.x at this point.

bigtop-packages/src/common/qfs/do-component-build
fsareshwala: If I remove this, then JAVA_HOME isn't passed on through the makefile for qfs and we can't locate jni and build the jni bindings. Should this not be the case?

bigtop-packages/src/deb/qfs/rules
fsareshwala: We do pass it through in the packages.gradle. This is just a default in case it isn't passed through for some reason.

bigtop-packages/src/rpm/qfs/SPECS/qfs.spec
fsareshwala: We do pass it through from the bigtop bom. See my changes to packages.gradle to see how this is handled and flows through.

c0s: Hmm, my bad then. I though the up-level build.gradle will enforce it. I will fix it later, for now just disregard this comment. Thanks

fsareshwala:
Olaf committed this code into bigtop: https://issues.apache.org/jira/browse/BIGTOP-2285

I am going to close this pull request out.

PR #86, opened 2016-02-12
BIGTOP-2317. Add smoke tests for QFS

fsareshwala:
This change adds some simple testing for qfs running under hadoop. The bigtop
repository already has smoke tests for the hadoop mapreduce framework. It simply
runs the examples through the hadoop infrastructure and ensures that the output
is valid and expected files within the filesystem are present. This change
modifies the bigtop code to support running the same examples but using qfs as
the backend filesystem. This is done through dynamically supplying the hadoop
command to use. In the hadoop hdfs mapreduce scenario, we use the standard
hadoop command that uses hdfs as the backend filesystem. In the hadoop qfs
mapreduce scenario, we use the hadoop-qfs command that uses qfs as the backend
filesystem. The qfs tests simply reference the mapreduce tests and tell it to
use the hadoop-qfs command instead of the hadoop command.

I have tested this works by running the qfs smoke tests using the BigTop
provisioner under docker. The build and tests pass without issue.

### DIFF
--- a/bigtop-tests/smoke-tests/qfs/build.gradle
+++ b/bigtop-tests/smoke-tests/qfs/build.gradle
@@ -1,0 +1,45 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+def tests_to_include() {
+  def smoke_tests = System.getenv('SMOKE_TESTS')
+  if (smoke_tests == null || !smoke_tests.contains("qfs")) {
+    return [];
+  }
+
+  return [
+    "TestHadoopExamples.groovy"
+  ];
+}
+
+sourceSets {
+  test {
+    groovy {
+      srcDirs = [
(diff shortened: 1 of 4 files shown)


bigtop-tests/smoke-tests/qfs/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+      ]
+      exclude {
+        FileTreeElement elem -> (doExclude(elem.getName()))
+      }
+    }
+  }
+}
+
+test.doFirst {
+  environment("HADOOP_COMMAND", "hadoop-qfs")
```
c0s: One last question before I commit this: is there any way skip the test if the QFS isn't configured in the cluster? Perhaps, checking for an executable in a well-known location or something like that? That will let us to avoid false-negatives.

fsareshwala: That's a good suggestion, @c0s. I've updated the code to check the environment variable `SMOKE_TESTS` before running the qfs smoke tests.

c0s: This isn't what I meant. Someone can set the SMOKE_TESTS variable to qfs as he is intent to run such tests. However, the cluster might not be configured with qfs, i.e. the packages aren't installed, etc. This will lead to a false negative. The reason I am being overly cautious is because in case of HDFS it is always there, so there's no need to specifically check for its presence. QFS being an optional alternative won't be implicitly required by other downstream components like YARN, hence it configuration might be omitted. See where I am going with this?

fsareshwala: I worry that checking for a binary existing might hide other problems with installing the binary itself. That's why I used the environment variable. One option would be to pass in the list of components to the `smoke-tests.sh` script from `docker-hadoop.sh` and I can search through that instead. What do you think about that? If that's not a good solution, can you give me a suggestion on how I might go about doing what you are asking for?

c0s: I guess you right. You know - let's scratch my initial suggestion and leave it as it was. After all other components don't do any checks like that and so far we haven't blown anything ;) Thanks!

fsareshwala: Sure thing @c0s. I still think it's valuable to check for the `SMOKE_TESTS` environment variable inside the qfs `build.gradle`. I've left that in. I think if you are ready for the merge, this is ready to go. Let me know if there's anything else you want me to change.

fsareshwala:
This change was committed to the bigtop repository already. Closing out the pull request.

PR #92, opened 2016-02-22
BIGTOP-2314 Added deb and rpm package of Apache Apex to bigtop.

chinmaykolhatkar:
This pull request add deb and rpm packaging of apex as a part of bigtop.


bigtop-packages/src/common/apex/do-component-build
```diff
@@ -0,0 +1,81 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+. `dirname $0`/bigtop.bom
+
+HADOOP_VERSION=2.2.0
```
youngwookim: Is it intended? if you need a version of the dependency, just use the HADOOP_VERSION from bigtop.bom like this https://github.com/apache/bigtop/blob/master/bigtop-packages/src/common/hive/do-component-build#L24

bigtop-packages/src/common/apex/install_apex.sh
```diff
@@ -0,0 +1,111 @@
...
+install -d -m 0755 $PREFIX/$LIB_DIR/bin
+install -d -m 0755 $PREFIX/$LIB_DIR/lib
+install -d -m 0755 $PREFIX/$BIN_DIR
+install -d -m 0755 $PREFIX/$DOC_DIR
+install -d -m 0755 $PREFIX/$MAN_DIR
+
+# Install dtcli
+cp engine/src/main/scripts/dtcli $PREFIX/$LIB_DIR/bin/apex
+chmod 755 $PREFIX/$LIB_DIR/bin/apex
+# Create symlinks
```
youngwookim: Not sure why it is needed. if we install $PREFIX/$LIB_DIR/bin/apex then, it would be nice to have wrappers for that like these: https://github.com/apache/bigtop/blob/master/bigtop-packages/src/common/hadoop/install_hadoop.sh#L167 and https://github.com/apache/bigtop/blob/master/bigtop-packages/src/common/hive/install_hive.sh#L135

bigtop-packages/src/common/apex/do-component-build
chinmaykolhatkar: I removed this... Though we need this to exclude the dependencies of hadoop 2.2 while copying min set of runtime libraries.

bigtop-packages/src/common/apex/install_apex.sh
chinmaykolhatkar: Modified as per the suggestion.

chinmaykolhatkar:
Closing as this PR is merged.

PR #98, opened 2016-03-23
BIGTOP-2364. QFS packaging should use HADOOP_VERSION environment variable

fsareshwala:


bigtop-tests/smoke-tests/pig/passwd
```diff
@@ -0,0 +1,33 @@
+root:x:0:0:root:/root:/bin/bash
```
c0s: I don't think this file should be a part of the PR

fsareshwala: Oops, you're right. I've removed it now.

fsareshwala:
This has been merged.

PR #99, opened 2016-03-29
BIGTOP-2367. QFS packages do not handle /var/run/qfs properly

fsareshwala:


fsareshwala:
This has been merged.

PR #108, opened 2016-05-10
BIGTOP-2435 Add Juju charms for hadoop component

kwmonroe:
Initial PR to include Juju charms and bundles in the Bigtop repo.  See the "Bigtop and Juju" note to the dev list on May 9 for more detail and discussion.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,5 +1,6 @@
 Apache Bigtop
 Copyright 2014, The Apache Software Foundation
+Portions Copyright 2015-2016 Canonical Ltd.
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).
--- a/bigtop-packages/src/charm/README.md
+++ b/bigtop-packages/src/charm/README.md
@@ -1,0 +1,75 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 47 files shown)


c0s:
Thanks for your plans of contributing the deployment code back into Apache Bigtop! Could you please open a JIRA ticket in https://issues.apache.org/jira/browse/BIGTOP (I will add you as a contributor so the ticket can be assigned to you) and make the commit comment in the form of
  `BIGTOP-####. JIRA synopsis`
It is minor stuff, but helps us to keep the patches uniformed. I will start reviewing your patch tomorrow, so expect more comments  soon ;)

c0s:
Let's see if now it gets linked to the JIRA

c0s:
I have reviewed the [dev@ thread discussion](https://lists.apache.org/thread.html/Z2uoq25segpbsm6) on this issue and while I agree with Olaf, that commit is a bit on the large side, it seems to be doing all the correct things. I believe the directory layout also makes sense. And the patch is passes the RAT too.

It might makes sense to add weather-report part separately, but after all it is just one tiny file and a README in there, so it doesn't make that much difference. 

The only real comment I can make is about 2 vs 4 spaces of the indentation as we are using the former. However, the majority of the new code is Python, so I am not really sure if 2-spaces indents look good (or even are acceptable) in that language.

Hence, I am all +1 on getting this committed, unless there are hard objections from the rest of the community?

johnsca:
4 spaces is standard for Python code (it is strongly recommended in PEP-8, the official code style guideline for Python), and Python linters will complain if you use something else without adding explicit overrides.  If you don't feel strongly about the consistency, I'd recommend keeping 4 spaces for the Python code.

kwmonroe:
In an effort to reduce the size of this request, I have pulled out the Juju bundles (bigtop-deploy/juju) and Cloud Weather Report (bigtop-tests/cloud-weather-report).  This just leaves the core set of hadoop charms up for review.

c0s:
Thanks for getting out of the way to address the community feedback, guys.
I believe in this shape, the patch stands to be as small as it possibly can be. I don't really have anything to say about it, so I will just go ahead and commit it. Thanks!

c0s:
Committed and pushed to the Apache tree. Please feel free to close this PR

kwmonroe:
Thanks c0s! https://github.com/apache/bigtop/commit/d639645e7726577e5898f5da3722c6be65b2c9bd has the goods.  Closing this out.

PR #113, opened 2016-06-03
BIGTOP-2469: Add cloud-weather-report test plan

johnsca:
We've split the original request to enable Bigtop/Juju support into a couple smaller subtasks. This one is for delivering the cloud-weather-report test plan for running the Juju bundle against multiple clouds and verifying the results.

### DIFF
--- a/bigtop-tests/cloud-weather-report/README.md
+++ b/bigtop-tests/cloud-weather-report/README.md
@@ -1,0 +1,43 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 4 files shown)


c0s:
This is in conflict now with the master branch

c0s:
Also, please make the name of the PR matches the JIRA title exactly, otherwise the latter won't get updated

johnsca:
Conflict resolved.  It was the addition of a couple of lines to `pom.xml` and `build.gradle` for the RAT fixes.  They were both added lines, but git didn't have quite enough common context to resolve it automatically.

c0s:
+1 all seems to be good. 
Would be nice to start penciling how this can be added to the Bigtop CI, perhaps?

PR #114, opened 2016-06-06
BIGTOP-2472: zeppelin puppet module needs to expose ZEPPELIN_WAR_TEMPDIR

kwmonroe:


c0s:
Interesting, I have fixed this back in Jan as 
  a97479a9d37d7c94f5128f36df2bad8b3676d2c2
Are you saying this change is gone?

kwmonroe:
Negative c0s, your commit is in there, but that only appears to affect the zeppelin deb.  Your change to zeppelin-env.sh is not present in the puppet:

https://github.com/apache/bigtop/blob/master/bigtop-deploy/puppet/modules/zeppelin/templates/zeppelin-env.sh

So I'm guessing puppet installs the zeppelin deb (which does the right thing), but then it goes off and overwrites the /etc/zeppelin/conf/zeppelin-env.sh with its template from ./bigtop-deploy/puppet/modules/zeppelin/templates.

This PR is to ensure your changes from:

bigtop-packages/src/common/zeppelin/zeppelin-env.sh

make it into:

bigtop-deploy/puppet/modules/zeppelin/templates/zeppelin-env.sh

c0s:
oh yeah, makes sense! +1 will commit it in a bit. Thanks!

kwmonroe:
This came in with:

https://github.com/apache/bigtop/commit/9299713a2c59b7c605ec8a3d118244562485bf59

Closing out PR.

PR #116, opened 2016-06-08
BIGTOP-2467 Zookeeper Puppet script does not setup Zookeeper node ids correctly

petevg:
Zookeeper wants its peer ids to be consistent across nodes, but the
current puppet script auto-generates ids for nodes, based on their index
in a list. This breaks Zookeeper quorum.

We fix the issue by requiring that puppet take a list comprising
explicit ids and ip addresses, rather than just a list of ip addresses.

### DIFF
--- a/bigtop-deploy/puppet/hieradata/bigtop/cluster.yaml
+++ b/bigtop-deploy/puppet/hieradata/bigtop/cluster.yaml
@@ -147,7 +147,7 @@ qfs::common::chunkserver_client_port: "22000"
 
 hadoop_zookeeper::server::myid: "0"
 hadoop_zookeeper::server::ensemble:
-  - "%{hiera('bigtop::hadoop_head_node')}:2888:3888"
+  - ["0", "%{hiera('bigtop::hadoop_head_node')}:2888:3888"]
 hadoop_zookeeper::server::kerberos_realm: "%{hiera('kerberos::site::realm')}"
 
 # those are only here because they were present as extlookup keys previously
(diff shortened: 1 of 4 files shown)


c0s:
That seems about right, but I don't know Ruby so well to tell if this is the best practices. Can anyone else chime in?

rich-reactor8:
The Ruby looks fine

oflebbe:
Please wait. This changes the hiera config of ensemble. I like to doublecheck with my puppet conf

petevg:
@oflebbe I read ensemble as only applying to Zookeeper. If it's used elsewhere, then it will definitely break other things that use it.

If that's the case, should we just create a zookeeper_ensemble entry in the puppet script, and have Zookeeper read from that?

petevg:
Updated this PR to include some lines of code that referenced the old way of configuring the zookeeper ensemble, which I missed on my first pass.

The only other references to "ensemble" in Bigtop are in the Solr modules, and it does not look like solr configures Zookeeper directly, or writes to the Zookeeper puppet scripts. I believe that it is okay to leave Solr as it is.

@oflebbe How do things look from your end?

petevg:
Updated PR to add backwards compatibility for the old way of formatting ensemble. @oflebbe does that address your concerns?

johnsca:
This was merged as 87e7d01 so this PR should be closed.

petevg:
Closing, as this has been merged (outside of github)

PR #120, opened 2016-06-10
BIGTOP-2480: Fix occasional timeout on hadoop-processing Juju bundle

johnsca:
During test runs, we are sometimes seeing the bundle test timeout during the initial deploy on slow clouds. GCE in particular is slow to deploy (but is faster once deployed, go figure).

There is a feature of the bundletester tool which can be used to increase the deployment timeout.

### DIFF
--- a/bigtop-deploy/juju/hadoop-processing/bundle.yaml
+++ b/bigtop-deploy/juju/hadoop-processing/bundle.yaml
@@ -1,6 +1,6 @@
 services:
   openjdk:
-    charm: cs:trusty/openjdk-1
+    charm: cs:trusty/openjdk-4
     annotations:
       gui-x: "500"
       gui-y: "400"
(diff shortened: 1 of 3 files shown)


kwmonroe:
@johnsca  is this still valid?  It's not clear to me how `00-deploy` is doing anything different than `01-deploy.py` is doing with its `d.setup` and `wait_for_messages` calls:

https://github.com/apache/bigtop/blob/master/bigtop-deploy/juju/hadoop-processing/tests/01-bundle.py#L37

johnsca:
@kwmonroe Yes, the motivation behind this change is still valid but there is now another, cleaner way to get the same result.

The difference between `00-deploy` and `01-bundle` is that the former is being used as a pre-deploy step ([line 2](https://github.com/apache/bigtop/pull/120/files#diff-707de698394ac34e432f2dec065a7916R2) in `tests.yaml`).  This lets us override the timeout of the initial deploy and at the time of this PR's submission was the only way to do so.  juju-solutions/bundletester#50 has since been released, however, so we could replace `00-deploy` and the `bundle_deploy` option with the `deployment_timeout` option.

I can update this PR, or close and submit a new one.

Oh, and I also think its worthwhile to rev the charms to the most recent versions, yes?

kwmonroe:
Let's kill this JIRA/PR @johnsca.. I've added `deployment_timeout` to the bundle refresh jira, which already bumped charm revnos:

https://issues.apache.org/jira/browse/BIGTOP-2550

Code:

https://github.com/juju-solutions/bigtop/blob/feature/BIGTOP-2550/bundle-to-xenial/bigtop-deploy/juju/hadoop-processing/tests/tests.yaml#L2

```diff
--- a/bigtop-deploy/juju/hadoop-processing/tests/tests.yaml
+++ b/bigtop-deploy/juju/hadoop-processing/tests/tests.yaml
@@ -1,4 +1,5 @@
 reset: false
+deployment_timeout: 3600
 packages:
   - amulet
   - python3-yaml
```

PR #122, opened 2016-06-16
BIGTOP-2478: Zookeeper does not have any smoke tests

petevg:
These tests are very basic for now. We just run "zkServer.sh status" and
verify that the output is reasonable.

More complex tests are blocked by either a) needing multiple Zookeeper
nodes (doesn't really make sense for a smoke test), or b) needing to
have Zookeeper REST spun up, which Bigtop does not currently do.

### DIFF
--- a/bigtop-tests/smoke-tests/zookeeper/TestZookeeper.groovy
+++ b/bigtop-tests/smoke-tests/zookeeper/TestZookeeper.groovy
@@ -27,7 +27,7 @@ import org.apache.commons.logging.Log
 
 import static org.apache.bigtop.itest.LogErrorsUtils.logError
 
-class TestSpark {
+class TestZookeeper {
   static private Log LOG = LogFactory.getLog(Object.class)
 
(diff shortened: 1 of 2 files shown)


bigtop-tests/smoke-tests/zookeeper/TestZookeeper.groovy
```diff
@@ -0,0 +1,66 @@
...
+
+import org.apache.bigtop.itest.shell.Shell
+import static org.junit.Assert.assertTrue
+import org.junit.Test
+import org.apache.commons.logging.LogFactory
+import org.apache.commons.logging.Log
+
+import static org.apache.bigtop.itest.LogErrorsUtils.logError
+
+class TestSpark {
```
rvs: You probably don't want it to be called TestSpar, do you?

rvs:
This looks good as a very, very basic ZK test. See the small nit I had tho.

petevg:
@rvs: Whoops. Fixed the name. Thank you :-)

PR #123, opened 2016-06-17
BIGTOP-2483: Add Mahout Charm

ktsakalozos:


kwmonroe:
Hey Kostas, this doesn't feel right to me.  I've added a hadoop-client -> openjdk relation, followed by a hadoop-client -> mahout relation, but mahout still reports "Waiting for Java".  It's only after I explicitly add an openjdk -> mahout relation that the charm deploys.

Is this expected behavior?  If it is, I don't like it!  The principal (hadoop-client) already has openjdk related to it, so i shouldn't have to add java -> mahout.  Since you surely know that already, I'm probably doing something wrong.. Either way, let's figure it out before committing upstream.

bigtop-packages/src/charm/mahout/layer-mahout/actions/smoke-test
```diff
@@ -0,0 +1,20 @@
...
+hdfs dfs -put resources/users.txt /tmp/input/
+
+if hdfs dfs -stat temp &> /dev/null; then
+    hdfs dfs -rm -r -skipTrash temp || true
+fi
+if hdfs dfs -stat output &> /dev/null; then
+    hdfs dfs -rm -r -skipTrash output || true
+fi
+
+hadoop jar /usr/lib/mahout/mahout-mr-0.11.1-job.jar org.apache.mahout.cf.taste.hadoop.item.RecommenderJob -Dmapred.input.dir=/tmp/input/links-converted.txt -Dmapred.output.dir=output --usersFile /tmp/input/users.txt --booleanData -s SIMILARITY_LOGLIKELIHOOD
```
smarthi: we now have mahout 0.12.2, u may wanna go with that

ktsakalozos: As far as I can tell Bigtop does not yet package Mahout 0.12.2.

So, I updated the action script and used a wildcard instead of a fixed Mahout version. This was tested against a manual deployment of 0.12.2 Mahout. Unless there are significant changes in either the packaging of Mahout or the internals of the jar, the test should be valid for both the current and future Mahout versions.

Thank you for spotting this and helping improve the quality of the code.

ktsakalozos:
Hi Kevin, as per our yesterday's discussion we agreed to favour explicit relations even if they seem redundant. The user is indeed required to add a relation between openjdk and mahout. If we are to remove that requirement charm authors that want their charms to relate to mahout would also need to relate to java and make sure java is available before mahout deployment is triggered.

petevg:
Tests complete successfully, with the latest juju 2.0 beta, against AWS. I am +1 on this.

PR #125, opened 2016-06-23
BIGTOP-2481: Add HBase Charm

ktsakalozos:


petevg:
I am +1 on this. Tests all pass when run with bundletester (I had to drop in a small fix to make zookeeper work with both old and new zookeeper), including smoke tests, which run upstream Bigtop smoke tests.

PR #129, opened 2016-06-27
BIGTOP-2476 Add Zookeeper Charm

petevg:
Based on the vanilla apache-zookeeper charm, but utlizes Bigtop to
reduce complexity in the charm itself.

This covers all the changes that we discussed, including requiring a manual rolling restart after adding or removing zk peers (per http://www.benhallbenhall.com/2011/07/rolling-restart-in-apache-zookeeper-to-dynamically-add-servers-to-the-ensemble/)


kwmonroe:
LGTM, verified deployment and smoke tests on azure and aws.

PR #130, opened 2016-06-29
Fix webapps/ path error

lucasa:
Set the correct path to the webapps/ folder.


rvs:
Hi @lucasa first of all, thanks a million for the patch. Also, I think this has to be fixed on the build side rather than on the deployment side. In fact, it looks like a few things need to be fixed on the build side for solr to actually work properly. The current build just doesn't seem to cut it.

It would be great if you could file an Apache JIRA for that https://issues.apache.org/jira/browse/BIGTOP/

lucasa:
JIRA issue opened: https://issues.apache.org/jira/browse/BIGTOP-2509

PR #131, opened 2016-07-07
BIGTOP-2487: Alluxio /var/log/alluxio is not created

ktsakalozos:


ktsakalozos:
Closing the PR since it got merged.

PR #132, opened 2016-07-13
BIGTOP-2492 Split flink debian packaging

dmvk:


bigtop-packages/src/deb/flink/control
```diff
@@ -27,3 +27,13 @@ Depends: bigtop-utils (>= 0.7), adduser
 Description: Flink’s core is a streaming dataflow engine that provides data
              distribution, communication, and fault tolerance for distributed
              computations over data streams.
+
+Package: flink-jobmanager
+Architecture: all
+Depends: flink
```
rvs: you need 
    Depends: flink (= ${source:Version})
here instead to make sure that there's 1-1 correspondence version-wise

bigtop-packages/src/deb/flink/control
```diff
@@ -27,3 +27,13 @@ Depends: bigtop-utils (>= 0.7), adduser
...
+Description: Flink’s jobmanager.
+
+Package: flink-taskmanager
+Architecture: all
+Depends: flink
```
rvs: See above.

bigtop-packages/src/deb/flink/flink-jobmanager.postinst
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+case "$1" in
+    configure)
```
rvs: This is a good place to make sure you update the alternatives files for the /etc/flink/conf if needed

bigtop-packages/src/rpm/flink/SPECS/flink.spec
```diff
@@ -72,19 +81,38 @@ Some of the key features of Apache Flink includes.
     * Fault-tolerance via Lightweight Distributed Snapshots
     * Hadoop-native YARN & HDFS implementation
 
-# Additions for master-worker configuration #
+%package jobmanager
+Summary: Provides the Apache Flink Job Manager service.
```
rvs: Its better to keep the description exactly the same between Debian and RPM packages

bigtop-packages/src/rpm/flink/SPECS/flink.spec
```diff
@@ -106,29 +134,17 @@ bash $RPM_SOURCE_DIR/do-component-build
...
-done
-
 %pre
 getent group flink >/dev/null || groupadd -r flink
 getent passwd flink >/dev/null || useradd -c "Flink" -s /sbin/nologin -g flink -r -d %{lib_flink} flink 2> /dev/null || :
 
-
 %post
 %{alternatives_cmd} --install %{config_flink} %{flink_name}-conf %{config_flink}.dist 30
 systemctl daemon-reload
```
rvs: I don't think this belongs here: systemctl is only available on systemd-based Linux distros, Bigtop currently supports init.d

bigtop-packages/src/rpm/flink/SPECS/flink.spec
```diff
@@ -53,6 +54,14 @@ Source6: bigtop.bom
 Requires: bigtop-utils >= 0.7
 Requires(preun): /sbin/service
 
+%if  %{?suse_version:1}0
```
rvs: This whole block needs to be removed. Once you add two of the subpackages there's no reason to have main package required these

PR #133, opened 2016-07-14
HT 466

amit-altiscale:
Pick the correct file to diff against.


rvs:
@amit-altiscale this is a huge diff, can you please let us know (pref. via JIRA) what is it you're trying to achieve?

amit-altiscale:
Hi Roman,
I'm sorry that request was sent out on error. My apologies.
regards
--amit

PR #136, opened 2016-08-04
BIGTOP-2510: Add role in puppet scripts to deploy journal node of Hadoop HA

ktsakalozos:


petevg:
I'm still not 100% happy with this approach, but this is the correct thing to do if we want to do it. +1 from me.

PR #137, opened 2016-08-10
BIGTOP-2516: Add Zeppelin Charm

ktsakalozos:


bigtop-packages/src/charm/zeppelin/layer-zeppelin/LICENSE
```diff
@@ -0,0 +1,177 @@
+
```
c0s: You don't need to include LICENSE file to each charm - it is already done at the top of the project.

bigtop-packages/src/charm/zeppelin/layer-zeppelin/reactive/zeppelin.py
```diff
@@ -0,0 +1,158 @@
...
+        set_state('zeppelin.spark.configured')
+        update_status()
+
+
+@when('zeppelin.installed', 'zeppelin.spark.configured')
+@when_not('spark.ready')
+def unconfigure_spark():
+    hookenv.status_set('maintenance', 'removing spark relation')
+    zeppelin = Zeppelin()
+    zeppelin.configure_spark(None)
```
kwmonroe: I think we should check for `hadoop.ready` here, and if that's true, use `yarn-client` as the spark master url instead of `None`.  I'm pretty sure zeppelin can run it's internal spark interpreter in yarn-client mode without any external spark master.

FWIW, bigtop defines `spark_master_url` as `yarn-client` by default:
https://github.com/apache/bigtop/blob/master/bigtop-deploy/puppet/modules/zeppelin/manifests/init.pp#L25

bigtop-packages/src/charm/zeppelin/layer-zeppelin/tests/01-zeppelin-smoke.py
```diff
@@ -0,0 +1,66 @@
...
+import unittest
+import amulet
+
+
+class TestDeploy(unittest.TestCase):
+    """
+    Deployment and smoke test for the Apache Bigtop Zeppelin service.
+    """
+    @classmethod
+    def setUpClass(cls):
```
kwmonroe: We can simplify this class by removing all references to 'openjdk' and 'java' below.  Do it!

bigtop-packages/src/charm/zeppelin/layer-zeppelin/reactive/zeppelin.py
johnsca: If Spark isn't related to Zeppelin, there's a fairly good chance that it's not related to Hadoop, either, and thus it would fail in `yarn-client` mode.

Perhaps a better solution would be to remove the override entirely and just let Bigtop fall back to its default naturally.

johnsca: Since the overrides are additive, removing one just leaves it as its current value, so we'll need to explicitly change it back to `yarn-client`.

PR #142, opened 2016-08-27
BIGTOP-2521: Problem with deprecated methods in puppet code with puppet option parser=future

kirill-morozov-epam:


c0s:
The changes look good, but I feel they will break the BIGTOP-2524 patch and it'd needed to get rebased. I will be pushing this one shortly unless I hear otherwise.

c0s:
Committed, please close the PR

PR #145, opened 2016-09-05
BIGTOP-2529. Vagrant Provisioner failed to start up daemons

evans-ye:


c0s:
Seems to be legit (without testing).

PR #146, opened 2016-09-21
Update Solr to 4.10.4

akuchling:
I tried bumping the version of Solr to the last 4.10 release and running the 'gradlew' build script to build RPM packages.  That build process worked fine, and the resulting solr-4.10.4 packages at least install on our CentOS machines and start up as expected.

I don't know if my employer will decide to try this with 5.5.x or 6.x.  I expect that version bump would be more complicated because of Solr's switch from using Tomcat to Jetty, but haven't tried it yet.  For those versions, we might just install Solr directly from a .tgz without using Bigtop's re-packaging.

### DIFF
--- a/bigtop.bom
+++ b/bigtop.bom
@@ -264,7 +264,7 @@ bigtop {
     'solr' {
       name    = 'solr'
       relNotes = 'Apache Solr'
-      version { base = '4.9.0'; pkg = base; release = 1 }
+      version { base = '4.10.4'; pkg = base; release = 1 }
       tarball { destination = "$name-${version.base}-src.tgz"
                 source      = destination }
       url     { download_path = "lucene/$name/${version.base}"


bigtop-deploy/vm/vagrant-puppet-vm/README.md
```diff
@@ -56,13 +56,18 @@ num_instances: 5
 
 5) Test on local built packages is available by:
 
-first, build up local yum repo
```
c0s: The sequence of the YUM build has changed indeed after some recent refactorings. Now it is
- ./gradlew componentname-pkg
- ./gradlew yum|apt

akuchling: Thanks!  Done -- I removed the README changes and filed a Jira issue for the version bump, BIGTOP-2538.

c0s:
it has been committed - you can close the PR now.

PR #148, opened 2016-10-07
BIGTOP-2548: Refresh charms for Juju 2.0 and Xenial

kwmonroe:
- update layer repo to upstream bigtop location
- update metadata to inherit tags defined in bigtop base layer
- update charm tests to use xenial base OS
- alter plugin to correctly make java relation optional
- use layer options to override NN and RM ports in puppet recipes
- run bigtop smoke tests for RM (yarn) and Slave (hdfs + mapreduce)
- run simple smoke tests for NN and plugin:
  - plugin lacks venv modules required to instantiate the Bigtop class, so it cannot call Bigtop.run_smoke_tests.
  - NN cannot successfully run the 'hdfs' smoke test. It fails on the distcp test, presumably because it lacks a component like the DN.
- readme updates for consistent messaging and 2.0 instructions


johnsca:
This is a pretty big PR, but it looks good.  I was able to build and deploy the updated charms and run the smoke-tests.  The README changes look good.

The commits do need to be squashed, though.

+1

petevg:
I get an error when deploying the bundle, but it may just be me. Here's what I did:

1) Edited bundle-local.yaml to fix the paths, as juju 2.0 doesn't like local:blah entries.
2) Ran the following bash commands (I omitted some extraneous 'ls' and 'pwd' commands):
```
   2004  cd bigtop-packages/src/charm/hadoop/
   2007  rm -Rf ~/Code/charms/trusty/*
   2010  for i in `ls`; do charm build $i; done
   2012  cd ../../../../bigtop-deploy/juju/hadoop-processing/
   2014  juju deploy bundle-local.yaml
```
3) I get the following error:

ERROR cannot deploy bundle: cannot add relation between "openjdk" and "namenode": cannot add relation "openjdk:java namenode:java": principal and subordinate applications' series must match

johnsca:
The charms must all be built and deployed as Xenial, including openjdk.  However, we're also removing openjdk from the bundle, as it is now optional (for choosing a JDK other than openjdk, as openjdk is provided by default), so you can just remove that from the bundle entirely.

The changes to the bundle should be addressed under https://issues.apache.org/jira/browse/BIGTOP-2550 once this is merged.

kwmonroe:
Thanks for the review guys.  @johnsca you're right in that this was larger than expected, mostly due to the same changes (README/metadata/layer/smoke-test) going into each of the 4 hadoop charms (NN, RM, Slave, Plugin).

In the future, I'll open a PR per charm to make the review more manageable.

@petevg cory is correct in that testing these charms via the bundle won't be ready until BIGTOP-2550 is merged.

PR #151, opened 2016-10-17
BIGTOP-2553: add rpc/http/https ports to template for non-HA mode

kwmonroe:
Honor the `[namenode|http|https]_port` settings in non-HA mode.

### DIFF
--- a/bigtop-deploy/puppet/modules/hadoop/templates/hdfs-site.xml
+++ b/bigtop-deploy/puppet/modules/hadoop/templates/hdfs-site.xml
@@ -86,7 +86,25 @@
   </property>
 
 <%   end -%>
-<% elsif @hadoop_security_authentication == "kerberos" -%>
+<% else -%>
+  <!-- non HA -->
+
+  <property>
+    <name>dfs.namenode.rpc-address</name>
+    <value><%= namenode_hosts[0] %>:<%= @hadoop_namenode_port %></value>
+  </property>
+
+  <property>
+    <name>dfs.namenode.http-address</name>
+    <value><%= namenode_hosts[0] %>:<%= @hadoop_namenode_http_port %></value>
+  </property>
+
+  <property>
+    <name>dfs.namenode.https-address</name>
+    <value><%= namenode_hosts[0] %>:<%= @hadoop_namenode_https_port %></value>
+  </property>
+
+<%   if @hadoop_security_authentication == "kerberos" -%>
   <property>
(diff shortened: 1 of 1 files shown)


johnsca:
Would the `namenode_hosts` contain more than one item in non-HA?  Perhaps it would be cleaner to just move the loop out of the conditional and rely on it only running once in the non-HA case.

Also, why are HA and Kerberos mutually exclusive?  The change preserves the existing logic, but the logic seems odd to me.

kwmonroe:
@johnsca, `namenode_hosts` has only 1 item in non-HA mode.  Note that once we hit the `else` to enter the "non HA" condition, we're not in any loop -- `namenode_hosts[0]` is how the one and only namenode was being referenced in other parts of the template, so I kept it the same:

https://github.com/apache/bigtop/blob/master/bigtop-deploy/puppet/modules/hadoop/templates/hdfs-site.xml#L98

Fwiw, I could use `@hadoop_namenode_host` if you think that would be cleaner.

As for HA/Kerberos, I do not know why they are mutually exclusive; as you noted, I preserved the original logic.

johnsca:
@kwmonroe My point was that, since `namenode_hosts` only contains one item in the non-HA case, and the only difference in the code between the HA case and non-HA case was whether we looped over the code, it could be factored out into a single loop outside of the HA / non-HA check and it would do the right thing by dint of there being only a single item in the list.

It would save some code repetition, and avoid needing to change the `elif` condition, though the other part of my comment is that that seems strange to me in the first place (but, since it's outside the scope of this change, anyway, avoiding changing it at all seems better).

kwmonroe:
I think you're saying the `namenode_hosts[host, index]` loop is the only difference, but that's a pretty big difference.  The hdfs-site.xml property keys are not the same in HA and non-HA cases.

For example, property key in HA:
`dfs.namenode.rpc-address.<%= @nameservice_id %>.nn<%= idx+1 %>`

Same property key for non-HA:
`dfs.namenode.rpc-address`

By the time we check for a single item list and have an `if/else` block rendering different properties based on that list size, I think we're no better than we are now.

johnsca:
Oh, of course you're correct.  My mistake.

+1

kwmonroe:
Pushed upstream with:

https://github.com/apache/bigtop/commit/cd55c144955e36188640c199b3100fefb3b28612

I forgot to add a "closes #151" to the commit, so i'm closing this out the old fashioned way.

PR #152, opened 2016-10-19
BIGTOP-2554: expose bind-host options in hieradata

kwmonroe:
Expose bind-host for RM and NN (rpc, http, and https) as hieradata options so we can set the appropriate config xml changes.


ktsakalozos:
LGTM +1

PR #156, opened 2016-11-07
BIGTOP-2575: zk charm test updates

kwmonroe:
Update the Juju charm for Zookeeper:

- ensure we deploy correct zk charm and watch for correct status during tests
- use new syntax for executing smoke tests
- include the network interface binding test (previously skipped)
- report the application version for status output


bigtop-packages/src/charm/zookeeper/layer-zookeeper/tests/10-bind-address.py
```diff
@@ -34,16 +33,16 @@ class TestBindClientPort(unittest.TestCase):
     def setUpClass(cls):
         cls.d = amulet.Deployment(series='xenial')
 
-        cls.d.add('zookeeper', charm='zookeeper', units=3)
+        cls.d.add('zk-test', charm='cs:xenial/zookeeper')
```
kwmonroe: @petevg, i think there was an issue where test tooling might pick a series that was different than the one given in our Deployment(), so I got in the habit of being explicit.  This did require tweaks to test locally, but it was better than stuff like cwr using the wrong charm.

This doesn't seem to be needed anymore.  Thanks for the catch, and thanks for improving my local test experience ;)

Fix incoming...

petevg:
LGTM/+1

PR #157, opened 2016-11-07
BIGTOP-2577: kafka charm test updates

kwmonroe:
Similar updates to what we did for the zk charm (BIGTOP-2575). Update the Juju charm for Kafka:

- update readme with language consistent with other bigtop charms
- ensure we deploy correct kafka charm and watch for correct status during tests
- use new syntax for executing smoke tests
- include the network interface binding test (previously skipped)
- report the application version for status output


petevg:
Tests work for me, and code looks good. LGTM/+1

kwmonroe:
Good catch in zookeeper -- you suggested the better approach of relaxing the charm definition in the test.. Most recent commit should address your concerns.  Thanks @petevg!

petevg:
My charms having been addressed, I am +1 on this one :-)

PR #158, opened 2016-11-08
[WIP] Added matrix test.

petevg:
This needs some refinement -- right now, it's just a copy of one of our
existing matrix test examples, with the hadoop bundle replacing
wiki-simple.

@cory_fu: my plan was to simplify this test after I got it verified working. We run into issues w/ the first deploy, though; apologies for the dust.

### DIFF
--- a/bigtop-deploy/juju/hadoop-processing/tests/test.matrix
+++ b/bigtop-deploy/juju/hadoop-processing/tests/test.matrix
@@ -1,0 +1,36 @@
+#!/usr/bin/env matrix
+"tests":
+- "name": deployment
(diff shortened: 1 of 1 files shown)


johnsca:
I feel like we need to parameterize the test in matrix to accept any bundle.  That is, this shouldn't really live in the bundle, but the bundle might contain a traffic plugin to generate Hadoop-specific work to run while glitch is running.

petevg:
Agree w/ @johnsca: We should parameterize bundle passing for the standard matrix test, rather than requiring everyone to drop a .matrix file into their bundle.

PR #161, opened 2016-11-14
BIGTOP-2586: hadoop-processing tests.yaml should include PPA for Amulet

johnsca:
Add source PPA for Juju bundles' tests.yaml (both hadoop-processing and
spark-processing).


kwmonroe:
This went in with PR #167

PR #162, opened 2016-11-17
BIGTOP-2570: ease hadoop charm debugging

kwmonroe:
We need to make sure our bigtop services are actually started before setting the `.started` state.

Without this fix, our NN will set `.started` even if `host.service_restart('hadoop-hdfs-namenode')` fails.  This is bad because our slave/datanode units will attempt to do things (like relate to the NN) when they can't.

This doesn't help us know why the service failed to start, but it does prevent other charms from relating to an unready application.


kwmonroe:
Though this started with a systemctl service that failed to start, it has turned into a crusade to make charm actions/status/logging better so we can more easily debug problems like this in the future.

In addition to handling `start_foo` better, i added some additional logging, corrected some bad status, and updated actions to better log their output.  I've updated the jira/pr title to reflect this extra scope.

bigtop-packages/src/charm/hadoop/layer-hadoop-resourcemanager/reactive/resourcemanager.py
```diff
@@ -131,16 +131,28 @@ def send_nn_spec(namenode):
...
+    if rm_started:
+        for port in get_layer_opts().exposed_ports('resourcemanager'):
+            hookenv.open_port(port)
+        set_state('apache-bigtop-resourcemanager.started')
+        hookenv.status_set('maintenance', 'resourcemanager started')
+        hookenv.application_version_set(get_hadoop_version())
+    else:
+        hookenv.log('YARN ResourceManager failed to start')
+        hookenv.status_set('blocked', 'resourcemanager failed to start')
+        remove_state('apache-bigtop-resourcemanager.started')
```
ktsakalozos: I guess this is "just in case"

ktsakalozos:
LGTM +1

bigtop-packages/src/charm/hadoop/layer-hadoop-resourcemanager/reactive/resourcemanager.py
johnsca: No, this will prevent related services from trying to use the RM before it is started.

johnsca:
LGTM as well :+1:

PR #163, opened 2016-11-22
BIGTOP-2603: NN/RM charm should include a spark user/group

kwmonroe:
When NN/RM are not colocated with spark, they do not know about the spark user/group.  When in `yarn-client` mode, the spark charm will create hdfs dirs with `spark` ACLs.  This leads to warnings like this on the namenode:

`WARN org.apache.hadoop.security.UserGroupInformation: No groups available for user spark`

Always create the spark user/group on NN and RM.  This also works if spark/NN/RM are all colocated.

Note that `hadoop` should be a secondary group (not primary) for extra users created by these charms, so this PR ensures that as well.

### DIFF
--- a/bigtop-packages/src/charm/hadoop/layer-hadoop-namenode/layer.yaml
+++ b/bigtop-packages/src/charm/hadoop/layer-hadoop-namenode/layer.yaml
@@ -8,14 +8,17 @@ options:
   apache-bigtop-base:
     groups:
       - 'mapred'
+      - 'spark'
       - 'yarn'
     users:
       mapred:
-        groups: ['hadoop', 'mapred']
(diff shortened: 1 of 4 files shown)


ktsakalozos:
LGTM +1 . Do you think we should be doing the same with other components that may use hdfs, like hbase?

kwmonroe:
Yeah, we probably should.  `oozie` and `hue` come to mind.  We chatted about this a little while ago.  It would really be nice if the charm that relied on `hadoop-plugin` would send user info over the plugin relation.

The plugin would then proxy that to the NN or RM as needed.  Unfortunately, this isn't really a priority at the moment, so I think we'll treat `spark` as special for now and work up a more generic solution later.

PR #167, opened 2016-12-02
BIGTOP-2616: refresh juju hadoop-processing bundle

kwmonroe:
We've made a number of improvements to the hadoop-processing bundle that are ready for review and inclusion upstream:
- ganglia/rsyslog have been updated from trusty to xenial
- add new machine constraints since aws now defaults to an 8gb root disk (we want 32g at a minimum)
- ensure applications are actually ready before running tests
- ensure the proper amulet test harness is included in the bundle tests

### DIFF
--- a/bigtop-deploy/juju/hadoop-processing/bundle-dev.yaml
+++ b/bigtop-deploy/juju/hadoop-processing/bundle-dev.yaml
@@ -39,7 +39,7 @@ services:
     to:
       - "4"
   ganglia:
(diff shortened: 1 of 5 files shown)


johnsca:
This looks good to me.  The tests are more specific, and are all passing for deployments on AWS and Azure, as evinced by http://cwr.vapour.ws/bundle_hadoop_processing/aba491a0c58843e0ab60d7c5291bf135/report.html

:+1:

PR #168, opened 2016-12-03
BIGTOP-2617: refresh juju spark-processing bundle

kwmonroe:
Similar to BIGTOP-2616, we've made improvements to the spark-processing bundle that are ready for review and inclusion upstream:

- update readme for consistency with other bundles as well as calling out new sparkbench capabilities
- refresh bundled charms, including the use of bigtop-zookeeper
- reduce spark machine count (2 units instead of 3, which still allows HA without as much $$$ in cloud resources)
- ganglia/rsyslog have been updated from trusty to xenial
- add new machine constraints since aws now defaults to an 8gb root disk (we want 32g at a minimum)
- run the bigtop smoke tests for spark/zookeeper
- ensure the proper amulet test harness is included in the bundle tests

### DIFF
--- a/bigtop-deploy/juju/spark-processing/README.md
+++ b/bigtop-deploy/juju/spark-processing/README.md
@@ -23,11 +23,11 @@ applications are included to monitor cluster health and syslog activity.
 
 ## Bundle Composition
 
-The applications that comprise this bundle are spread across 7 units as
+The applications that comprise this bundle are spread across 6 units as
 follows:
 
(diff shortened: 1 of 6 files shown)


johnsca:
This seems good to me.  The tests are improved, and are passing for deployments on GCE and Azure, as evinced by http://cwr.vapour.ws/bundle_spark_processing/d3bc4a08ce3c4b42a64f19d9bef0baea/report.html

:+1:

PR #179, opened 2017-02-20
[BIGTOP-2690] gradlew toolchain fails trying to download Ant 1.9.8

felixcheung:
Since Ant 1.9.9 is out, 1.9.8 is removed from some (all?) apache mirror, and gradlew toolchain fails


evans-ye:
Can you fire up a Bigtop JIRA and explain why you want this upgrade?

felixcheung:
merged

PR #182, opened 2017-03-01
[BIGTOP-2693] update readme for build and configuring git repo

felixcheung:
on configuring git repo for packages


evans-ye:
Good catch! Our doc is for sure outdated.
Can you refer to BIGTOP-2690 and create a JIRA with patch?

felixcheung:
committed

PR #186, opened 2017-03-14
Build/install protobuf-2.5 from source for Debian instead of using Ubuntu deb

SahilAggarwal:
Currently we use ubuntu deb to install protobuf-2.5 on debian jessie, this patch build and install protobuf-2.5 from source for debian.

### DIFF
--- a/bigtop_toolchain/manifests/protobuf.pp
+++ b/bigtop_toolchain/manifests/protobuf.pp
@@ -16,7 +16,7 @@
 class bigtop_toolchain::protobuf {
 
   case $operatingsystem{
-    /Ubuntu|Debian/: {
+    Ubuntu: {
       case $architecture {
         'amd64' : { $url = "https://launchpad.net/ubuntu/+source/protobuf/2.5.0-9ubuntu1/+build/5585371/+files/"
                     $arch= "amd64" }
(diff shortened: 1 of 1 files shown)


oflebbe:
maybe we need a require bigtop::packages

SahilAggarwal:
Appreciate your quick review.

Yes, we should add dependency to bigtop::packages.

Do you suggest to enable this for all platforms in separate PR or shall I do it here only?

oflebbe:
It's o.k. to handle this here.  We have the policy to have a JIRA with the commit message as summary linked to the github entry. If the JIRA is called something like "Build/install protobuf-2.5 from source" it would be sufficient.

evans-ye:
Sounds good to apply to all platforms.
Can you create a Bigtop JIRA to track the changes?

oflebbe:
Thanks Evans to tune in as well!

SahilAggarwal:
Have created Jira <https://issues.apache.org/jira/browse/BIGTOP-2698> but
can't assign it to me. If someone has proper creds please do it. Thanks.

SahilAggarwal:
@oflebbe Have made it generic for all platforms. I have tested it for ubuntu and debian jessie.

PR #187, opened 2017-03-20
BIGTOP-2703: refresh juju charms/bundles with metric and CI support

kwmonroe:
While this looks like a large PR, the majority is simply bumping charm revisions/locations in yaml files and updating READMEs to be consistent.  Other notable mentions include:

- include basic metric collection for NN/RM/ZK (e.g.: node count, useful to help application scale decisions)
- include ci-info.yaml to help Juju CI systems determine what to test in a bundle
- include hadoop-kafka bundle source, which was made generally available in the juju charm store months ago (yet missing from upstream repo)
- move bundle machine constraints to the application (else, scaling an application would get default cloud constraints)


c0s:
If it is all working and look good for other Juju's team guys - please go and commit this, perhaps after squashing them together. Thanks!

PR #188, opened 2017-03-21
BIGTOP-2706 Apex package does not run successfully after installation

chinmaykolhatkar:
While generating the rpm/deb, the package also bundles hadoop jars... Because of the clash of hadoop jars the apex cli fails to run.
Recently there has been change in min dependency of apex on hadoop from 2.2.0 to 2.6.0.
Proposal is to identify the min dependency of apex on hadoop at package time and exclude those jars in packaging.

### DIFF
--- a/bigtop-packages/src/common/apex/do-component-build
+++ b/bigtop-packages/src/common/apex/do-component-build
@@ -26,10 +26,10 @@ command -v mvn > /dev/null 2>&1 || {
 }
 
 # Build apex
-mvn clean package -DskipTests "$@"
+mvn clean package -DskipTests -Dhadoop.version=$HADOOP_VERSION "$@"
 
 # Find all the dependencies of apex
-mvn dependency:copy-dependencies -DoutputDirectory=`pwd`/all-deps -DincludeScope=runtime -DexcludeScope=test
+mvn dependency:copy-dependencies -DoutputDirectory=`pwd`/all-deps -DincludeScope=runtime -DexcludeScope=test -Dhadoop.version=$HADOOP_VERSION
 
 # Copy apex-engine jar from compiled path
 cp engine/target/apex-engine.jar all-deps/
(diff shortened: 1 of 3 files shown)


bigtop-packages/src/common/apex/do-component-build
```diff
@@ -28,6 +28,9 @@ command -v mvn > /dev/null 2>&1 || {
 # Build apex
 mvn clean package -DskipTests "$@"
 
+#Find hadoop.version from pom.xml
+APEX_HADOOP_VERSION=$(grep -oPm1 "(?<=<hadoop.version>)[^<]+" pom.xml)
```
c0s: The very first material line of this file is to source `bigtop.bom` that contains the version of the components, included into the stack. Doing this grep you might effectively build Apex against a different version of Hadoop, hence introducing the inconsistency from the get go. 
Any reason you don't want to build this component against Bigtop's Hadoop version? Thanks.

chinmaykolhatkar: Apex min hadoop version is 2.6.0. I believe if I use hadoop version from bigtop.bom, that would give me 2.7.3. What I want to do here is, find out the min version of hadoop which apex uses and exclude those from packaging.

Hence done this way.

chinmaykolhatkar:
@c0s Change the title of PR to be same as that of jira.

bigtop-packages/src/common/apex/do-component-build
tweise: For the purpose of this build, shouldn't it be the bigtop hadoop version, since Apex will run on top of it?

chinmaykolhatkar: I wanted to use the same. But that would mean excluding hadoop 2.7.3 dependencies and not 2.6.0 dependency.. We want to exclude the one that is referenced by apex. Hence done this way.

rvs: I think what @tweise and @c0s are saying is that you need to pass whatever flags are required to apex build so that it DOES pick up Hadoop version that is coming from Bigtop BOM. That's step #0 

Then you can decide whether you want to exclude the JAR or not -- but it HAD to build with Hadoop that is enforced by Bigtop stack.

Does it make it clear?

tweise: @rsv that sums it up. Indeed it should be possible to pass the hadoop.version to the apex build.

I would also like to point out that the Apex package can be assembled without building Apex from source, by optionally using the official release artifacts from central:  https://github.com/atrato/apex-cli-package/blob/master/pom.xml

That script also shows how the somewhat quirky dependency filtering pom can be simplified (and the stray legacy datatorrent references removed).

chinmaykolhatkar: Thanks @tweise @rvs @c0s . I'll update the PR.

c0s: > I would also like to point out that the Apex package can be assembled without building Apex from source, by optionally using the official release
We actually _always_ build all the components from scratch. That's the policy aimed to guarantee the compatibility through the stack.

chinmaykolhatkar: I'll be updating the PR where component is built from scratch.

chinmaykolhatkar:
@c0s @rvs @tweise Made the changes. I had to add a diff for apex to compile successfully with 2.7.x.
The reason for the patch is captured here for reference: APEXCORE-679

In future releases of apex, we'll take care of jira in apex, that is when I can remove the diff.

Requesting to review and merge the PR.

bigtop-packages/src/common/apex/do-component-build
```diff
@@ -26,10 +26,10 @@ command -v mvn > /dev/null 2>&1 || {
 }
 
 # Build apex
-mvn clean package -DskipTests "$@"
+mvn clean package -DskipTests -Dhadoop.version=$HADOOP_VERSION "$@"
 
 # Find all the dependencies of apex
-mvn dependency:copy-dependencies -DoutputDirectory=`pwd`/all-deps -DincludeScope=runtime -DexcludeScope=test
+mvn dependency:copy-dependencies -DoutputDirectory=`pwd`/all-deps -DincludeScope=runtime -DexcludeScope=test -Dhadoop.version=$HADOOP_VERSION
```
tweise: @chinmaykolhatkar look at how the relevant jar files are filtered here: https://github.com/atrato/apex-cli-package/blob/master/pom.xml

It does not require the brittle exclusions list that you have below, because it works based on the apex-engine dependencies.

chinmaykolhatkar: Agreed. I request this to be taken as an improvement for building apex component as a seperate Jira. Currently the apex component is broken and I believe that should be fixed. Let me know if this is fine.

rvs: @chinmaykolhatkar I've committed your change. I would like to make sure that we fixed the problem. What's the easiest way to run a test for this change?

chinmaykolhatkar: Thanks @rvs  Fastest way is to check if the deb/rpm package that there is no hadoop jar present. I'm writing package tests after for this for another jira (BIGTOP-2375).. That would make it even easier.

tweise: @chinmaykolhatkar agreed, that should be taken up separately.

tweise: @chinmaykolhatkar did you create a JIRA for the build improvement?

chinmaykolhatkar: @tweise Here is the jira: https://issues.apache.org/jira/browse/BIGTOP-2735

PR #197, opened 2017-04-21
BIGTOP-2747: new charm revs for bigtop-1.2

kwmonroe:
Bring charm/bundle source up to date with bigtop-1.2.  This is mostly a README/metadata refactor with sprinkles of a more consistent UX for charm actions.

There is one functional difference that comes with the move to support bigtop-1.2.  The `hadoop-spark` bundle previously deployed 3 units of zookeeper, but with spark in yarn-client mode, this makes no sense.  If users want Spark HA with Zookeeper, they should use the `spark-processing` bundle.  Therefore, while rev'ing `hadoop-spark` for this PR, I've removed zookeeper references from the bundle.

### DIFF
--- a/bigtop-deploy/juju/hadoop-spark/bundle.yaml
+++ b/bigtop-deploy/juju/hadoop-spark/bundle.yaml
@@ -43,7 +43,7 @@ services:
     to:
       - "4"
   spark:
-    charm: "cs:xenial/spark-33"
+    charm: "cs:xenial/spark-34"
     constraints: "mem=7G root-disk=32G"
     num_units: 1
     options:
--- a/bigtop-deploy/juju/hadoop-spark/bundle.yaml
+++ b/bigtop-deploy/juju/hadoop-spark/bundle.yaml
@@ -1,6 +1,6 @@
 services:
   namenode:
-    charm: "cs:xenial/hadoop-namenode-13"
+    charm: "cs:xenial/hadoop-namenode-14"
     constraints: "mem=7G root-disk=32G"
     num_units: 1
     annotations:
@@ -9,7 +9,7 @@ services:
     to:
       - "0"
   resourcemanager:
-    charm: "cs:xenial/hadoop-resourcemanager-14"
+    charm: "cs:xenial/hadoop-resourcemanager-16"
     constraints: "mem=7G root-disk=32G"
     num_units: 1
     annotations:
@@ -18,7 +18,7 @@ services:
     to:
       - "0"
   slave:
-    charm: "cs:xenial/hadoop-slave-13"
+    charm: "cs:xenial/hadoop-slave-15"
     constraints: "mem=7G root-disk=32G"
     num_units: 3
     annotations:
@@ -29,13 +29,13 @@ services:
       - "2"
       - "3"
   plugin:
-    charm: "cs:xenial/hadoop-plugin-13"
+    charm: "cs:xenial/hadoop-plugin-14"
     annotations:
       gui-x: "1000"
       gui-y: "400"
   client:
     charm: "cs:xenial/hadoop-client-3"
-    constraints: "mem=3G"
+    constraints: "mem=7G root-disk=32G"
     num_units: 1
     annotations:
       gui-x: "1250"
@@ -43,7 +43,7 @@ services:
     to:
       - "4"
   spark:
-    charm: "cs:xenial/spark-31"
+    charm: "cs:xenial/spark-33"
     constraints: "mem=7G root-disk=32G"
     num_units: 1
     options:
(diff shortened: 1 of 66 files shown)


bigtop-deploy/juju/hadoop-spark/bundle.yaml
```diff
@@ -29,21 +29,21 @@ services:
...
   plugin:
-    charm: "cs:xenial/hadoop-plugin-13"
+    charm: "cs:xenial/hadoop-plugin-14"
     annotations:
       gui-x: "1000"
       gui-y: "400"
   client:
     charm: "cs:xenial/hadoop-client-3"
-    constraints: "mem=3G"
+    constraints: "mem=7G root-disk=32G"
```
johnsca: Is this right?  Does the client / gateway node really need such large constraints?

Actually, do we need any application-level constraints if we're already specifying them in the machines section?

bigtop-packages/src/charm/hadoop/layer-hadoop-resourcemanager/actions/smoke-test
```diff
@@ -43,6 +43,6 @@ smoke_env = {
 bigtop = Bigtop()
 result = bigtop.run_smoke_tests(smoke_components, smoke_env)
 if result == 'success':
-    hookenv.action_set({'outcome': result})
+    hookenv.action_set({'outcome': 'success'})
 else:
-    fail('{} smoke tests failed'.format(smoke_components), result)
+    fail('{} smoke tests failed with: {}'.format(smoke_components, result))
```
johnsca: Why combining the output and message here?  Seems like it would make the output harder to handle separately.  Is it just because the message is more important for CI?

I do like the consistent `outcome` field.

bigtop-packages/src/charm/pig/layer-pig/actions/smoke-test
```diff
@@ -63,8 +62,8 @@ else:
 
 # The smoke tests analyze /etc/passwd. Successful runs will have an 'ubuntu'
 # passwd entry in the output.
+hookenv.action_set({'raw': output})
```
johnsca: Why do some spots use `meta.raw` and others just use `raw`?

bigtop-packages/src/charm/spark/layer-spark/actions/restart-spark-job-history-server
```diff
@@ -15,22 +15,22 @@
...
-    from charmhelpers.core import host, hookenv, unitdata
-    from jujubigdata import utils
-    charm_ready = True
-except ImportError:
-    charm_ready = False
 
-if not charm_ready:
-    from subprocess import call
-    call(['action-fail', 'Spark service not yet ready'])
-    sys.exit(1)
```
johnsca: This was originally intended to handle the case where the action somehow ran before the bootstrap code, but I don't think that's actually possible, so +1 to this change.

bigtop-deploy/juju/hadoop-spark/bundle.yaml
kwmonroe: @johnsca, this is correct - in this rev of the bundle, spark is colocated with client.  I did this because in yarn-client mode, spark doesn't need a whole new unit by itself.  The driver process can tax the unit a bit more, so I bumped ram to 7g.  This also benefits users because they can run `spark-submit` and `hadoop jar` on the same unit.  Previously, they'd have to run spark jobs on the spark unit and hadoop jobs on the client unit.

Yes, app-level constraints are required in addition to machine constraints.  Without them, the initial deployment would use the machine constraints, but a subsequent `add-unit` would fall back to the provider default.  Ideally, we could remove the machine constraints and solely rely on app-constraints.  We'll do that once https://bugs.launchpad.net/juju/+bug/1676986 is fixed.

bigtop-packages/src/charm/hadoop/layer-hadoop-resourcemanager/actions/smoke-test
kwmonroe: @johnsca, I merged the output with the message to be consistent with other bigtop charms.  When an action fails, users expect to see the reason in the `message` key.  Without this change, they would see the abbreviated failure message and need to know to look elsewhere for an `output` key.

bigtop-packages/src/charm/pig/layer-pig/actions/smoke-test
kwmonroe: @johnsca, proper benchmarks (using `charms.benchmark`) set benchmark-related bits like start/stop/composite/raw under the `meta` key.  Non-benchmark actions may still have helpful `raw` data, but I chose to display that as `results.raw` instead of creating a new `meta` key that only included `results.meta.raw`.

bigtop-packages/src/charm/spark/layer-spark/actions/restart-spark-job-history-server
kwmonroe: Thanks for the history.  I originally assumed this came in before `is_state` was a thing.

bigtop-deploy/juju/hadoop-spark/bundle.yaml
johnsca: Ah, that makes sense.  Thanks for the info!

bigtop-packages/src/charm/hadoop/layer-hadoop-resourcemanager/actions/smoke-test
johnsca: Fair enough.  +1

evans-ye:
Thanks for the patch and review @kwmonroe @johnsca.

PR #198, opened 2017-04-22
Small fixes for bigtop 1.2.0

alaunay:
Those small fixes allow for compilation using a fresh Debian distro/user, without prior downloaded files


alaunay:
this one went sideways :)

PR #199, opened 2017-04-26
BIGTOP-2355: Update Mahout version to 0.13.0

andrewmusselman:
Tested with deb, rpm, pkg, and tar; expected files are all created but was unable to test deb install due to a hadoop-client versioning issue detailed on the dev list.


c0s:
Seems LGTM

PR #200, opened 2017-04-27
BIGTOP-2723:Fix asciidoctor-maven-plugin for HBase build on AArch64

guyuqi:
When we build the hbase-pkg, we are facing below issue:
"ERROR] Failed to execute goal org.asciidoctor:asciidoctor-maven-plugin:1.5.2.1:process-asciidoc (output-pdf) on project hbase: Execution output-pdf of goal org.asciidoctor:asciidoctor-maven-plugin:1.5.2.1:process-asciidoc failed: (NotImplementedError) fstat unimplemented unsupported or native support failed to load -> [Help 1]".

This error took place because asciidoctor-maven-plugin used an old version of Jruby (1.7.26) which didn’t work with AArch64 (due to the fact that Jffi 1.2.12 which Jruby 1.7.26 depends on has no native AArch64 libraries).
The submitted patch resolves the issue by upgradeing Jruby to version 9.1.8.0 that supports AArch64. Please note that HBase itself also uses JRuby 1.7.x which is not upgraded because Jruby 9.1.8.0 is not compatible with Hbase.


JunHe77:
@guyuqi, could you pls rebase the patch as it has conflicts with HEAD and re-submit? Thanks.

PR #205, opened 2017-05-10
BIGTOP-2764: deployment failure when roles include spark::common and spark::yarn*

kwmonroe:
We need to ensure the `spark-datanucleus` package is only defined once.  There's a hack in place so it is always defined in `spark::common` (see BIGTOP-2268):

https://github.com/apache/bigtop/blob/master/bigtop-deploy/puppet/modules/spark/manifests/init.pp#L154

This PR comments out the other definition, which should be reinstated if/when 2268 is fixed.

### DIFF
--- a/bigtop-deploy/puppet/modules/spark/manifests/init.pp
+++ b/bigtop-deploy/puppet/modules/spark/manifests/init.pp
@@ -148,14 +148,13 @@ class spark {
       $history_log_dir = "hdfs:///var/log/spark/apps",
   ) {
 
-    package { 'spark-core':
-      ensure => latest,
-    }
 ### This is an ungodly hack to deal with the consequence of adding
 ### unconditional hive-support into Spark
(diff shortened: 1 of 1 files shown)


kwmonroe:
I've updated the original workaround to include `spark::datanucleus` instead of redefining the package.  This is much cleaner than including multiple comment blocks to explain the 2268 hack.

kwmonroe:
[spark-39](https://jujucharms.com/spark/39) contains this fix, and is green across clouds for hadoop + spark in yarn mode:

http://bigtop.charm.qa/cwr_bundle_hadoop_spark/33/report.html

ktsakalozos:
LGTM +1

evans-ye:
Thanks for the review @ktsakalozos.
+1 as well.

kwmonroe:
Closed by https://github.com/apache/bigtop/commit/73864a30b29ee89655772ce82b4041c2564d0711

PR #206, opened 2017-05-10
BIGTOP-2765: update roles logic in spark/zeppelin

kwmonroe:
Rework role logic for spark/zeppelin charms.

- spark:  only include master/worker when running in standalone mode
- spark:  include spark-yarn roles when in yarn-* mode
- zeppelin: include spark-client and yarn roles to ensure the embedded spark is configured properly for standalone zeppelin deployments.


johnsca:
LGTM

PR #222, opened 2017-05-27
BIGTOP-2776: corrects patch for Bump Apache Apex version to 3.6.0

alaunay:


alaunay:
sorry, again a bad one, still having trouble with git

PR #229, opened 2017-06-02
BIGTOP-2795: spark charm: fix sparkpi and rework start/stop logic

kwmonroe:
See [bigtop-2795](https://issues.apache.org/jira/browse/BIGTOP-2795) for details.

- add bundle descriptions to the hadoop-spark yaml files
- expose the worker ui port
- rework start/stop logic to handle ports and wait when appropriate
- fix sparkpi.sh by not hard coding the spark-examples.jar path


kwmonroe:
During test, I found the `submit` action to be waaay too complex.  I refactored it to use a simpler set of arguments, and then documented this in the readme.

While I was in the readme, I noticed we didn't document the `*_memory` config options, so I added a section for those.

PR #231, opened 2017-06-06
BIGTOP-2798 Apex component has duplicate slf4j binding

tweise:


tweise:
@evans-ye can you please take a look?

PR #234, opened 2017-06-10
BIGTOP-2802. Some packages don't create necessary groups on debian-based system if the users correspond to them already exist

sekikn:


sekikn:
Thanks! In addition to that, this fix also covers the case that a user already exists but its group is not created. In such a case, the original scripts don't create the groups, so the final state differs from the normal case. It's undesirable.

oflebbe:
adduser on centos-7 does not have that flag '--ingroup'. Sadly.

oflebbe:
IMHO ambari should be fixed. Why does it create broken user/group configs?

rvs:
@sekikn I think the patch looks reasonable, @oflebbe did you mistype? This patch only applies on Debian side -- so if you know any caveats of using --ingroup on that platform -- please let us know. Otherwise -- I guess the patch can go in?

oflebbe:
Oh yes, it was too late that night. The patch looks very good for debian. Please apply

sekikn:
@c0s @oflebbe @rvs thanks for the reviews! I'm going to check this in.

PR #241, opened 2017-06-29
BIGTOP-2827: juju bundle refresh (june 2017)

kwmonroe:
Refresh Juju bundles.  See [BIGTOP-2827](https://issues.apache.org/jira/browse/BIGTOP-2827) for details.


ktsakalozos:
LGTM +1

PR #245, opened 2017-07-05
BIGTOP-2832. Toolchain failed to install on Debian 8

evans-ye:


oflebbe:
does it break debian-9?

c0s:
We aren't supporting Debian-9 yet, right? Shall we worry about this once we are dealing with the new version?

evans-ye:
I can't get puppet 4 to run on Debian-9 regardless of this patch. If you already have some experience playing around with Puppet 4, I'd love to make this patch more robust.

This is required to have 1.2.1 slaves successfully built.
Then, we can produce artifacts along with the release.

c0s:
+1

evans-ye:
Okay. Let me commit this first to unlock the release.

PR #252, opened 2017-07-07
BIGTOP-2836: charm metric collector race condition

kwmonroe:
Ensure `echo 0` is the last thing to run so that the metric hook does not cause a failed deployment. Works in all tested scenarios:

False && False || echo 0
- 0

True && False || echo 0
- 0

False && True || echo 0
- 0

True && True || echo 0
- True


c0s:
wow, an interesting approach indeed. Looks legit, though

johnsca:
LGTM

PR #260, opened 2017-07-23
BIGTOP-2851: Add debian 8 aarch64 bigtop puppet

nareshgbhat:
Add Debian 8 bigtop-puppet for AArch64 architecture.

Signed-off-by: Naresh Bhat <naresh.bhat@linaro.org>


c0s:
That looks good. I will commit this shortly. Thank you!

PR #263, opened 2017-07-24
BIGTOP-2857: Add aarch64 support for fedora25

JunHe77:
Change-Id: I00f8f6a9bb4c46cb93891477153f728ac945dff5


c0s:
+1, will commit in a bit.

PR #264, opened 2017-07-24
BIGTOP-2858:Add AArch64 support for Debian 9

guyuqi:
Change-Id: I5c77ba239375658223eb9025308f98d2bdb57e58

Signed-off-by: Yuqi Gu <yuqi.gu@arm.com>


c0s:
While I am not sure we are going to have the official support for Debian 9 - well, I just don't remember if we have discussed it yet ;) - the patch looks good and I am committing this. Thanks!

PR #268, opened 2017-07-28
BIGTOP-2860:Fix TestSpark.groovy syntax error

guyuqi:
Change-Id: If3dba95133219dd16b9870b54f072c01328fbdaf

Signed-off-by: Yuqi Gu <yuqi.gu@arm.com>


nareshgbhat:
As far as I understand the below code change must be by mistake in the patch

static final String SPARK_PORT = System.getenv("SPARK_MASTER_PORT")

should be changed to as below ?

static final String SPARK_MASTER_PORT = System.getenv("SPARK_MASTER_PORT")

guyuqi:
Fixed! thanks naresh!

c0s:
I will squash the commits to a single one before pushing to the master branch

PR #272, opened 2017-08-04
BIGTOP-2864:Multiple SLF4J bindings in tez

loleek:
When using Tez, there will be "Class path contains multiple SLF4J bindings" warnings. This caused by existence of 'slf4j-log4j12-*.jar' in tez lib.
The same issue resolved before.
Reference : https://issues.apache.org/jira/browse/BIGTOP-583


c0s:
So, I assume the one from Hadoop will be used instead, right? 
Looks reasonable, I will commit this shortly.

loleek:
Yes, there is one in hadoop.

PR #274, opened 2017-08-07
BIGTOP-2866:Fix rmr and people directory error

nareshgbhat:
The below error is issued while running the spark smoke
tests on AArch64 architecture.  The patch will fix the issues.
...
....
org.apache.bigtop.itest.spark.TestSpark STANDARD_OUT
Failed command: hdfs dfs -rmr people* examples
error code: 1
stdout: [Deleted examples]
stderr: [rmr: DEPRECATED: Please use 'rm -r' instead.,
rmr: `people*': No such file or directory,
...
.....

Signed-off-by: Naresh Bhat <naresh.bhat@linaro.org>


c0s:
That must break the test everywhere, ie independent of a platform. +1

PR #276, opened 2017-08-16
Add unomi test artifacts

cscmucmm:
Add unomi test artifacts pom.xml file for testing


cscmucmm:
Add unomi test artifacts pom.xml file for testing

PR #282, opened 2017-09-05
BIGTOP-2878 : Download task does not work for git repositories

zenmyo:


packages.gradle
```diff
@@ -185,7 +185,7 @@ def genTasks = { target ->
     def final GIT_USER_NAME = config.bigtop.components[target].git.user ?: config.bigtop.git.user
     def final GIT_ACCESS_TOKEN = config.bigtop.components[target].git.token ?: config.bigtop.git.token
 
-    if (!DOWNLOAD_URL)
+    if (!DOWNLOAD_URL && !(GIT_REPO && GIT_REF))
```
c0s: Good catch!
Took me a while to get the full picture, and I even had to write a small test to see the truth matrix
```
#!/usr/bin/env groovy

def final BOOLS = [false, true]
for (DOWNLOAD_URL in BOOLS)
  for (GIT_REPO in BOOLS)
    for (GIT_REF in BOOLS) {
      println("$DOWNLOAD_URL $GIT_REPO $GIT_REF")
      println ("\t\t\t ${(!DOWNLOAD_URL && !(GIT_REPO && GIT_REF)).toString().toUpperCase()}")
    }
```
but everything worked out just fine. 
+1 will commit it shortly

PR #293, opened 2017-10-02
ADH-207 Remove h2 pig classifier in Phoenix deps

acmnu:
There is a regression in Pig relase 0.17
After PIG-4923 they drop h2 classifer.
There is only one jar now and it suports Hadoop 2

That commit also adds BIGTOP_PATCH macro to Phoenix


acmnu:
Wrong caption

PR #300, opened 2017-11-14
BIGTOP-2917: Fix Ignite smoke test failure with JDK8

JunHe77:
Change-Id: I687a59386d5c9dfa4d7dff054bfdf16d2883e1ac

### DIFF
--- a/bigtop-packages/src/common/ignite-hadoop/do-component-build
+++ b/bigtop-packages/src/common/ignite-hadoop/do-component-build
@@ -20,4 +20,5 @@ set -ex
 
 export MAVEN_OPTS="-Xmx512M"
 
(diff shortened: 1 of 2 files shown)


c0s:
While it looks like a good solution in the interim, could we open an upstream ticket for the Ignite to fix this issue?

```sed -i '/asm-all/{n;s/4.2/5.0.2/}' modules/hadoop/pom.xml```

c0s:
BTW, the synopsis of the PR (and the commit ticket) **should** be the same as the JIRA ticket, because the latter is The Source of truth.
I have fixed it in this particular case, but please take it into advisement in the future. Thanks!

PR #304, opened 2017-11-14
BIGTOP-2925. Missing license header for .travis.yml

evans-ye:
Signed-off-by: Evans Ye <evansye@apache.org>


c0s:
+1
Would be great to have automatic RAT check!

PR #306, opened 2017-11-18
BIGTOP-2932: Patch to compile with openssl 1.1 (HADOOP-14597)

valtri:


oflebbe:
Could you please close it since BIGTOP-2932 is committed ?
Thanks you very much
  Olaf

PR #308, opened 2017-11-28
hue: make it buildable on Debian 9 'stretch'

hrw:
Debian moved from MySQL to MariaDB so we need to install proper package
to satisfy build dependencies.


evans-ye:
I'm a little bit confused? Is this PR linked to BIGTOP-2946?

evans-ye:
Oh my bad! This is already attached. 
Could you help to amend your commit message so that the git log is easier to trace? Typically we use JIRA title as commit message. Thanks!

hrw:
hue is no more

PR #309, opened 2017-11-28
oozie: disable dh_strip_nondeterminism to get it built on Debian 9

hrw:
Debian 9 uses dh_strip_nondeterminism to have reproducible builds. Part
of it is removing timestamps from tarballs and oozie has read-only one.


evans-ye:
Can you fire up a Bigtop JIRA and link this PR to it? Thanks!

PR #313, opened 2017-12-01
fix bug: spark/do-component-build

CZWin32768:
The mvn command in the last line of spark/do-component-build uses "mvn", but here should be "./build/mvn".

make-distribution.sh will only run "mvn clean package", so in order to get the Spark packages installed in the local Maven repository (or to run the tests), we need to run "mvn install" again.

make-distribution.sh uses "./build/mvn" but in do-component-build "mvn" is used, which may be in wrong version.

This may cause that "mvn install" build failure after "mvn clean package" build successfully.

### DIFF
--- a/MAINTAINERS.txt
+++ b/MAINTAINERS.txt
@@ -20,7 +20,7 @@ pig: evans ye, daniel dai
 puppet recipes: jay vyas, cos, evans ye, rvs
 qfs: kstinson <kstinson@quantcast.com>
 smoke-tests: jay vyas, david capwell
-spark: jay vyas, youngwoo kim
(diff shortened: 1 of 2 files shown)


evans-ye:
Hi @CZWin32768 thanks for the PR.
Can you create a Bigtop JIRA for this and attach this PR to it?
https://cwiki.apache.org/confluence/display/BIGTOP/How+to+Contribute#HowtoContribute-Howtogeneratepatchesbycreatingpullrequests

CZWin32768:
Of course. Thanks.

CZWin32768:
@evans-ye Thanks. I'll create a JIRA for this.

PR #315, opened 2017-12-01
BIGTOP-2954: fix bug of do-component-build which may cause spark build failure

CZWin32768:
The mvn command in the last line of spark/do-component-build uses "mvn", but here should be "./build/mvn".
make-distribution.sh will only run "mvn clean package", so in order to get the Spark packages installed in the local Maven repository (or to run the tests), we need to run "mvn install" again.
make-distribution.sh uses "./build/mvn" but in do-component-build "mvn" is used, which may be in wrong version.
It will cause build failure when a mvn of old version was set in PATH.
This may cause that "mvn install" build failure after "mvn clean package" build successfully.


CZWin32768:
Now use --mvn in spark do-component-build to make sure the build environment is aligned across Bigtop builds.

PR #316, opened 2017-12-08
BIGTOP-2957: Upgrade YCSB to 0.12.0

youngwookim:


bigtop-packages/src/common/ycsb/install_ycsb.sh
```diff
@@ -74,6 +74,25 @@ for var in PREFIX BUILD_DIR ; do
...
 
 install -d -m 0755 $PREFIX/$LIB_DIR
+install -d -m 0755 ${BIN_DIR}
+
 (cd ${BUILD_DIR} && tar -cf - .)|(cd $PREFIX/${LIB_DIR} && tar -xf -)
+rm -rf $PREFIX/$LIB_DIR/bin/*.bat
+
+wrapper=$BIN_DIR/ycsb
+cat >>$wrapper <<EOF
+#!/bin/sh
```
evans-ye: I remember once before we've agreed to use /bin/bash ;)

PR #325, opened 2017-12-25
BIGTOP-2968: Bump Hive version to 2.3.2

youngwookim:


c0s:
Wow, quite a bit of work has gone into this one! On one side, I am bit worries about the number of the patches for an upstream project. On the other hand, perhaps Hive isn't yet ready to work with Spark2? Are you planning to offer the changes to the Hive community as well, so we can get rid off the dynamic patching in the future?

Thansk!

youngwookim:
Cos, Thanks for the comment.
HIVE-16302 is in Hive 3.0.0 so the patch can be removed when we upgrade Hive to 3. On the Hive side, HoS(Hive on Spark) should work with Spark 2+ and the Spark side, SparkSQL 2.2+ should work with Hive 2.x Metastore (AFAIK) Please see my comment @ BIGTOP-2945

in short, if we have both Hive 2.3 and Spark 2.2(BIGTOP-2969), it should work at all.

c0s:
Ah, got it. Cool, thanks for fixing this! LGTM!

PR #326, opened 2017-12-26
Add Ranger

Xingwd:
package ranger rpm in bigtop

### DIFF
--- a/bigtop-packages/src/common/ranger/do-component-build
+++ b/bigtop-packages/src/common/ranger/do-component-build
@@ -1,0 +1,28 @@
+#!/bin/sh
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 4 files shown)


vgrivel:
https://github.com/apache/bigtop/pull/326/files#diff-f77acdf59075d87e5ef032c3ec6b14bbR415
`file /usr/lib/kafka/libs from install of ranger-kafka-plugin-2.0.0-1.el7.x86_64 conflicts with file from package kafka-2.4.1-1.el7.noarch`
I think this line need to be updated by %{kafka_home}/libs/* to not create a dependency on the folder as this folder is owned by kafka.
I didn't check the other plugins but it worth checking if there is other conflicts between master components and plugins.

sekikn:
Hi @Xingwd, we are going to add Ranger to our component stack in the next release.
Its latest version is 2.2.0. Would you be willing to update your PR, or may I take over your effort?

bhavikpatel9977:
Hi @sekikn, I think @Xingwd is not working on this feature. You can take the ownership of this feature.

sekikn:
I've just submitted the updated version of this PR as #929, so I'll close this PR.
Thank you so much for your contribution, @Xingwd!

PR #329, opened 2017-12-26
BIGTOP-2960: Building Hue requires python-ipaddress package

youngwookim:


oflebbe:
@youngwookim  : Yep, this may fix this problem . But I like to fix it more the python way by installing dependencies via pip upstream in the hue virtualenv . Did you do a runtime test for hue? I found a couple of issues with our new setup  .

youngwookim:
Olaf, Thanks for the comment. That's just what I wanted to ask :-) Patch does not work as I expected. Need to dig around in upstream project. I'll close this PR

PR #338, opened 2018-02-05
BIGTOP-2988: Incorrect use "dnf update" in puppetize.sh

JunHe77:
The cmd "dnf -y update" is not correctly used as the
purpose here is to "synchronize the package index
files from their sources".
For this purpose the cmd should be "dnf check-update"

Change-Id: I9ed7b7b631f5fabc00a4ef679d30a68f5c68b539


JunHe77:
Merged by upstream.

PR #341, opened 2018-02-26
BIGTOP-3005 Add zkpeer-relation-changed hook to zookeeper charm.

Pekkari:
Check quorum status and update config accordingly if zkpeer relation changes.

This change add a zkpeer-relation-changed hook to refresh the configuration of
the server.[\d] lines in zoo.cfg to match any new information received either through the
relation manually, or from juju.

Signed-off-by: José Pekkarinen <jose.pekkarinen@canonical.com>

### DIFF
--- a/bigtop-packages/src/charm/zookeeper/layer-zookeeper/reactive/zookeeper.py
+++ b/bigtop-packages/src/charm/zookeeper/layer-zookeeper/reactive/zookeeper.py
@@ -196,6 +196,12 @@ def check_cluster_departed(zkpeer, zkpeer_departed):
     check_cluster(zkpeer)
 
 
+@when('zookeeper.started', 'leadership.is_leader', 'zkpeer.changed')
+def check_cluster_changed(zkpeer):
+    check_cluster(zkpeer)
+    zkpeer.dismiss_changed()
+
+
 @when('leadership.changed.restart_queue', 'zkpeer.joined')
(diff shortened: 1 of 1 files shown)


kwmonroe:
@Pekkari I don't understand what a `zkpeer.changed` handler buys us given that we already run a `check_cluster` when peers join and depart.  Am I missing a case where a peer would change that is not covered by join/depart?

PikkuJose:
I'm trying to ease the case where you have several spaces, and zookeeper was deployed in the wrong one. Nowadays, you can set the constraint in juju, and force departure and rejoin, or you can set the right value in the relation, and update manually the config files in every unit. This would allow you to set the value in the relationship, and get it propagated to the config automatically.

Pekkari:
I'm trying to ease the case where you have several spaces, and zookeeper was deployed in the wrong one. Nowadays, you can set the constraint in juju, and force departure and rejoin, or you can set the right value in the relation, and update manually the config files in every unit. This would allow you to set the value in the relationship, and get it propagated to the config automatically.

kwmonroe:
+1, thx for the explanation @Pekkari.  A quick deploy with this addition lgtm.

kwmonroe:
Rev 42 contains this update and has been released to the edge channel in the charm store:

https://jujucharms.com/zookeeper/42

Pending CI tests to move it through to stable.

PR #342, opened 2018-02-27
BIGTOP-3004: Fix HBase build failure on Debian/Fedora

JunHe77:
Debian/Fedora uses bash-4.4.12 which treats the extra
trailing backslash as error. HBase has patch for this
and should be backported.

Change-Id: I11353ec0fcdca6e04373c0dac34e0e23aa757f30


JunHe77:
Close this as it has been merged.

PR #344, opened 2018-02-28
BIGTOP-3007: add properties to hive-site.xml for zookeeper support

jamesbeedy:
In support of hiveserver2 table lock, hive config needs to expose the properties:
* hive.support.concurrency
* hive.zookeeper.quorum

Added hive.zookeeper.quorum and hive.support.concurrency in support of
hiveserver2 table lock.

Add vars to init.pp.

Add zookeeper interface and corresponding reactive bits.

Modified install function in lib to facilitate zk_hosts.

Update deployment_matrix.

Fixes: https://issues.apache.org/jira/browse/BIGTOP-3007


bigtop-deploy/puppet/hieradata/bigtop/cluster.yaml
```diff
@@ -192,6 +192,9 @@ hadoop_hive::common_config::hbase_zookeeper_quorum: "%{hiera('hadoop_hbase::comm
 hadoop_hive::common_config::kerberos_realm: "%{hiera('kerberos::site::realm')}"
 hadoop_hive::common_config::metastore_uris: "thrift://%{hiera('bigtop::hadoop_head_node')}:9083"
 
+hadoop_hive::common_config::hive_zookeeper_quorum: "%{hiera('hadoop_hive::common_config::hive_zookeeper_quorum')}"
```
kwmonroe: Per the dev list [discussion](https://mail-archives.apache.org/mod_mbox/bigtop-dev/201803.mbox/%3C85DF9170-CEC9-4462-B554-5CAF892B0570%40oflebbe.de%3E), we don't need to alter cluster.yaml for these new properties.

bigtop-packages/src/charm/hive/layer-hive/lib/charms/layer/bigtop_hive.py
```diff
@@ -69,35 +83,44 @@ def configure_hive(self):
         config = hookenv.config()
         hive_env = self.dist_config.path('hive_conf') / 'hive-env.sh'
         utils.re_edit_in_place(hive_env, {
-            r'.*export HADOOP_HEAPSIZE *=.*': 'export HADOOP_HEAPSIZE=%s' % config['heap'],
```
kwmonroe: Looks like the rest of the changes to this file are lint fixes.  While they are much appreciated, in the future, please make them in a separate PR.  It's a bit harder to focus on the functional code changes with these formatting changes.

bigtop-packages/src/charm/hive/layer-hive/reactive/hive.py
```diff
@@ -78,9 +87,13 @@ def install_hive(hadoop):
     else:
         hbserver = None
 
+    # Get zookeepers or None
+    zks = KV.get('zookeepers', None)
```
kwmonroe: I'm curious why we're using `unitdata.kv` here.  If you look a few lines higher at the hbase block, we could use the same logic to get zks when `is_state(zookeeper.ready)` is available.

bigtop-deploy/puppet/hieradata/bigtop/cluster.yaml
kwmonroe: Doh!  @jamesbeedy you ninja'd me :)  Thx for removing `cluster.yaml` updates.

bigtop-packages/src/charm/hive/layer-hive/lib/charms/layer/bigtop_hive.py
```diff
@@ -36,31 +36,45 @@ def install(self, hbase=None):
         # Prep config
         roles = ['hive-client', 'hive-metastore', 'hive-server2']
         metastore = "thrift://{}:9083".format(hookenv.unit_private_ip())
-        if hbase:
-            roles.append('hive-hbase')
-            hb_connect = "{}:{}".format(hbase['host'], hbase['master_port'])
-            zk_connect = hbase['zk_connect']
-        else:
-            hb_connect = ""
-            zk_connect = ""
```
kwmonroe: Upon closer inspection, there's a side effect of not forcefully setting this to `""`.  With no value, we fall into the default provided by `cluster.yaml`:

https://github.com/apache/bigtop/blob/master/bigtop-deploy/puppet/hieradata/bigtop/cluster.yaml#L191

which falls up to the hbase config:

https://github.com/apache/bigtop/blob/master/bigtop-deploy/puppet/hieradata/bigtop/cluster.yaml#L132

which makes the namenode (bigtop head node) the default value.  This means on a deployed hive charm, you'll get the following in the config, even when hbase is not deployed at all:

```
<property>
  <name>hbase.zookeeper.quorum</name>
  <value>ip-172-31-68-129.ec2.internal</value>
  <description>http://wiki.apache.org/hadoop/Hive/HBaseIntegration</description>
</property>
```

That hostname is for my namenode.  This may not be a problem as the hive unit in this case has no hbase roles defined, but I haven't put "hive plus some hbase config even when hbase isnt there" to the test to see if anything is afoul.

jamesbeedy: @kwmonroe totally, nice catch. I'll set those bits appropriately.

PR #345, opened 2018-03-02
BIGTOP-3010: juju bundle refresh (feb 2018)

kwmonroe:
Charms were refreshed back when 1.2.1 was released, but the bundle revisions and READMEs in the upstream source were not updated.  This PR fixes that.


kwmonroe:
Picked up new changes for hive (#344) and zookeeper (#341).  We should be g2g, but I'll let CWR finish on these latest revs before merging upstream.

PR #348, opened 2018-03-21
BIGTOP-3015: hadoop-spark bundle correction

jamesbeedy:
This change removes the "hadoop-plugin" charm from the "hadoop-spark" bundles and tests as it isn't needed in the bundle.

Fixes: https://issues.apache.org/jira/browse/BIGTOP-3015

### DIFF
--- a/bigtop-deploy/juju/hadoop-spark/README.md
+++ b/bigtop-deploy/juju/hadoop-spark/README.md
@@ -48,14 +48,12 @@ follows:
   * Slave (DataNode and NodeManager) v2.7.3
     * 3 separate units
   * Spark (Driver in yarn-client mode) v2.1.1
-  * Client (Hadoop endpoint)
-    * Colocated on the Spark unit
   * Plugin (Facilitates communication with the Hadoop cluster)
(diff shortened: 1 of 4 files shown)


kwmonroe:
@jamesbeedy, the reason `hadoop-client` was in this bundle was to act as an endpoint for people that were accustomed to running jobs and managing hdfs from there.  iow, muscle memory and scripts that do this will work on any of the `hadoop-x` bundles::

```
juju run --unit hadoop-client/0 'hdfs dfs -cmd'
```

It was also there in case someone started with `hadoop-spark`, switched the spark runtime from YARN to HA, then scaled spark to x units.  In that scenario, the single `hadoop-client` would be a recognizable endpoint to facilitate hdfs administration, versus making users do that on a random spark unit.

That said, I'm coming around to your proposal to remove `hadoop-client`. Hulk smashing apps on a single unit has never been recommended and is only done to balance app density with resource cost.  As you already know, `spark` *is* a `hadoop-client` from the charms perspective, so an additional explicit client isn't needed.

I'm +1 to this removal if you'll also pull out the Client-related statements from the bundle README.md.

PR #351, opened 2018-04-17
Hadoop 3.1.0

acmnu:
Following actions was done to build 3.1.0:

Version bump to 3.1.0
Ant version bump
MAPREDUCE-6875 Rename mapred-site.xml.template to mapred-site.xml
HADOOP-13597 There is no kms-config.sh anymore
HADOOP-15122 patch has been removed because it had been merge in 3.1.0
cmake has been replaced with cmake3


acmnu:
@oflebbe, Please take a look at this.

bigtop-packages/src/rpm/hadoop/SPECS/hadoop.spec
```diff
@@ -623,7 +623,7 @@ fi
 #%config(noreplace) %{etc_hadoop}/conf.empty/mapred-site.xml
 %config(noreplace) %{etc_hadoop}/conf.empty/mapred-env.sh
 %config(noreplace) %{etc_hadoop}/conf.empty/mapred-queues.xml.template
-%config(noreplace) %{etc_hadoop}/conf.empty/mapred-site.xml.template
+%config(noreplace) %{etc_hadoop}/conf.empty/mapred-site.xml
```
evans-ye: Not sure why remove the template here. Is it because this is also renamed by upstream?

acmnu: That is upstream change. It was done in https://github.com/apache/hadoop/commit/f1de2c8560a4ce1a346727c2c265570c5b6b872e

acmnu: Commit and pull request description has been changed to clarify question above.

acmnu:
BTW, I am still in doubt about my solution with cmake3. It needs to bump other projects to understand result.

acmnu:
Seems no one care about it. So I close that one.

evans-ye:
@acmnu sorry got "distracted by work". Did you managed to get it work, or have customer using it?

PR #352, opened 2018-04-17
BIGTOP-3022 Auto find of latest Ant

acmnu:
That patch allow to derminate latest version of Ant
available on apache dist.

There is one restriction. We query main apache mirror instead
of nearest one because we can't be sure it use Apache with same
configuration as main.


acmnu:
@evans-ye, I changed function. It's support parameter now.

PR #358, opened 2018-05-18
BIGTOP-2101: ignite-hadoop contains an arch-dependent shared lib

JunHe77:
ignite-shmem contains x86 binary only. A build step is added to create
arch specific library for non-x86 platforms.

Change-Id: I9f4f03ec61250da21c13a915dc7efd5303e95d5b

### DIFF
--- a/bigtop-packages/src/common/ignite-hadoop/do-component-build
+++ b/bigtop-packages/src/common/ignite-hadoop/do-component-build
@@ -21,4 +21,19 @@ set -ex
 export MAVEN_OPTS="-Xmx512M"
 
 sed -i '/asm-all/{n;s/4.2/5.0.2/}' modules/hadoop/pom.xml
+
+cd ipc/shmem
+# patch and build it
+sed -i 's/\&ipcData->closed\,/(char *)\&ipcData->closed\,/' igniteshmem/org_apache_ignite_internal_util_ipc_shmem_IpcSharedMemoryUtils.cpp
+./configure && make
+# create layout
+mkdir -p META-INF/native/linux64
+cp ./igniteshmem/.libs/libigniteshmem.so META-INF/native/linux64/libigniteshmem.so
+# pack and install
+jar cf ignite-shmem-1.0.0.jar META-INF
+mvn install:install-file -Dfile=ignite-shmem-1.0.0.jar -DgroupId=org.gridgain -DartifactId=ignite-shmem -Dversion=1.0.0 -Dpackaging=jar -DcreateChecksum=true
+# clean up
+rm -rf META-INF
+cd -
+
 mvn clean install -DskipTests -Dhadoop.version=$HADOOP_VERSION -Dspark.version=$SPARK_VERSION -Dignite.edition=hadoop


bigtop-packages/src/common/ignite-hadoop/do-component-build
```diff
@@ -21,4 +21,30 @@ set -ex
...
 
 sed -i '/asm-all/{n;s/4.2/5.0.2/}' modules/hadoop/pom.xml
+
+if [ "$HOSTTYPE" != "x86_64" ] ; then
+    # download ignite-shmem from maven central
+    mvn dependency:get -DremoteRepositories=https://repo1.maven.org/maven2/ -DgroupId=org.gridgain -DartifactId=ignite-shmem -Dversion=1.0.0
+    mkdir tmp-build
+    cd tmp-build
+    # download source jar
+    wget https://repo1.maven.org/maven2/org/gridgain/ignite-shmem/1.0.0/ignite-shmem-1.0.0-sources.jar
```
evans-ye: Any chance we can evaluate the signature for this file purely obtained from internet?
It should be fairly easy since the version of jar is already fixed, the signature to check can be fixed here as well.

oflebbe: Right beside the file there is an .asc file. Evans, you may proposed the same approach as I did for ant in puppet code ?

"/usr/bin/gpg -v --verify --auto-key-retrieve --keyserver hkp://keyserver.ubuntu.com:80 $ant-bin.tar.gz.asc"

This would be rather lame, since we would test sources by comparing with a file retrieved from the same location. Situation is different for the ant scenario since we download from an untrusted party.

evans-ye: @oflebbe yup downloading from the same remote seems no much help to security. I think having the .asc file content hard-coded in our code base should be OK.

JunHe77: @evans-ye @oflebbe Would it be OK to verify both sha1 and md5 checksum of the jar file, and compare them with hardcoded in our code base?

evans-ye: That's what I'm proposing. It should be fairly easy to implement. If possible, let's try this and get this in.

bigtop-packages/src/common/ignite-hadoop/do-component-build
```diff
@@ -21,4 +21,39 @@ set -ex
...
+    # verify it
+    md5sum=$(/usr/bin/md5sum ignite-shmem-1.0.0-sources.jar | cut -d ' ' -f 1)
+    sha1sum=$(/usr/bin/sha1sum ignite-shmem-1.0.0-sources.jar | cut -d ' ' -f 1)
+    if [ $md5sum != $SAVED_MD5 ] || [ $sha1sum != $SAVED_SHA1 ]; then
+        echo "Downloaded file checksum doesn't match!"
```
evans-ye: I'd prefer to have ignite-shmem-1.0.0-sources.jar in the error message so that it's super clear what's going wrong :)

JunHe77: Got it and updated it.

oflebbe:
Hi @JunHe77 

Sorry to jump in late again, just being back from vacation.

MD5 sums are deprecated and insecure, please use sha1 only. Release policies at apache.org have been changed a while ago. 

I am not a big fan of architecture specifc code: Shouldn't we replace the shared libs for every architecture? That way we may not introduce arch specific bugs. What do you think?

JunHe77:
@evans-ye @oflebbe , updated the patch based on your comments and new findings:
1. the ignite-shmem source code is shipped with ignite itself, located in <IGNITE_ROOT>/ipc/shmem. I compared the source code with the one from ignite-shmem-1.0.0-sources.jar, they are identical.
2. Removed arch specific process. All platforms use the same way.

Pls review.

oflebbe:
I have a different proposal: 

create a temporary directory
Place attached pom.xml from 
[pom.txt](https://github.com/apache/bigtop/files/2349398/pom.txt)

run `mvn dependency:unpack-dependencies@src-dependencies`
and `mvn dependency:unpack-dependencies`

(First will download source and unpack in `target/sources` second line will download original jar)
run compile job.

Replace with .. TBC ...

JunHe77:
The shmem source is included in ignite, see: https://github.com/apache/ignite/tree/master/ipc/shmem
So once we get ignite code, we get shmem source.

bigtop-packages/src/common/ignite-hadoop/do-component-build
```diff
@@ -21,4 +21,23 @@ set -ex
 export MAVEN_OPTS="-Xmx512M"
 
 sed -i '/asm-all/{n;s/4.2/5.0.2/}' modules/hadoop/pom.xml
+
+# download ignite-shmem from maven central
+mvn dependency:get -DremoteRepositories=https://repo1.maven.org/maven2/ -DgroupId=org.gridgain -DartifactId=ignite-shmem -Dversion=1.0.0
```
oflebbe: I do not think we need this at all since we replace the only necessary file anyway.

bigtop-packages/src/common/ignite-hadoop/do-component-build
```diff
@@ -21,4 +21,23 @@ set -ex
...
+
+cd ipc/shmem
+# patch and build it
+sed -i 's/\&ipcData->closed\,/(char *)\&ipcData->closed\,/' igniteshmem/org_apache_ignite_internal_util_ipc_shmem_IpcSharedMemoryUtils.cpp
+./configure && make
+# unpack
+jar xf ~/.m2/repository/org/gridgain/ignite-shmem/1.0.0/ignite-shmem-1.0.0.jar
```
oflebbe: IMHO a mkdir -p META-INFA/native/linux64 would work here as well

bigtop-packages/src/common/ignite-hadoop/do-component-build
```diff
@@ -21,4 +21,23 @@ set -ex
...
+# replace
+cp ./igniteshmem/.libs/libigniteshmem.so META-INF/native/linux64/libigniteshmem.so
+# repack
+jar cf ignite-shmem-1.0.0.jar META-INF
+cp ignite-shmem-1.0.0.jar ~/.m2/repository/org/gridgain/ignite-shmem/1.0.0/ignite-shmem-1.0.0.jar
```
oflebbe: Please use `mvn install:install-file -Dfile=target/dependency/ignite-shmem-1.0.0.jar -DgroupId=org.gridgain -DartifactId=ignite-shmem -Dversion=1.0.0    -Dpackaging=jar -DcreateChecksum=true` instead.

It will create correct checksums in maven repo as well.

oflebbe:
OOps you are right,  thanks for the hint

JunHe77:
Updated.

evans-ye:
Excellent thread. I've learned a lot. Thank you both.

PR #369, opened 2018-06-25
BIGTOP-3046: correct script to find lasted mvn version

deart2k:


deart2k:
The current variant of the script returns not correct value, if try to run this script:
`curl --stderr /dev/null 'https://www.apache.org/dist/maven/maven-3/?F=0&V=1' | grep -o '<li>.*href="3.5.[0-9]*/"' | grep -o '3.5.[0-9]*'`
it's return two line:
`3.5.3`
`3.5.4`

acmnu:
Strange that original patch for Ant was right, but for some reason it was changed for Maven

JunHe77:
Thanks for reporting this issue, @deart2k !
The previous implementation assumes that only one minor version is listed in the url, say, only one 3.5.x. But it is obviously not true, as there are 3.5.3 and 3.5.4 on the link now. 😀 

Your patch LGTM, while I think there might be a little to add: it may not true to assume the last `<li>` is the latest version. So an extra cmd can be added to make sure the output is ordered by version, like following:
`curl --stderr /dev/null 'https://www.apache.org/dist/maven/maven-3/?F=0&V=1' | grep -o '<li>.*href="3.5.[0-9]*/"' | grep -o '3.5.[0-9]*' | sort -V --field-separator=- | tail -1 | tr -d '\n'`

What do you think?

And could you pls create a JIRA to link with this PR so we can track this, just like https://issues.apache.org/jira/browse/BIGTOP-3031.

Thank you very much.

acmnu:
@JunHe77, list is already sorted on server side (V=1 in GET parameters). I have no idea is this sorting equal to ```sort -n```, but authors of auto_index module called that "version sorting".

http://httpd.apache.org/docs/2.2/mod/mod_autoindex.html#page-header

deart2k:
@JunHe77 I create  issue: https://issues.apache.org/jira/browse/BIGTOP-3046

JunHe77:
@acmnu , I see. So the `sort` cmd is not needed.
@deart2k , great! Thank you for your efforts and contribution.
LGTM, +1.

PR #370, opened 2018-06-29
Expanded Hive Testing

bharatm363:
These tests validate the Hive JDBC driver and have fleshed out usage to include complex, generated SQL and meta-data testing.


evans-ye:
Hi @bharatm363 can you create a JIRA on BIGTOP and describe what feature you'd like to add?

bharatm363:
Sure @evans-ye, how can I go about doing that on https://issues.apache.org/jira/projects/BIGTOP/issues

bharatm363:
@evans-ye I have created an issue on JIRA for this testing, https://issues.apache.org/jira/browse/BIGTOP-3049, please let me know if that is correct.

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
```diff
@@ -0,0 +1,392 @@
...
+    final File f = new File(HiveJdbcGeneralTest.class.getProtectionDomain()
+        .getCodeSource().getLocation().getPath());
+    // String jdbcDriver = propertyValue("hive-site.xml",
+    // "javax.jdo.option.ConnectionDriverName");
+
+    // String qualifiedName = propertyValue("hdfs-site.xml",
+    // "dfs.internal.nameservices");
+    // String[] haNodes = propertyValue("hdfs-site.xml",
+    // "dfs.ha.namenodes."+qualifiedName).split(",");
+    // String primaryNode = haNodes[0];
```
evans-ye: Why don't we remove these code block?

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
```diff
@@ -0,0 +1,392 @@
...
+      executeStatement(stmt, "Drop view testview");
+      printResults(stmt, "Describe formatted btest");
+      dropTable(stmt, newTableName);
+      dropTable(stmt, newTableName + "NT");
+      dropTable(stmt, newTableName + "T");
+      dropTable(stmt, newTableName + "P");
+      dropTable(stmt, newTableName + "V");
+      setNegativeFetchSize(stmt);
+    }
+  }
```
evans-ye: It seems that we're testing many cases here. Can we separate them one by one? That way we can precisely get which part is failing. If the whole series of operation is what you'd like to test. Then we should rename the test name or add comments to describe what we're going to test here.

bigtop-tests/smoke-tests/hive/TestMethods.java
```diff
@@ -0,0 +1,331 @@
...
+      int length = nodeList.getLength();
+      for (int i = 0; i < length; i++) {
+        if (nodeList.item(i).getNodeType() == Node.ELEMENT_NODE) {
+          Element element = (Element) nodeList.item(i);
+          if (element.getNodeName().contains("property")) {
+            names.add(
+                element.getElementsByTagName("name").item(0).getTextContent());
+            values.add(
+                element.getElementsByTagName("value").item(0).getTextContent());
+
```
evans-ye: Not sure why additional newline here

bigtop-tests/smoke-tests/hive/TestMethods.java
```diff
@@ -0,0 +1,331 @@
...
+          }
+        }
+      }
+    }
+    String[] nameslist = names.toArray(new String[names.size()]);
+    String[] valueslist = values.toArray(new String[values.size()]);
```
evans-ye: Is namesList & valuesList better? Though a bit confusing cause they're array.

bharatm363:
Thanks for the comments, I'll look into these, I'm going to refactor it some to make everything more readable and address some of the points you raised above. @evans-ye

evans-ye:
Sure. Just want to let you know that it doesn't mean that everything I comment need to be addressed. If you think I'm wrong, reply with your thought.

evans-ye:
@bharatm363 can we have this in 1.3?

bharatm363:
@evans-ye Uploading some changes now, the main thing that I didn't directly address were the sql calls which are still performed in a large group. I thought that would be better as they all test the same kind of behavior without major differences between each one.

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
```diff
@@ -0,0 +1,390 @@
...
+ */
+public class HiveJdbcGeneralTest extends TestMethods {
+
+  static String hivePort = System.getenv("HIVE_PORT");
+  static String jdbcConnection = System.getenv("HIVE_JDBC_URL");
+  static Connection con;
+  String newTableName = "btest";
+
+  @BeforeClass
+  public static void onTimeSetup() throws Exception {
```
evans-ye: is this a typo? Should it be oneTimeSetup?

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
```diff
@@ -0,0 +1,390 @@
...
+      dropTable(stmt, "class");
+      dropTable(stmt, "test42");
+      stmt.executeUpdate(
+          "create table `class` (name varchar(8), sex varchar(1), age double precision, height double precision, weight double precision)");
+      stmt.executeUpdate(queryValues);
+      // specifying the location fails in hortonworks 3.0
+      try {
+        stmt.execute(locationQuery);
+      } catch (Exception e) {
+
```
evans-ye: Not quite sure what you'd like to test here. There's no assertion in this test.
Perhaps asserting an exception?

bharatm363: Yep, the location should be invalid as a new directory has been created. Asserting that the location is invalid.

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
```diff
@@ -0,0 +1,390 @@
...
+        + " ('Robert', 'M', 68.3, NULL, 'BJH', 2),"
+        + " ('Thomas', 'M', 57.5, 85.0, 'AJH', 1),"
+        + " ('Thomas', 'M', 59.1, 92.3, 'AJH', 2),"
+        + " ('Wakana', 'F', 61.3, 99.0, 'AJH', 1),"
+        + " ('Wakana', 'F', 63.8, 102.9, 'AJH', 2),"
+        + " ('William', 'M', 66.5, 112.0, 'BJH', 1),"
+        + " ('William', 'M', 68.3, 118.2, 'BJH', 2)";
+
+    // failing query from SAS Institute, code generated by PROC MEANS
+    String failingQuery =
```
evans-ye: What do you mean by failingQuery? Arn't we supposed to be testing positive cases here?

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
```diff
@@ -0,0 +1,390 @@
...
+          "(Flight int, Dates varchar(255), Depart varchar(10), Orig varchar(5), Dest varchar(5), Miles int, Boarded int, Capacity double)";
+      String partitionedColumns =
+          "(Flight int, Dates varchar(255), Depart varchar(10), Orig varchar(5), Dest varchar(5), Miles int, Boarded int)";
+      String localFilepath = f + "/samdat1.csv";
+      String HdfsURI = "hdfs://" + hdfsConnection;
+      String filePath = "/tmp/htest/00000_";
+      String fileDestination = HdfsURI + filePath;
+      assertFalse(con.getMetaData().supportsRefCursors());
+      assertTrue(con.getMetaData().allTablesAreSelectable());
+      assertEquals("Apache Hive", con.getMetaData().getDatabaseProductName());
```
evans-ye: It's OK to assert versioning and configurations. Just move these out from testTableCreation to another test and name it clearly.

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
```diff
@@ -0,0 +1,390 @@
...
+      System.out.println(
+          "Hive Version: " + con.getMetaData().getDatabaseMajorVersion() + "."
+              + con.getMetaData().getDatabaseMinorVersion());
```
evans-ye: It's OK to assert versioning and configurations. Just move these out from testTableCreation to another test and name it clearly.

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
```diff
@@ -0,0 +1,390 @@
...
+          String section = res2.getString(1);
+          if (section.contains("Detailed"))
+            break;
+        }
+        boolean found = false;
+        while (res2.next()) {
+          String value = res2.getString(2);
+          if (value != null && value.contains("SASFMT"))
+
+          {
```
evans-ye: The style here is broken. Reformat using IDE helps.

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
bharatm363: Should be, but in this case the query is failing for some reason, I believe this to be a regression in Hive 3.0 Beta

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
bharatm363: Done

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
bharatm363: Fixed

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
evans-ye: From Bigtop point of view we are testing against the release version, which is 2.3.3. If user want's to test other version of Hive, that's very much reasonable. 
In this example, the test should be passed in 2.3.3(I'm not sure though). If we expect it should be passed in 3.0 as well, then we should write the test as a normal one which assert that it should pass. 
Failing is another story that developer need to deal with, which don't need to encode in the test.
Do you think this is reasonable?

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
bharatm363: If no exception is thrown then there is no way to know which directory the file was written to (given that transactions now result in a more complex and verbose file strucutre)

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
bharatm363: You are correct, in 2.3.3 the test does pass correctly, I can ammend the comment to reference version 3.0 but leave the test case in. Using the current Bigtop Harness the test will pass, but when the release version becomes 3.0 I anticipate there will be a failure.

bharatm363:
@evans-ye In terms of squashing I'm not sure I have the access to do so since the pull request is in the apache repo. Would you be able to squash the commits in this pull request?

evans-ye:
@bharatm363 sorry to answer back late. I was on a week long vacation ;)
I know there's a squash button in Github, however since github is just only a mirror of Bigtop codebase, we don't use it more than reviewing the code.

Usually I'll just rebase my folked branch which creates the PR. Afterwards the changes shall be reflected.

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
evans-ye: Isn't that what we want? A failure raised telling us something different when we bump up the version.

bigtop-tests/smoke-tests/hive/HiveJdbcGeneralTest.java
evans-ye: Hey @bharatm363  sorry I didn't get your point. Probably because I didn't ask clearly.
Let me rephrase my question. For this test case: testSpecifiedLocation, you'd like to test a statement that creates table with location specified. So the result is either:
a) the statement success
b) the statement failed
Now, we need to know whether the result is a or b. But here you catch the exception with no-op. So to my knowledge either a or b this test is marked success. Am I correct?

PR #372, opened 2018-07-03
BIGTOP-3047 - Adding Zookeeper nagios checks

tbaumann:
Adding nagios checks  using the nrpe-external-master and local-monitors interfaces.

The test helper is from https://github.com/andreisavu/zookeeper-monitoring/
This test helper in the current form does not fail if zookeeper servers
are not available.

Requires latest upstream versions of nrpe-external-master and local-monitors

Implementing: https://issues.apache.org/jira/browse/BIGTOP-3047
Bigtop issue: BIGTOP-3047


tbaumann:
I don't know what that Jenkins error means. Anything I can do?

PR #374, opened 2018-07-16
Add support ubuntu 18.04

epikhinm:
Hi!

You are deleted support of ubuntu 14.04 LTS before end of life, soo i've added support ubuntu 18.04 LTS for amd64.
At ubuntu 18.04 we only have puppet 5.3 and we should add old directory /etc/puppet/modules as modulepath.
Pls, check this out. Feel free for feedback.


JunHe77:
I'll try this patch and see if the bigtop packages work on Ubuntu-18.04.

epikhinm:
Nice, could you show me how are you testing?
Today documentation is lean, or maybe i doesn't fully understood it

evans-ye:
Basically, you need to first get packages built:
https://cwiki.apache.org/confluence/display/BIGTOP/How+to+build+Bigtop-trunk

And then deploy:
https://cwiki.apache.org/confluence/display/BIGTOP/Bigtop+Provisioner+User+Guide

If you hit any problem, feel free to post it on dev@bigtop.apache.org

bigtop_toolchain/bin/puppetize.sh
```diff
@@ -26,7 +26,7 @@ case ${ID}-${VERSION_ID} in
       dnf -y check-update
       dnf -y install hostname findutils curl sudo unzip wget puppet puppetlabs-stdlib
         ;;
-    ubuntu-16.04)
+    ubuntu-1[6,8].04)
```
JunHe77: Would be better to create another entry for 18.04 as there are different packages to be installed.

bigtop_toolchain/bin/puppetize.sh
```diff
@@ -26,7 +26,7 @@ case ${ID}-${VERSION_ID} in
...
         apt-get update
         apt-get -y install wget curl sudo unzip puppet software-properties-common puppet-module-puppetlabs-apt puppet-module-puppetlabs-stdlib
```
JunHe77: On ubuntu-18.04 two extra packages need to be installed to get successful bigtop build:
rsync, dh-python

epikhinm: Hmm, sounds strange. I've builded bigtop and some packages (hadoop, hive, hbase,  and zookeeper) without these dependencies.

But ok, i'll add your opinion as a patch.

I also have small troubles with systemd on ubuntu-18.04, standard "sysvinit compatibility" doesn't work properly. In the current initd scripts we creating lock directories, pid files. Systemd expect than file have same pair owner-group as a main process, but it don't.

I wrote new template for systemd services, but doesn't debugged it yet.

Give me a couple of days for this patches :)

PR #382, opened 2018-08-20
Bigtop-3070 support Spark Thrift Server in Charm

buggtb:
The thrift server has no exposure and should be configurable for Spark SQL support etc.

### DIFF
--- a/bigtop-packages/src/charm/spark/layer-spark/lib/charms/layer/bigtop_spark.py
+++ b/bigtop-packages/src/charm/spark/layer-spark/lib/charms/layer/bigtop_spark.py
@@ -400,6 +400,10 @@ class Spark(object):
             else:
                 hookenv.log("Spark Worker did not start")
 
+        # Open port for Thrift Server
+        if hookenv.config()['spark_enable_thriftserver']:
+            hookenv.open_port('10000')
+
     def stop(self):
         """
         Stop all services (and close associated ports). Stopping a service
--- a/bigtop-packages/src/charm/spark/layer-spark/reactive/spark.py
+++ b/bigtop-packages/src/charm/spark/layer-spark/reactive/spark.py
@@ -235,7 +235,8 @@ def reconfigure_spark():
     # Almost all config changes should trigger a reinstall... except when
     # changing the bigtop repo version. Repo version changes require the user
     # to run an action, so we skip the reinstall in that case.
-    if not is_state('config.changed.bigtop_version'):
+    if not is_state('config.changed.bigtop_version') \
+            and not is_state('config.changed.spark_enable_thriftserver'):
         # Config changes should reinstall even if the deployment topology has
         # not changed. Hence, pass force=True.
         reinstall_spark(force=True)
--- a/bigtop-packages/src/charm/spark/layer-spark/reactive/spark.py
+++ b/bigtop-packages/src/charm/spark/layer-spark/reactive/spark.py
@@ -21,7 +21,7 @@ from charmhelpers.core import hookenv, host, unitdata
 from charms import leadership
 from charms.reactive.helpers import data_changed
 from jujubigdata import utils
-
+from subprocess import check_call
 
 ###############################################################################
 # Status methods
@@ -138,7 +138,6 @@ def set_deployment_mode_state(state):
     # set app version string for juju status output
     spark_version = get_package_version('spark-core') or 'unknown'
     hookenv.application_version_set(spark_version)
-
 
 ###############################################################################
 # Reactive methods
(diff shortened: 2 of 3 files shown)


bigtop-packages/src/charm/spark/layer-spark/reactive/spark.py
```diff
@@ -236,7 +235,8 @@ def reconfigure_spark():
     # Almost all config changes should trigger a reinstall... except when
     # changing the bigtop repo version. Repo version changes require the user
     # to run an action, so we skip the reinstall in that case.
-    if not is_state('config.changed.bigtop_version'):
+    if not is_state('config.changed.bigtop_version') \
+            and not is_state('config.changed.spark_enable_thriftserver'):
```
kwmonroe: I'm a bit unsettled with this.  If someone changed multiple config options with something like `juju config spark driver_memory=2048M spark_enable_thriftserver=true`, we'd miss the driver_mem change.

What if we adjusted the `reconfigure_spark` function to be:
```
@when('spark.started', 'config.changed')
@when_not('config.changed.bigtop_version', 'config.changed.spark_enable_thriftserver')
 def reconfigure_spark():
...
```

We could then remove the check for `config.changed.bigtop_version` in the `reconfigure_spark` handler (it gets handled anyway with the `check_repo_version` function a few lines down). Then we could adjust your `start_thrift` function to be:
```
@when('spark.started', 'config.changed.spark_enable_thriftserver')
def start_thrift():
    enable_thrift = hookenv.config()['spark_enable_thriftserver']
    if enable_thrift:
        check_call(['/usr/lib/spark/sbin/start-thriftserver.sh'])
    else:
        check_call(['/usr/lib/spark/sbin/stop-thriftserver.sh'])
```

kwmonroe:
One other thing @buggtb, since this opens a new port, we'll need 10000 listed in layer.yaml:

https://github.com/apache/bigtop/blob/master/bigtop-packages/src/charm/spark/layer-spark/layer.yaml#L25

As well as a check for `spark_enable_thriftserver` in the `start` method to know if we should run `hookenv.open_port`, similar to what we do for opening the spark-master-ui port:

https://github.com/apache/bigtop/blob/master/bigtop-packages/src/charm/spark/layer-spark/lib/charms/layer/bigtop_spark.py#L377

bigtop-packages/src/charm/spark/layer-spark/reactive/spark.py
kwmonroe: @buggtb, danger!  this isn't going to work and the mighty @johnsca has confirmed the error of my ways.  Since the handler is a catch-all with 'config.changed', if something is set in addition to `spark_enable_thrift`, we'll skip the handler because of the @when_not -- basically forget my suggestion to `reconfigure_spark`.  We'll need to keep what you had to process config_changed, skipping the reinstall if our bigtop_version or spark_enable_thrift are set.

Still, I think the `start_thrift` method would nice to refactor to run @when(spark.started, spark_thrift_enabled) and drop the data_changed logic.

buggtb: "Still, I think the start_thrift method would be nice to refactor to run @when(spark.started, spark_thrift_enabled) and drop the data_changed logic." < can you react to a specific config change? I've not noticed that pattern in a charm before.

I see elsewhere in that charm bigtop.version.changed, but then it doesn't use that config option to look anything up.

bigtop-packages/src/charm/spark/layer-spark/reactive/spark.py
```diff
@@ -318,3 +318,10 @@ def client_present(client):
 def client_should_stop(client):
     if is_state('leadership.is_leader'):
         client.clear_spark_started()
+
+@when('spark.started', 'config.changed.spark_enable_thriftserver')
```
buggtb: Having read @johnsca 's doc I think this is correct. LMK @kwmonroe

bigtop-packages/src/charm/spark/layer-spark/lib/charms/layer/bigtop_spark.py
```diff
@@ -400,6 +400,10 @@ def start(self):
             else:
                 hookenv.log("Spark Worker did not start")
 
+        # Open port for Thrift Server
+        if hookenv.config()['spark_enable_thriftserver']:
```
buggtb: @kwmonroe I stuck this in here, but then realised that we don't cycle spark on a thrift server start, so I'm tempted to just stick it in with the start/stop logic for the thrift stuff. WDYT?

bigtop-packages/src/charm/spark/layer-spark/reactive/spark.py
kwmonroe: > can you react to a specific config change

yup, you're doing it right :)

bigtop-packages/src/charm/spark/layer-spark/reactive/spark.py
kwmonroe: +1

bigtop-packages/src/charm/spark/layer-spark/reactive/spark.py
```diff
@@ -318,3 +318,10 @@ def client_present(client):
...
+def start_thrift():
```
kwmonroe: add those open/close port bits here

PR #390, opened 2018-09-05
BIGTOP-3077: OpenSuse-42.3 packages install error

JunHe77:
Update libapr1 to use distro version. Fixed packages dependency issue
on aarch64.

Change-Id: I043b701b4bd5b89eea7488e21ace939624528310


JunHe77:
On aarch64, these libraries require libncurses5-5.9-57.1.aarch64, while the distro installed is newer libncurses5-5.9-62.1.aarch64, so it has to be specified with "force-resolution" to downgrade libncurses5 to proceed installation.

PR #393, opened 2018-09-11
BIGTOP-3080. New shell script zkTxnLogToolkit.sh should be added to the install script

anmolnar:
I added `TxnLogToolkit` invocation script to Zk install script.


anmolnar:
Anyone?

evans-ye:
@anmolnar Please see my comment in https://issues.apache.org/jira/browse/BIGTOP-3080. Let's make sure we're aligned on the same page before we proceed the PR.

anmolnar:
@evans-ye Sorry, I didn't notice that. Let me check.

evans-ye:
@anmolnar zookeeper is upgrade to 3.4.13. Maybe the tool is available now. Could you check with it?

PR #401, opened 2018-10-03
BIGTOP-3088: provisioner failed to run puppet deployment on CentOS-7

JunHe77:
Deploy with puppet on centos-7 failed.
The root cause is puppet-stdlib (4.25) in distro is not compatible with
puppet <4. Pin stdlib to 4.12.0 should fix this problem.

Change-Id: I8ec557412b660cc51399da3c81014b47aa366d6a

### DIFF
--- a/bigtop_toolchain/bin/puppetize.sh
+++ b/bigtop_toolchain/bin/puppetize.sh
@@ -40,7 +40,12 @@ case ${ID}-${VERSION_ID} in
         ;;
     centos-7*)
         rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-        yum -y install hostname curl sudo unzip wget puppet puppetlabs-stdlib
+        yum updateinfo
+        # BIGTOP-3088: pin puppetlabs-stdlib to 4.12.0 as the one provided by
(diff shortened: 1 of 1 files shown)


bigtop_toolchain/bin/puppetize.sh
```diff
@@ -40,7 +40,9 @@ case ${ID}-${VERSION_ID} in
         ;;
     centos-7*)
         rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-        yum -y install hostname curl sudo unzip wget puppet puppetlabs-stdlib
+        yum updateinfo
+        yum -y install hostname curl sudo unzip wget puppet
+        puppet module install puppetlabs-stdlib --version 4.12.0
```
evans-ye: Mind to add a comment here to link to BIGTOP-3088? 
This tells future developer that this line is for special case and can be removed if obstacle removed.

JunHe77: Thanks. The suggestion reaaly helps. I used to run into similar case without knowing the code block's context.
Will update soon.

JunHe77:
Close this as it has been merged by upstream.

PR #402, opened 2018-10-03
BIGTOP-3089: Update provision config files with changes in 1.3.0 release

JunHe77:
v1.3.0 supports: centos-7, fedora-26, debian-9, ubuntu-16.04, opensuse-42.3
config files in provisioner/docker should be updated to sync with these changes

Change-Id: Ic5d66cb837e78908fcf19c2e55f819db28866a55


provisioner/docker/config_centos-7.yaml
```diff
@@ -14,10 +14,10 @@
 # limitations under the License.
 
 docker:
-        memory_limit: "4g"
-        image: "bigtop/puppet:trunk-centos-7"
+        memory_limit: "8g"
```
evans-ye: Let's stick with 4g here.

provisioner/docker/config_fedora-26.yaml
```diff
@@ -14,10 +14,10 @@
 # limitations under the License.
 
 docker:
-        memory_limit: "4g"
-        image: "bigtop/puppet:trunk-centos-6"
+        memory_limit: "8g"
+        image: "bigtop/puppet:1.3.0-fedora-26"
 
-repo: "http://repos.bigtop.apache.org/releases/1.2.1/centos/6/x86_64"
+repo: "http://repos.bigtop.apache.org/releases/1.3.0/fedora/26/$basearch"
```
evans-ye: I can't figure out why some are $basearch and some are still x86_64. What's the rule here?

JunHe77: rpm base distros support using $basearch to automatically load arch dependent repo:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-using_yum_variables
while for apt it doesn't use this way.

The target is to unified configurations/deployment across different architectures for a same distro. So this patch starts from rpm based distros (centos/fedora/opensuse)

epikhinm: Debian-based distro's support $(ARCH) variable https://manpages.debian.org/jessie/apt/sources.list.5.en.html

JunHe77: @epikhinm that's a good point. I didn't realize $ARCH can be used in repo url. Just thought I can only specify what kind of arch a certain repo supports, lie "arch=arch1,arch2".
Thanks for the suggestion and I will test and update new patch if it is OK for auto loading.

JunHe77: Using $ARCH to automatically load different architecture repo on deb systems is verified on x86_64, aarch64 and ppc64le.
Patch is updated. "docker-hadoop.sh" is modified to load correct puppet image by detecting CPU arch. This makes only one config file is needed for a certain distro acrossing different architectures.
Pls review. Thanks.

PR #403, opened 2018-10-03
BIGTOP-3090: provisioner failed on fedora-26 when deploying jdk

JunHe77:
The puppet recipe of jdk intends to setup cacerts link with pki/java/cacerts
while jdk is not installed. This security failure will cause later bigtop
components installation failure.

Change-Id: Ib728cdf60db9feaddb3f65643ee2134158947254


JunHe77:
Close this as it has been merged by upstream.

PR #434, opened 2018-12-25
BIGTOP-3118. [Puppet] Failed to deploy GPDB

evans-ye:


oflebbe:
I am pretty sure usr/bin/pip-python symlink is of no specifc use for rh / centis / fedora. may i suggest to remove the dependecy and this file and replace by the package python2-pip ?

is there really no dependency on python-devel on gcc for this platform as well, so we need to fix that ourselfs?

PR #442, opened 2019-01-14
BIGTOP-3132. Bump Alluxio up to 1.8.1

evans-ye:
* Change to sleep time to 5 seconds to so that alluxio-worker has sufficient time to startup
* Rearrange conf dir layout to conf.dist to align bigtop convention.
* link back conf and logs into /usr/lib/alluxio to align the convention of alluxio by default
* Fix several debian build bugs:
  * missing var dir because of an additional space
  * missing pre and post inst script
  * maintainer should be bigtop


oflebbe:
Looks great at first sight! 

Timeout (Sleep Statements) are problematic. Do we really dont't have any hint (any file , or open port) that the service is available ? (we could in a loop for instance with netcat that a specific service port is now available )

evans-ye:
@oflebbe That seems to be a simple fix we can approach for. Let's open up another JIRA for it.

PR #448, opened 2019-01-16
BIGTOP-3138. Maven installation failed on debian-9

evans-ye:


oflebbe:
LGTM +1. I never tested this interactively

PR #458, opened 2019-01-30
BIGTOP-3150. [Provisioner] Raise proper exit code and error messages when provisioning failed

evans-ye:


oflebbe:
Nice way to store and retrieve error codes.
LGTM +1

evans-ye:
Thanks for the review.

PR #462, opened 2019-02-03
BIGTOP-3153. Switch to docker cp based solution to build inside containers

evans-ye:


oflebbe:
Finally I see what you did propose

evans-ye:
Yeah. Basically is using "docker cp" to copy the bigtop repo into container and build inside.
After built copy the result out as what you did before.

PR #467, opened 2019-02-13
BIGTOP-3161. Upgrade to Puppet 5.X

evans-ye:


oflebbe:
does this work on ppc64 and aarch64 /arm64?

last time I tried there was no support for these archs from puppetlabs

evans-ye:
Good point. Let me try ;)

PR #468, opened 2019-02-13
BIGTOP-3166. Update Kerberos module to comply with Puppet's syntax

evans-ye:


evans-ye:
Thanks for the review @oflebbe

PR #470, opened 2019-02-14
BIGTOP-3145: Crunch smoke tests implementation

guyuqi:
Add Crunch smoke tests.

### DIFF
--- a/bigtop-tests/smoke-tests/crunch/TestCrunchSmoke.groovy
+++ b/bigtop-tests/smoke-tests/crunch/TestCrunchSmoke.groovy
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


bigtop-tests/smoke-tests/crunch/TestCrunchSmoke.groovy
```diff
@@ -0,0 +1,84 @@
...
+
+import org.apache.bigtop.itest.shell.Shell
+import org.apache.bigtop.itest.TestUtils
+
+public class TestCrunchSmoke {
+  static String runnerScript = "hadoop jar"
+
+  static String crunchJar = System.getProperty(
+    "org.apache.bigtop.itest.crunch.smoke.crunch.jar",
+    "/usr/share/doc/crunch/crunch-examples-*job.jar");
```
evans-ye: Is this the right path to get example jar?

guyuqi: Yes, refer to: [install_crunch.sh](https://github.com/apache/bigtop/blob/master/bigtop-packages/src/common/crunch/install_crunch.sh#L96-L98)

guyuqi: I tested it on Ubuntu 16.04, and then launched this PR.
When I tested it based on CentOS-7 and fedora-26 today,  the puppet can not deploy rpm correctly.

`diff --git a/provisioner/docker/config_centos-7.yaml b/provisioner/docker/config_centos-7.yaml
index 49f86ae..aabd6b9 100644
--- a/provisioner/docker/config_centos-7.yaml
+++ b/provisioner/docker/config_centos-7.yaml
@@ -19,6 +19,6 @@ docker:

 repo: "http://repos.bigtop.apache.org/releases/1.3.0/centos/7/$basearch"
 distro: centos
-components: [hdfs, yarn, mapreduce]
-enable_local_repo: false
-smoke_test_components: [hdfs, yarn, mapreduce]
+components: [hdfs, yarn, mapreduce, crunch]
+enable_local_repo: true
+smoke_test_components: [crunch]
` 
The rpm is not from local though 'enable_local_repo' is true.

evans-ye: I reach the same conclusion. We need to fix the packaging first...

PR #477, opened 2019-02-17
BIGTOP-3171. Update Kafka Puppet module for version 1.1.1

evans-ye:


bigtop-deploy/puppet/modules/kafka/templates/server.properties
```diff
@@ -125,6 +125,15 @@ log.retention.check.interval.ms=60000
 # If log.cleaner.enable=true is set the cleaner will be enabled and individual logs can then be marked for log compaction.
 log.cleaner.enable=false
 
+# The replication factor for the offsets topic (set higher to ensure availability).
+# Internal topic creation will fail until the cluster size meets this replication factor requirement.
+<% if @offsets_topic_replication_factor.nil? -%>
+# Set to 1 for less than 3 nodes deployment to work.
+offsets.topic.replication.factor=1
+<% else -%>
+offsets.topic.replication.factor=<%= @offsets_topic_replication_factor %>
```
oflebbe: All this stuff should be handled in init.pp IMO .The templating should boil down to 


`offsets.topic.replication.factor=<%= @offsets_topic_replication_factor %>`

evans-ye: Great idea. Let me rework!

evans-ye:
Fixed. Please review @oflebbe

bigtop-deploy/puppet/modules/kafka/templates/server.properties
oflebbe: Great. LGTM +1 (Tiny remark : Do we really need the <% if @offsets_topic_replication_factor.nil? -%> in template, since there may be now way to set that value to nil ? And if it's nil == not sent the the file it is using the default value 1 anyway ?, you may disregard this comment)

evans-ye:
You're right! I've remove that if-else clause. Thanks for the snd. eye!

PR #518, opened 2019-04-14
BIGTOP-3206. Missing /etc/init.d/functions when deploying Alluxio on Fedora 26

evans-ye:


evans-ye:
Thanks for the review @oflebbe I do not see lsb-support in our code base. Are you suggesting that there's a better way to fix this via setting dependency in packages?

oflebbe:
It is this dependency 
Requires: /lib/lsb/init-functions

evans-ye:
I should fix this via dependencies. Let me rework on it. Thanks @oflebbe

PR #522, opened 2019-04-19
BIGTOP-3204. Bump Flink to 1.6.4

docularxu:
Bump Flink version from 1.4.2 to 1.6.4. There are three changes involved:
    1. obviously, version number
    2. in Flink 1.6.4, there is no 'resources' folder any more. Removed it.
    3. set mode to 'legacy'
    
    Starting from Flink v1.5, there is a rewrite of Flink’s deployment and process
    model (internally known as FLIP-6). A new option 'mode' was introduced to
    indicate which executions to use. However, the way Bigtop puppet deploys
    'flink-jobmanager' and 'flink-taskmanager', ie. calling 'flink-daemon.sh
    jobmanager' and 'flink-daemon.sh taskmanager', is legacy. Without setting
    mode to 'legacy', flink's built-in exmaples fail with error
    "JobSubmissionException: Failed to submit JobGraph"

### DIFF
--- a/bigtop-deploy/puppet/modules/flink/templates/flink-conf.yaml
+++ b/bigtop-deploy/puppet/modules/flink/templates/flink-conf.yaml
@@ -26,6 +26,14 @@ jobmanager.web.port: <%= @ui_port %>
 taskmanager.tmp.dirs: <%= @storage_dirs.join(":") %>
 <% end %>
 
+# Starting from Flink v1.5, there is a rewrite of Flink’s deployment and process
+# model (internally known as FLIP-6). A new option 'mode' was introduced to
+# indicate which executions to use. However, the way Bigtop puppet deploys
+# 'flink-jobmanager' and 'flink-taskmanager', ie. calling 'flink-daemon.sh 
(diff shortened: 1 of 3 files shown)


evans-ye:
@docularxu the patch is great. Just one thing: I see 3 commits in this patch. Please make this PR clean with only one commit containing the fix.
And, regarding to the Kafka reverts, are you seeing problems to run Flink 1.6 with Kafka 2.X?

docularxu:
> @docularxu the patch is great. Just one thing: I see 3 commits in this patch. Please make this PR clean with only one commit containing the fix.

Fixed that.

> And, regarding to the Kafka reverts, are you seeing problems to run Flink 1.6 with Kafka 2.X?

I didn't test that yet. What I tested is Flink 1.6 with Kafka 0.10.x version.

Problem of Kafka 2.x is its kafka-server doesn't launch since in bigtop we use zookeeper 3.4.6 which misses this class: (java.lang.NoClassDefFoundError: org/apache/zookeeper/AsyncCallback$MultiCallback)

PR #523, opened 2019-04-22
BIGTOP-3210. Missing /lib/lsb/init-functions when deploying Flink on Fedora 26

evans-ye:


oflebbe:
The patch looks surprisingly easy . LGTM +1

evans-ye:
@oflebbe Thanks for the review.

PR #537, opened 2019-07-12
BIGTOP-3168: Kafka smoke tests implementation

guyuqi:
Add Kafka smoke tests. Kafka depends on ZooKeeper.

Smoke test yaml:
```
components: [zookeeper, kafka]
enable_local_repo: true
smoke_test_components: [kafka]
```

### DIFF
--- a/bigtop-tests/smoke-tests/kafka/TestKafkaSmoke.groovy
+++ b/bigtop-tests/smoke-tests/kafka/TestKafkaSmoke.groovy
@@ -32,7 +32,34 @@ class TestKafkaSmoke {
   static Shell sh = new Shell("/bin/bash -s");
 
   static final String KAFKA_HOME = "/usr/lib/kafka"
-  static final String KAFKA_TOPICS = KAFKA_HOME + "/bin/kafka-topics.sh"
+  static final String KAFKA_CONFIG = KAFKA_HOME + "/config/server.properties "
+  static final String KAFKA_TOPICS = KAFKA_HOME + "/bin/kafka-topics.sh "
+  static final String KAFKA_SERVER_START = KAFKA_HOME + "/bin/kafka-server-start.sh "
+  static final String KAFKA_SERVER_STOP = KAFKA_HOME + "/bin/kafka-server-stop.sh "
+
+  @BeforeClass
+  static void kafkaSetUp() {
+    /* Restart kafka server with new broker id 111
+     * Enable delete.topic.enable
+     */
+    sh.exec(KAFKA_SERVER_STOP);
+    sh.exec(KAFKA_SERVER_START + KAFKA_CONFIG
+      + " --override delete.topic.enable=true"
+      + " --override broker.id=111"
+      + " --override port=9192"
+      + " --override log.dirs=/tmp/kafka-logs-111 &"
+    );
+    assertTrue(" Restart Kafka server failed. " + sh.getOut() + " " + sh.getErr(), sh.getRet() == 0);
+  }
+
+  @AfterClass
+  public static void deleteKafkaTopics() {
+    sh.exec(KAFKA_TOPICS
+      + " --zookeeper localhost:2181"
+      + " --delete --topic test"
+    );
+    assertTrue(" Delete Kafka topics failed. " + sh.getOut() + " " + sh.getErr(), sh.getRet() == 0);
+  }
 
   @Test
   public void testCreateTopics() {
--- a/bigtop-tests/smoke-tests/kafka/TestKafkaSmoke.groovy
+++ b/bigtop-tests/smoke-tests/kafka/TestKafkaSmoke.groovy
@@ -1,0 +1,50 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


evans-ye:
Great to see this PR. Let me look into it tonight

bigtop-tests/smoke-tests/kafka/TestKafkaSmoke.groovy
```diff
@@ -0,0 +1,50 @@
...
+  static final String ZK_SERVER = ZOOKEEPER_HOME + "/bin/zkServer.sh"
+
+  static final String KAFKA_HOME = "/usr/lib/kafka"
+  static final String KAFKA_CONFIG = KAFKA_HOME + "/config/server.properties"
+  static final String KAFKA_TOPICS = KAFKA_HOME + "/bin/kafka-topics.sh"
+  static final String KAFKA_SERVER_START = KAFKA_HOME + "/bin/kafka-server-start.sh"
+
+  @Test
+  public void testCreateTopics() {
+    sh.exec(ZK_SERVER + " start");
```
evans-ye: During the test we should not change the state of the environment. I think the ZK server should already started.

bigtop-tests/smoke-tests/kafka/TestKafkaSmoke.groovy
```diff
@@ -0,0 +1,50 @@
...
+    sh.exec(KAFKA_SERVER_START + KAFKA_CONFIG + " &");
+    sh.exec(KAFKA_TOPICS + " --create --zookeeper localhost:2181 --replication-factor 1 --partitions 1 --topic test");
+    sh.exec(KAFKA_TOPICS + " --list --zookeeper localhost:2181");
+    assertTrue(" Create Kafka topics failed. " + sh.getOut() + " " + sh.getErr(), sh.getRet() == 0);
```
evans-ye: We should remove the test topic otherwise this test can only success the first time.

bigtop-tests/smoke-tests/kafka/TestKafkaSmoke.groovy
guyuqi: You're correct.  I'll remove it.

bigtop-tests/smoke-tests/kafka/TestKafkaSmoke.groovy
guyuqi: The smoke tests are taken by docker provisioner.  Smoke test routine is launched in dockers which will be  destroyed (--destroy) at the end of test process. 
In my understanding, smoke test would run only once for each test trigger.

evans-ye: Sorry for late reply.
Docker Provisioner integration is just one way to consume the smoke tests. Some other users might consume the test on a dev cluster that's not provisioned by us. Furthermore when troubleshooting real problems, smoke test may be re-executed until the test passed. So it's important to be designed atomic. Do you think this is reasonable?

guyuqi: It makes sense.  Really thanks for your patient explanation.

bigtop-tests/smoke-tests/kafka/TestKafkaSmoke.groovy
```diff
@@ -32,7 +32,34 @@ class TestKafkaSmoke {
...
   static final String KAFKA_HOME = "/usr/lib/kafka"
-  static final String KAFKA_TOPICS = KAFKA_HOME + "/bin/kafka-topics.sh"
+  static final String KAFKA_CONFIG = KAFKA_HOME + "/config/server.properties "
+  static final String KAFKA_TOPICS = KAFKA_HOME + "/bin/kafka-topics.sh "
+  static final String KAFKA_SERVER_START = KAFKA_HOME + "/bin/kafka-server-start.sh "
+  static final String KAFKA_SERVER_STOP = KAFKA_HOME + "/bin/kafka-server-stop.sh "
+
+  @BeforeClass
+  static void kafkaSetUp() {
+    /* Restart kafka server with new broker id 111
```
evans-ye: I'm curious why restart Kafka with broker id 111?

bigtop-tests/smoke-tests/kafka/TestKafkaSmoke.groovy
```diff
@@ -0,0 +1,70 @@
...
+     * Enable delete.topic.enable
+     */
+    sh.exec(KAFKA_SERVER_STOP);
+    sh.exec(KAFKA_SERVER_START + KAFKA_CONFIG
+      + " --override delete.topic.enable=true"
+      + " --override broker.id=111"
+      + " --override port=9192"
+      + " --override log.dirs=/tmp/kafka-logs-111 &"
```
evans-ye: I'm a little bit worry about this because it changes the broker settings. What if there's already a broker 111 exists in the testing cluster? 

Let's not jump into development directly and discuss of what's the best way to do it first. Maybe a better way to test Kafka is to make test topic idempotent(i.e. either create it or it's already exists). And then we test the produce and consume functionality instead?

If  there's no easy way to implement it another way. Can we just enable delete.topic.enable on the fly and don't change the broker settings?

guyuqi: I have removed the new broker and just enable `delete.topic.enable`. :)

evans-ye: Yup. That's good. But now the PR has conflict with the master because we merged livy PR before this one. Please resolve it in smoke-tests.sh

```
<<<<<<< BIGTOP-3168
export KAFKA_HOME=${KAFKA_HOME:-/usr/lib/kafka}
=======
export LIVY_HOME=${LIVY_HOME:-/usr/lib/livy}
>>>>>>> master
```

guyuqi: Rebased.

evans-ye:
The test is good! +1
https://ci.bigtop.apache.org/view/Test/job/Build-Deploy-Smoke-Test-Pull-Request-All-Distros/

PR #548, opened 2019-10-18
BIGTOP-3255: Add Spark operator into distribution

youngwookim:


evans-ye:
Not a K8S expert but seems good to me!

youngwookim:
Thanks @evans-ye for the review.

PR #558, opened 2019-10-31
BIGTOP-3266: Rat check failed for one file with unapproved/unknown li…

guyuqi:
Add .dockerignore to Rat exclude-file list.


JunHe77:
Thanks @guyuqi for the fix.

PR #568, opened 2019-11-26
BIGTOP-3234: TEZ smoke tests implementation

guyuqi:
Add smoke tests for Bigtop Tez.

### DIFF
--- a/bigtop-tests/smoke-tests/tez/TestTezSmoke.groovy
+++ b/bigtop-tests/smoke-tests/tez/TestTezSmoke.groovy
@@ -1,0 +1,86 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import org.junit.BeforeClass
+import org.junit.AfterClass
+import org.apache.bigtop.itest.shell.Shell
+import static org.junit.Assert.assertNotNull
+import static org.junit.Assert.assertTrue
+import org.junit.Test
+import org.apache.bigtop.itest.JarContent
+import org.apache.bigtop.itest.TestUtils
+import org.junit.runner.RunWith
+
+class TestTezSmoke {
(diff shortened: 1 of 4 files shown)


bigtop-tests/smoke-tests/tez/TestTezSmoke.groovy
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.bigtop.itest.TestUtils
+import org.junit.runner.RunWith
+
+class TestTezSmoke {
+  static Shell sh = new Shell("/bin/bash -s");
+
+  static final String TEZ_HOME = System.getenv("TEZ_HOME");
+  static final String HADDOP_CMD = "hadoop jar "
+  static final String TEZ_EXMP_JAR = TEZ_HOME + "/tez-examples-0.9.1.jar "
+  static final String TEZ_TEST_JAR = TEZ_HOME + "/tez-tests-0.9.1.jar "
```
evans-ye: Is it possible to make the version *? Otherwise when tea upgraded, the test will just failed w/ no jar can be found.

bigtop-tests/smoke-tests/tez/TestTezSmoke.groovy
```diff
@@ -0,0 +1,86 @@
...
+  static final String SET_HADOOP_CLASSPATH = "export HADOOP_CLASSPATH=/etc/tez/conf:/usr/lib/tez/*:/usr/lib/tez/lib/*;"
```
evans-ye: Do we need `export` or we just need to set HADOOP_CLASSPATH for the command to run?

guyuqi: Set HADOOP_CLASSPATH here just for the command to run.

bigtop-tests/smoke-tests/tez/TestTezSmoke.groovy
guyuqi: Updated, thanks!

bigtop-tests/smoke-tests/tez/TestTezSmoke.groovy
evans-ye: Yes my thinking is we should fix the bigtop-deploy to set proper tez execution runtime instead of doing it in smoke-test. Do you agree?

guyuqi: Yes, it makes sense. 

Patch updated:
1. Set `HADOOP_CLASSPATH` in the stage of Bigtop-deploy.

2. Smoke-test process would fork several `shell `processes(`sh.exec`)  to execute test cases respectively. And smoke-test process wasn't able to get the 'environment variables' that was set in the stage of Bigtop-deploy.  So `HADOOP_CLASSPATH ` would be set here in the smoke-test like [ignite-hadoop tests](https://github.com/apache/bigtop/blob/master/bigtop-tests/smoke-tests/ignite-hadoop/TestIgniteHadoop.groovy#L36) does.

evans-ye: OK I see. Thanks!

evans-ye:
@guyuqi  Please rebase the PR so that we can proceed to the merge.

guyuqi:
Rebased, thanks @evans-ye.

evans-ye:
Looks great! +1.

PR #577, opened 2020-01-09
BIGTOP-3236: Zeppelin smoke tests implementation

guyuqi:
The Smoke tests deployment by provisioner would be broken due to [BIGTOP-2742](https://issues.apache.org/jira/browse/BIGTOP-2742), but actually zeppelin server has been started.

### DIFF
--- a/bigtop-tests/smoke-tests/zeppelin/TestZeppelinSmoke.groovy
+++ b/bigtop-tests/smoke-tests/zeppelin/TestZeppelinSmoke.groovy
@@ -34,14 +34,11 @@ class TestZeppelinSmoke {
 
   private static String ZEPPELIN_HOME = System.getenv("ZEPPELIN_HOME");
 
-  /* Align with the version defined in bigtop.bom*/
-  private static String ZEPPELIN_VERSION = "0.7.3";
-
   @Test
   public void InstallIntperTest() {
     sh.exec(ZEPPELIN_HOME
       + "/bin/install-interpreter.sh "
-      + "--name shell --artifact org.apache.zeppelin:zeppelin-shell:${ZEPPELIN_VERSION}"
+      + "--name cassandra,scio,md,python,flink,hbase,bigquery,jdbc"
     );
     logError(sh);
     assertTrue("Install Interpreter failed." + sh.getOut() + " " + sh.getErr(), sh.getRet() == 0);
--- a/bigtop-tests/smoke-tests/zeppelin/TestZeppelinSmoke.groovy
+++ b/bigtop-tests/smoke-tests/zeppelin/TestZeppelinSmoke.groovy
@@ -1,0 +1,50 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


evans-ye:
Looks good. Let me test this as well.

bigtop-tests/smoke-tests/zeppelin/TestZeppelinSmoke.groovy
```diff
@@ -0,0 +1,50 @@
...
+class TestZeppelinSmoke {
+  static Shell sh = new Shell("/bin/bash -s");
+
+  static final String ZEPPELIN_HOME = System.getenv("ZEPPELIN_HOME");
+
+  @Test
+  public void InstallIntperTest() {
+    sh.exec(ZEPPELIN_HOME
+      + "/bin/install-interpreter.sh "
+      + "--name shell --artifact org.apache.zeppelin:zeppelin-shell:0.7.3"
```
evans-ye: Same here if we can't make the version *. At least we can make it configurable?

provisioner/utils/setup-env-debian.sh
```diff
@@ -24,6 +24,8 @@ apt-get -y install rng-tools
 sed -i.bak 's@#HRNGDEVICE=/dev/null@HRNGDEVICE=/dev/urandom@' /etc/default/rng-tools
 service rng-tools start
 
+apt-get install -y locales
```
evans-ye: How about we add a comment for the reason why we need to install this?

guyuqi: The testing process would be broken due to "No such file or derictory: /etc/default/locale" in ubuntu16.04.

evans-ye: Got it. So let's just add a comment to reveal your finding.

guyuqi: Sure.

bigtop-tests/smoke-tests/zeppelin/TestZeppelinSmoke.groovy
guyuqi: Done.

evans-ye:
This one conflict with the codebase. Could you rebase @guyuqi ?

guyuqi:
> This one conflict with the codebase. Could you rebase @guyuqi ?

Rebased, thanks.  @evans-ye

bigtop-tests/smoke-tests/zeppelin/TestZeppelinSmoke.groovy
```diff
@@ -0,0 +1,53 @@
...
+  private static String ZEPPELIN_HOME = System.getenv("ZEPPELIN_HOME");
+
+  /* Align with the version defined in bigtop.bom*/
+  private static String ZEPPELIN_VERSION = "0.7.3";
+
+  @Test
+  public void InstallIntperTest() {
+    sh.exec(ZEPPELIN_HOME
+      + "/bin/install-interpreter.sh "
+      + "--name shell --artifact org.apache.zeppelin:zeppelin-shell:${ZEPPELIN_VERSION}"
```
evans-ye: So the wildcard does not work here?

guyuqi: Agreed and let me refine it.

evans-ye:
Depends on BIGTOP-3327 #631

bigtop-tests/smoke-tests/zeppelin/TestZeppelinSmoke.groovy
sekikn: Can we omit the `--artifact` option for the shell interpreter, as shown here? https://zeppelin.apache.org/docs/0.8.2/usage/interpreter/installation.html#install-specific-interpreters

guyuqi: Yep, it makes sense. Let me modify it later, thanks @sekikn

guyuqi:
The PR was updated: Refine the smoke test and fix issue of zeppelin init-broken

sekikn:
+1, tested with CentOS 8 and Debian 10 on x86_64 platform locally.
Thanks Yuqi!

PR #583, opened 2020-01-24
BIGTOP-3290. Fix build failure of Oozie.

iwasakims:


iwasakims:
Thanks for the comment, @oflebbe. I updated the patch to retain the license URL.

iwasakims:
Thanks, @oflebbe.

evans-ye:
+1.

PR #584, opened 2020-01-28
BIGTOP-3293. Add hadoop-kms package.

iwasakims:


iwasakims:
Only rpm fixes are available now. Fixes for deb is wip.

iwasakims:
I manually tested that hadoop-kms service starts and `hadoop key` works after `sudo yum install hadoop-kms`.

Keystore password must set set before starting hadoop-kms service. You may need to set JAVA_OPTS to make Tomcat bind ipv4 address depending on the OS settings.
```
$ sudo bash -c 'echo "export HADOOP_KEYSTORE_PASSWORD=mysecret" >> /etc/hadoop-kms/conf/kms-env.sh'
$ sudo bash -c 'echo "export JAVA_OPTS=\"\$JAVA_OPTS -Djava.net.preferIPv4Stack=true -Djava.net.preferIPv4Addresses=true \"" >> /etc/hadoop-kms/conf/kms-env.sh'
$ sudo service hadoop-kms start
```

Client-side settings must be set in /etc/hadoop/conf/core-site.xml.
```
  <property>
    <name>hadoop.security.key.provider.path</name>
    <value>kms://http@localhost:16000/kms</value>
  </property>
```

iwasakims:
Manual testing worked on Ubuntu 18.04 too. I will appreciate review comments. I would like to add puppet manifest and smoke-tests in follow-up jiras.

iwasakims:
merged as 4a4c6a2bbf79d1f04cabd327b846feac228f6a66.

PR #585, opened 2020-01-29
BIGTOP-3283. Bump HBase to 2.2.3.

iwasakims:


iwasakims:
`./docker-hadoop.sh -c 3 --smoke-tests` passed on my local with the following configuration.
```
docker:
        memory_limit: "8g"

distro: centos
components: [hdfs, yarn, mapreduce, hbase]
enable_local_repo: true
smoke_test_components: [hbase]
```

bigtop-packages/src/common/hbase/do-component-build
```diff
@@ -36,13 +36,20 @@ if [ $HOSTTYPE = "aarch64" ] ; then
...
-    -Dcheckstyle.skip=true                    \
-    ${MAVEN_ADDITIONAL} \
-    install site assembly:single "$@"
+
+MAVEN_PARAMS=" -DskipTests "
+MAVEN_PARAMS+=" -Dslf4j.version=1.7.25 "
+MAVEN_PARAMS+=" -Dhadoop-two.version=${HADOOP_VERSION} "
+MAVEN_PARAMS+=" -Dzookeeper.version=${ZOOKEEPER_VERSION} "
+MAVEN_PARAMS+=" -Dcheckstyle.skip=true "
+MAVEN_PARAMS+=" -Dmaven.repo.local=${HOME}/.m2/repository "
```
evans-ye: I wonder why introduce this line? It seems to be identical to default path.

evans-ye:
Let's see whether Phoenix can be built ;)
https://ci.bigtop.apache.org/view/Test/job/Build-Deploy-Smoke-Test-Pull-Request-All-Distros/26/

youngwookim:
https://phoenix.apache.org/download.html
> Apache HBase 2.0 is supported by 5.0.0
So if you would like to upgrade HBase 2, we need to bump up Phoenix version to 5.0.0-HBase-2.0 as well.

evans-ye:
I think we can merge this first and then tackle on Phoenix in another JIRA, how do you think?

iwasakims:
@youngwookim Phoenix is planned to be updated too in [BIGTOP-3284](https://issues.apache.org/jira/browse/BIGTOP-3284). I will work on it next.

bigtop-packages/src/common/hbase/do-component-build
iwasakims: I used the line for my try and error. I think we can remove it now. updating the PR.

bigtop-packages/src/common/hbase/do-component-build
```diff
@@ -36,13 +36,19 @@ if [ $HOSTTYPE = "aarch64" ] ; then
...
-    ${MAVEN_ADDITIONAL} \
-    install site assembly:single "$@"
+
+MAVEN_PARAMS=" -DskipTests "
+MAVEN_PARAMS+=" -Dslf4j.version=1.7.25 "
+MAVEN_PARAMS+=" -Dhadoop-two.version=${HADOOP_VERSION} "
+MAVEN_PARAMS+=" -Dzookeeper.version=${ZOOKEEPER_VERSION} "
+MAVEN_PARAMS+=" -Dcheckstyle.skip=true "
+
+# separate site and assembly:single to avoid assembly issues.
```
evans-ye: May I know what's the issue here that require the separation?

iwasakims:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:3.0.0:single (default-cli) on project hbase-assembly: Failed to create assembly: Error adding file 'org.apache.hbase:hbase-common:test-jar:tests:2.2.2' to archive: /home/centos/bigtop/build/hbase/rpm/BUILD/hbase-2.2.2/hbase-common/target/test-classes isn't a file. -> [Help 1]
```
The error was discussed in [ML](https://lists.apache.org/thread.html/0c39364a2949fab134107eaf164723e09d895b3eb2c16610b841e033%40%3Cuser.hbase.apache.org%3E) and [HBASE-21513](https://issues.apache.org/jira/browse/HBASE-21513). While the root cause seems to be unclear, `site` and `assembly:sinble` are separated as a workaround in the create-release script of HBase side. It worked for me too.

evans-ye: Got it. Thanks for the explanation. I suggest to add HBASE-21513 in the comment so other users can find the root cause themselves. We used to have the format like this:
```
# HBASE-21513: separate site and assembly:single to avoid assembly issues.
```

iwasakims: Thanks. I updated the comment as you suggested.

iwasakims:
[PHOENIX-4836](https://issues.apache.org/jira/browse/PHOENIX-4836) says that phoenix-5 does not support hadoop-2.. If we use hadoop-2.10.0, hbase-1.5.0 and phoenix-4.15.0-HBase-1.5 could be the candidate.

evans-ye:
Sounds like a plan. Let's move on to what you suggested and left this for the future release.

iwasakims:
I'm closing this one and going to submit new PR rebased on bigtop-1.5.

PR #596, opened 2020-02-14
BIGTOP-3304. Fix build failure of Flume against hive-2.3.6.

iwasakims:


bigtop-packages/src/common/flume/patch3-FLUME-3354.diff
```diff
@@ -0,0 +1,67 @@
...
+diff --git a/pom.xml b/pom.xml
+index bdd998d..2304421 100644
+--- a/pom.xml
++++ b/pom.xml
+@@ -79,7 +79,7 @@ limitations under the License.
+     <junit.version>4.10</junit.version>
+     <kafka.version>0.10.2.2</kafka.version>
+     <kite.version>1.0.0</kite.version>
+-    <hive.version>1.0.0</hive.version>
++    <hive.version>2.3.6</hive.version>
```
sekikn: I think this rewrite is unnecessary because Hive version is passed via environment variable. For consistency with other variables, would you remove the patch for pom.xml? Otherwise +1.
https://github.com/apache/bigtop/blob/3b1bf8064693059423608cfd276ef7ba81cf9e54/bigtop-packages/src/common/flume/do-component-build#L30-L35

evans-ye: Seems like the patch is identical to what has been provided in https://github.com/apache/flume/pull/315/files.
So I'm fine with the version since it's not the logic we specifically added.

sekikn: Oh, sorry that I missed the original PR. Thanks for the comment @evans-ye, will merge into master.

evans-ye: @sekikn thx for confirming as well. 
I guess both of us are +1 :)

iwasakims: Thank, @sekikn and @evans-ye. The patch3-FLUME-3354.diff could be updated on [BIGTOP-3305](https://issues.apache.org/jira/browse/BIGTOP-3305). Since pom.xml tends to be adjusted by bigtop patches, removing unnecessary change of pom.xml will be useful to avoid future conflict. Actually patch0-FLUME-2662.diff already touches pom.xml too.

PR #599, opened 2020-02-15
BIGTOP-3305. Bump Flume to 1.9.0.

iwasakims:


iwasakims:
This will conflict with the patch of [BIGTOP-3285](https://issues.apache.org/jira/browse/BIGTOP-3285) since this includes the change to the version of Kafka. I will rebase this once BIGTOP-3285 is done.

iwasakims:
I rebased this since BIGTOP-3285 was committed. [The unit tests of flume-kafka-source](https://github.com/apache/flume/tree/flume-1.9/flume-ng-sources/flume-kafka-source/src/test/java/org/apache/flume/source/kafka) are passed with the patch even kafka.version is 2.4.0 (instead of 2.3.1) on my local.

sekikn:
+1, successfully built on CentOS 7 & 8, Debian 9 & 10 and Ubuntu 16.04 & 18.04. It also passed the smoke test.

iwasakims:
Thanks, @sekikn!

PR #602, opened 2020-02-19
BIGTOP-3311: Fix Alluxio build failure with upgraded hadoop version

JunHe77:
Alluxio build failed after hadoop is bumpped to 2.10.0. The reason is
the version check regex in integration/yarn/pom.xml doesn't cover
version higher than 2.9.x.
Spring release repo is upgraded to https access.

Change-Id: I83e38af640881e8d61213d58a636af7d50393cdb


iwasakims:
+1(non-binding). The patch can be applied even after bumping Alluxio to 1.8.2.

evans-ye:
+1. 
A follow up is we should open up a JIRA in Alluxio to fix this in upstream and link it here. Once it's fixed upstream we can remove our own patch.

evans-ye:
image

PR #603, opened 2020-02-20
BIGTOP-3310: URL of maven and ant package list page is changed

JunHe77:
bigtop_toolchain uses packages list page hosted on
https://www.apache.org/dist/ to detect latest version of maven and ant.
This url has been moved to https://downloads.apache.org/ and latest_*_binary
functions need to be updated to reflect this.
To adapt to future URL change,'-L' option is added too.

Change-Id: Ia732be3c79f5d54b3db0eaa010adbfc8ff0d7395


JunHe77:
Verified with local bigtop-slaves image building on both x86 and arm64.

PR #605, opened 2020-02-21
BIGTOP-3312: Fix Ambari build failure as maven repo only accepts http…

JunHe77:
…s now

Maven repo has been upgrade to accept https access only.
The URLs defined in ambari pom.xml need to be changed to https.

Change-Id: I77ff02631800cd2ba7abfef071e903262039ee17

### DIFF
--- a/bigtop-packages/src/common/ambari/do-component-build
+++ b/bigtop-packages/src/common/ambari/do-component-build
@@ -18,6 +18,11 @@ set -ex
 . `dirname $0`/bigtop.bom
 
 export _JAVA_OPTIONS="-Xmx2048m -Djava.awt.headless=true"
+
+# these two repos haven updated to accpet https access only
+sed -i "s|http://download.java.net|https://download.java.net|g" pom.xml
+sed -i "s|http://repo.spring.io|https://repo.spring.io|g" pom.xml
+
 mvn clean package -DskipTests -Drat.skip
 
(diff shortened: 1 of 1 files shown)


iwasakims:
While the change worked for me, it is covered by [the patch](https://github.com/apache/ambari/pull/3045/files) of [AMBARI-25334](https://issues.apache.org/jira/browse/AMBARI-25334). It might be better to bump Ambari to 2.6.2 or later?

JunHe77:
That would be better way. Let me try build on local. If that's OK I'll update this PR

JunHe77:
Hi, @iwasakims  Comparing to 2.6.0, 2.6.2 has removed `ambari-client` and `ambari-shell` subdir. This means `install_ambari.sh` and deb/rpm package scripts should also get upgraded. For the purpose of fixing build failure on v1.5, using 2.6.2 or later would be too much.
So I'm thinking leave this patch for current 2.6.0 version for v1.5 release, and create another PR for bumping up to later version (say 2.7.x). What do you think?

iwasakims:
Thanks, @JunHe77. I'm +1(non-binding) on the current patch if bumping Ambari is not quick fix. I'll file a follow-up JIRA to bump Ambari without targeting Bigtop 1.5.0.

JunHe77:
Thanks, @iwasakims .

evans-ye:
Sounds like a plan. Thanks!

PR #606, opened 2020-02-21
BIGTOP-3315. Bump Alluxio to 1.8.2.

iwasakims:


iwasakims:
We need #602 to fix build failure.

evans-ye:
+1.

iwasakims:
Thanks, @evans-ye.

PR #607, opened 2020-02-26
Fix link to apache mirror for maven and ant

E7su:
'https://www.apache.org/dist/maven/maven-3/?F=0&V=1' ->
     'https://downloads.apache.org/maven/maven-3/?F=0&V=1'

     'https://www.apache.org/dist/ant/binaries/?F=0&V=1' ->
     'https://downloads.apache.org/ant/binaries/?F=0'


JunHe77:
Hi, @E7su , thanks for reporting this.
From your change it looks like the same issue reported at [BIGTOP-3310](https://issues.apache.org/jira/browse/BIGTOP-3310). Would you pls check if it is same as yours? If yes, then this PR (https://github.com/apache/bigtop/pull/603) should fix it.
Welcome to leave any comment.

PR #608, opened 2020-02-26
BIGTOP-3316: Fix HBase build failure on Arm platform

JunHe77:
After HBase is bumped to 1.5.0, jruby patch command in build
script doesn't apply due to regex not match.
This patch refines sed command to make it more versatile to
various asciidoctorj-pdf versions.

Change-Id: Ib414ba3e7b66cf08e2080c75e1c83c563fa88224


JunHe77:
Verified with build tests of Ubuntu and Centos on arm servers.

evans-ye:
Looks good. +1.

JunHe77:
Thanks for review, @evans-ye

PR #610, opened 2020-02-29
BIGTOP-3317: fix broken symlink for hadoop-hdfs.jar

elukey:
On Debian 9 the hadoop-hdfs.jar is broken, and causes failures like
oozie sharedlib creation on HDFS. This change replaces it with
hadoop-hdfs-client.jar, that seems to work.


elukey:
> Could you please add the same change to bigtop-packages/src/rpm/oozie/SPECS/oozie.spec since I really doubt this would be an debian-format only bug

Makes sense, just updated the pull request!

elukey:
@oflebbe hi! Whenever you have time let me know if this pull request needs more work or not :)

oflebbe:
merge done

evans-ye:
@oflebbe you can merge this on github. The code will automatically merge into ASF git.

oflebbe:
Haha @evans-ye , are you really sure? I had to apply this to two different branches at once.

evans-ye:
Oh. So for master it's OK but yes you're right. For 1.4 it's not working. I got your point ;)

PR #612, opened 2020-03-09
BIGTOP-3322. Lintian Warnings in Hadoop.

iwasakims:


iwasakims:
reproduced the issue and verified the fix on Ubuntu 18.04.

evans-ye:
LGTM. @oflebbe would love to take a 2nd look?

oflebbe:
Thanks!

PR #618, opened 2020-03-22
BIGTOP-3330: avoid duplicate jars between oozie and oozie-client

elukey:
Add a workaround to the install_oozie.sh script to avoid duplicate
jars between oozie and oozie-client. This issue leads to override
errors when installing the oozie package on Debian (oozie-client
is among its dependencies).

Credis to Olaf Flebbe for the solution.

Tested the packages on a Debian system and it seems working fine.


oflebbe:
Thanks, merged into master

oflebbe:
Merged to branch-1.4 as well

oflebbe:
And thanks!

elukey:
thanks!

PR #619, opened 2020-03-30
BIGTOP-3331: Provisioner Deployment failed on Fedora

guyuqi:
'yum.conf' doesn't exist in Fedora-29/31 environment.
It causes format error when echo "gpgcheck=0" >> /etc/yum.conf
So the provisioner deployment failed in Fedora-29/31.


JunHe77:
LGTM. +1

sekikn:
+1, confirmed that it works as expected on Fedora 31 and CentOS 7. Thanks!

evans-ye:
+1

PR #620, opened 2020-03-30
BIGTOP-3319: Add Logstash as Bigtop component

guyuqi:
The PR works on:
1. Implement build and package scripts for Logstash deb package.
2. Implement scripts and resources for Logstash rpm package.
3. Add 'ruby' to Bigtop_toolchain.
4. Implement deployment scripts for Logstash.
5. Implement smoke tests for Logstash.

The Bigtop-puppet and Bigtop-slaves docker images will be updated by the PR.


bigtop-packages/src/common/logstash/do-component-build
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+. `dirname $0`/bigtop.bom
+
+export RELEASE=1
+export LANG="en_US.UTF-8"
+export JAVA_HOME=`readlink -f /usr/bin/java | sed "s:jre/bin/java::"`
```
evans-ye, line 23: bigtop have a script to detect java home. Maybe we can leverage it.
```
# Autodetect JAVA_HOME if not defined
. /usr/lib/bigtop-utils/bigtop-detect-javahome
```

bigtop-packages/src/common/logstash/install_logstash.sh
```diff
@@ -0,0 +1,135 @@
+#!/bin/bash
```
evans-ye, line 1: Where does this patch come from and why this is necessary instead of pulling an official released version?

bigtop-packages/src/common/logstash/patch0-logstash.diff
```diff
@@ -0,0 +1,2393 @@
+diff -ruaN logstash-5.4.1/bin/bundle linaro-logstash-5.4.1/bin/bundle
```
evans-ye, line 1: Where does this patch come from and why this is necessary instead of pulling an official released version?

bigtop_toolchain/manifests/ruby.pp
```diff
@@ -0,0 +1,38 @@
...
+     creates => "/usr/src/$ruby24dir",
+  }
+
+  exec { "install ruby":
+     cwd => "/usr/src/$ruby24dir",
+     command => "/usr/src/$ruby24dir/configure --prefix=/usr/local --disable-shared && /usr/bin/make install",
+     creates => "/usr/local/bin/ruby",
+     timeout => 1800,
+     require => EXEC["download ruby"]
+  }
```
evans-ye, line 37: So ruby 2.4 version is the only one that works with log stash? Or we can install it from Distro's pkg manager (yum/apt)?

bigtop-packages/src/common/logstash/patch0-logstash.diff
guyuqi: Thanks for comment, evans.
1. Pick the logstash version: 5.4.1. It corresponds to the elasticsearch version in Bigtop.
2.  `patch0-logstash.diff` is to support Arm64.
The Jruby-1.9 in logstash-5.4.1 doesn't support Arm64 due to the fact that Jffi 1.2.12 which Jruby 1.9 depends on has no native AArch64 libraries. Also see:  [jffi](https://github.com/jnr/jffi/pull/32), [jruby1](https://github.com/jruby/jruby/pull/4548), [jruby2](https://github.com/jruby/jruby/pull/4567).
So `patch0-logstash.diff` upgraded jruby to 9K and also fix dependency issues for Arm64.
The modification for Arm64 support was archived in [linaro repo](https://git.linaro.org/leg/bigdata/logstash.git) and `patch0-logstash.diff` comes from it.

bigtop_toolchain/manifests/ruby.pp
guyuqi: Yea, for the dependency issues, the ruby-2.4 is the latest version that that works with logstash-5.4.1.
And the bigtop should support different kind of distros and platforms(x86/arm64). 
It may not be guaranteed the various distros could offer the ruby-2.4 pkg (x86/arm64) installation. IMHO, it's robust to build the ruby-2.4 binaries from source though it may cost extra time in bigtop_toolchain build.

bigtop-packages/src/common/logstash/do-component-build
guyuqi: Agree to leverage Bigtop script to detect JAVA_HOME, thanks for the info.
But I found that it's unnecessary to detect JAVA_HOME in `do-component-build` , and in the phase of `do-component-build` , `bigtop-utils` has not be ready yet.
Removed the JAVA_HOME detection.
Thanks for the comments, envans!

bigtop_toolchain/manifests/ruby.pp
evans-ye: Sounds good to me. I'm a +1 but not an expert on Logstash. If no one chimed in I'll commit this in this week. 
Ping me if not. Thanks!

bigtop-packages/src/common/logstash/do-component-build
evans-ye: Got it thanks for checking it :)

bigtop-packages/src/common/logstash/patch0-logstash.diff
evans-ye: Usually for apache project we just name it like patch0-HBASE-16864.diff, which we can easily find the source.
For this one... how about we name it something like patch0-logstash-for-arm64.diff so that it's more clear. Also describe the source of the repo in Bigtop JIRA helps people to locate the change.

guyuqi: Yes, it maskes sense.
Updated, thanks!

evans-ye: Sorry I forgot this one. @guyuqi is this ready to merge?

guyuqi: @evans-ye  Yes please, thanks a lot!

bigtop-packages/src/common/logstash/do-component-build
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+
+set -ex
+
+. `dirname $0`/bigtop.bom
+
+export RELEASE=1
+export LANG="en_US.UTF-8"
+
+rake bootstrap
```
evans-ye, line 24: https://ci.bigtop.apache.org/view/Test/job/Build-Deploy-Smoke-Test-Pull-Request-All-Distros/OS=centos-7-x86_64-deploy/53/consoleFull

```
+ rake bootstrap
/bigtop/build/logstash/rpm/SOURCES/do-component-build: line 24: rake: command not found
error: Bad exit status from /var/tmp/rpm-tmp.LphHXs (%build)
    Bad exit status from /var/tmp/rpm-tmp.LphHXs (%build)

* Where:
Script '/bigtop/packages.gradle' line: 500

* What went wrong:
Execution failed for task ':logstash-rpm'.
> Process 'command 'rpmbuild'' finished with non-zero exit value 1

* Try:
Run with --stacktrace option to get the stack trace. Run with --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 24s
```

evans-ye:
Seems failed on  X86 debian-9:

```
TestLogstashSmoke > LogstashFilePipelineTest FAILED
    java.lang.AssertionError: Logstash output did not match expected output. expected:<0> but was:<2>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:743)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at org.junit.Assert.assertEquals(Assert.java:555)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:93)
        at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:325)
        at org.codehaus.groovy.runtime.callsite.StaticMetaMethodSite$StaticMetaMethodSiteNoUnwrap.invoke(StaticMetaMethodSite.java:133)
        at org.codehaus.groovy.runtime.callsite.StaticMetaMethodSite.callStatic(StaticMetaMethodSite.java:102)
        at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallStatic(CallSiteArray.java:56)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callStatic(AbstractCallSite.java:194)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callStatic(AbstractCallSite.java:222)
        at TestLogstashSmoke.LogstashFilePipelineTest(TestLogstashSmoke.groovy:69)

Gradle Test Executor 2 finished executing tests.

> Task :bigtop-tests:smoke-tests:logstash:test FAILED
Finished generating test XML results (0.016 secs) into: /bigtop-home/bigtop-tests/smoke-tests/logstash/build/test-results/test
Finished generating test html results (0.028 secs) into: /bigtop-home/bigtop-tests/smoke-tests/logstash/build/reports/tests/test
:bigtop-tests:smoke-tests:logstash:test (Thread[Task worker for ':' Thread 2,5,main]) completed. Took 1 mins 50.441 secs.

2 tests completed, 1 failed
```

bigtop-packages/src/common/logstash/do-component-build
guyuqi: The rake would be installed when ruby was installed in bigtop toolchain.
`include bigtop_toolchain::ruby`

It seems we should rebuild the provisioner docker image to add ruby installation.

evans-ye: Yeah. I think you mean the bigtop/slaves images, right?

guyuqi:
Yes.
The build or smoke tests would fail if they were running against by original docker images (bigtop/slaves:trunk-ubuntu-16.04).

Run the logstash smoke tests on my local test enviroments:
1. rebuild docker images to add ruby/rake support (bigtop/slaves: yuqi-ubuntu-16.04 )
2.  config_ubuntu-16.04.yaml
```
docker:
        memory_limit: "4g"
        image:  "bigtop/puppet:yuqi-ubuntu-16.04"

repo: "http://repos.bigtop.apache.org/releases/1.4.0/ubuntu/16.04/$(ARCH)"
distro: debian
components: [logstash]

enable_local_repo: true
smoke_test_components: [logstash]
```
The log was posted [here](https://docs.google.com/document/d/1rZfBOet1GfnVr4FK06j7q9TgCH5xaEPCkJEAiQHT3jw/edit?usp=sharing).

Maybe we can retrigger the CI to rebuild docker images related.

guyuqi:
For the modification of this PR is complicated,
How about to: 
1. Submit another PR to just add ruby to bigtop toolchain.
2. Then we could verify this PR smoothly.
What do you think?  @evans-ye

evans-ye:
Sure. I think your suggestion is great. Let's work on the ruby in another JIRA and revisit this one.
Thanks!

guyuqi:
@evans-ye  BIGTOP-3346 created and #632 posted.

guyuqi:
As disscussion on #620 :

1. Remove the ruby from bigtog_toolchain.
2. For dependency problem, we still take rvm to install ruby-2.4 to build logstash in `do-component-build`.
3. Do not install ruby native package and run the logstash with the ruby provided by distros.
In other words, just support Logstash on the distros that provides Ruby 2.4+ natively:   ubuntu-18.04, debian-10, centos-8 and fedora-31.

guyuqi:
The logstash smoke tests in this PR depends on BIGTOP-3343 ( PR #634 ) for docker provisioner support in Ubuntu-18.04 and Debian-10.

evans-ye:
For 1. what do you meant by remove ruby? a) remove compile ruby b) remove ruby installation
For 2, seems like we can still move the build env preparation stuffs into bigtop_toolchain? no?

guyuqi:
If we still put ruby-installation into Bigtop_toolchain, it would override the ruby package already provided by distros. 
So both remove ruby compiling and ruby installation from Bigtop_toolchain.

For compile time: we must take ruby-2.4 to build logstash:
So just install ruby-2.4 in the phase of build `do-component-build` : 
```
curl -sSL https://get.rvm.io | bash -s stable
[ -s /usr/local/rvm/scripts/rvm ] && source /usr/local/rvm/scripts/rvm
rvm install ruby-2.4.0 && rvm use 2.4.0

```
For logstash runtime, we could leverage the ruby > 2.4 which has been already provided by distros.
Then we doesn't support logstash on CentOS-7, Debian-9 ,ubuntu-16.04 which provides ruby < 2.4 by [the previous disscussion](https://github.com/apache/bigtop/pull/632#issuecomment-624187208)

bigtop-tests/smoke-tests/logstash/TestLogstashSmoke.groovy
```diff
@@ -0,0 +1,71 @@
...
+    def FILE_OUTPUT = 'resources/output.log';
+
+    sh.exec(LOGSTASH_CMD
+      + "-f "
+      + FILEPIPELINE_CONFIG
+      + "&"
+    );
+
+    // Wait for completion
+    sleep(50000);
```
evans-ye, line 66: I'm a bit worry about this one. So for the test to be finished we need to wait for 50 seconds?

bigtop-packages/src/deb/logstash/logstash.preinst
```diff
@@ -0,0 +1,65 @@
...
+        if ! getent passwd "$LS_USER" >/dev/null; then
+                # Adding system user: logstash
+                adduser \
+                  --system \
+                  --ingroup "$LS_GROUP" \
+                  --home /usr/lib/logstash \
+                  --gecos "Logstash" \
+		  --disabled-password \
+		  --shell /bin/false \
+                  "$LS_USER" >/dev/null
```
evans-ye, line 47: Seems that the indent is not aligned?

bigtop-packages/src/common/logstash/do-component-build
```diff
@@ -0,0 +1,61 @@
...
+export LANG="en_US.UTF-8"
+
+if [ -f /etc/os-release ]; then
+    . /etc/os-release
+fi
+
+case ${ID} in
+    ubuntu | debian)
+        sudo apt-get -y update
+        sudo apt-get install -y gnupg2
```
evans-ye, line 31: I'm still not get it why we need to install things during do-components-build. All the env specific packages should be installed by bigtop_toolchain. Could you elaborate more?

bigtop-packages/src/common/logstash/do-component-build
```diff
@@ -0,0 +1,61 @@
...
+        ;;
+esac
+
+case ${ID}-${VERSION_ID} in
+    ubuntu-18* | debian-9)
+    gpg2 --list-keys
+    echo "disable-ipv6" >> ~/.gnupg/dirmngr.conf
```
evans-ye, line 38: Would you mind to share what's the purpose of this line?

bigtop-packages/src/common/logstash/do-component-build
```diff
@@ -0,0 +1,61 @@
...
+gpg2 --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
+curl -sSL https://get.rvm.io | bash -s stable
+[ -s /usr/local/rvm/scripts/rvm ] && source /usr/local/rvm/scripts/rvm
+
+case ${ID} in
+    ubuntu | debian)
+        rvm install ruby-2.4.0 && rvm use 2.4.0
+        ;;
+    fedora | centos)
+        rvm install ruby-2.3.8 && rvm use 2.3.8
```
evans-ye, line 51: ditto. The installation.

evans-ye:
>If we still put ruby-installation into Bigtop_toolchain, it would override the ruby package already provided by distros.
>So both remove ruby compiling and ruby installation from Bigtop_toolchain.
>
>So just install ruby in the phase of build do-component-build.

What's the different between *ruby-installation into Bigtop_toolchain* and *install ruby in the phase of build do-component-build*? Isn't both of them lead to the same result?

bigtop-tests/smoke-tests/logstash/TestLogstashSmoke.groovy
guyuqi: This `FILEPIPELINE ` operation would be performed in the background and output a result file.
It takes time ranging from 20~40 seconds on different platform hosts (x86/Arm64) from different vendors.  
Maybe there are other synchronization mechanisms for this operation but i'm sorry i don't know. 50 seconds may be a safe estimate to wait for the operation to complete which is performed asynchronously.

bigtop-packages/src/deb/logstash/logstash.preinst
guyuqi: I missed it and will fix it, thanks.

bigtop-packages/src/common/logstash/do-component-build
guyuqi: Before installing rvm, we should install GPG keys with `gpg2`. (guide from https://rvm.io/ )
In ubuntu/deian, `gpg2` is not  pre-installed.

bigtop-packages/src/common/logstash/do-component-build
guyuqi: There are the IPv6 issues happening in docker containers on ubuntu18 and debian-9:
```
+ gpg2 --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
gpg: keyserver receive failed: Cannot assign requested address
```
The soution was provided by rvm.io: https://rvm.io/rvm/security#ipv6-issues

bigtop-packages/src/common/logstash/do-component-build
guyuqi: Yes, I agree that all the env specific packages should be installed by bigtop_toolchain. 
But `rvm` and `gpg2` are just used for logstash-package build, and the bigtop_toolchain is the common facility for all components of Bigtop. 
So how about to just install them during `do-components-build` of logstash?

guyuqi:
>> What's the different between ruby-installation into Bigtop_toolchain and install ruby in the phase of build do-component-build? Isn't both of them lead to the same result?

1. Install ruby-2.4.0 in Bigtop_toolchain:  
Bigtop_toolchain is the common facility for all components in Bigtop. 
It would override the ruby package already installed by distros. 
If put ruby-2.4.0 in Bigtop_toolchain, ruby-2.4.0 is used for both build-time and runtime.
Maybe there are some dependency issues for other components which is concened from cos ([comments](https://issues.apache.org/jira/browse/BIGTOP-3346?focusedCommentId=17093457&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17093457)).

2.  Install ruby-2.4.0 in do-component-build: 
It's just for logstash build-time.
For logstash runtime, we could leverage the ruby (2.4+ ) provided by distros.

3. Why do we not leverage the ruby provided by distro but install ruby-2.4.0 in do-component-build to build logstash?
When `rake bootstrap` in the phase of build, some dependency packages like `minitar-0.5.4` would not be installed properly by leveraged ruby 2.4+ (2.4.10 or 2.5.8) on Arm64 (the problems of `gem `ecosystem on Arm64). 
For runtime, logstash would work properly on every version of ruby 2.4.0+.

guyuqi:
The PR was updated.

bigtop-tests/smoke-tests/logstash/TestLogstashSmoke.groovy
evans-ye: https://www.elastic.co/guide/en/logstash/current/plugins-outputs-file.html#plugins-outputs-file-flush_interval
 Try set it to 0 so that it writes more frequently. I'm not an expert though...

bigtop-packages/src/common/logstash/do-component-build
evans-ye: I think that not our convention. bigtop_toolchain is exactly how weinstall all the required dependencies, across all the components for bigtop bistro. Install things in do-components-build is rather not aligned. toolchain is also served the purpose of formalize the build env and make it static so that we can iterate the building code on top of it w/o being effected by env changes. 

Any other reason you'd like to put in in do-components-build?

evans-ye: OK I see the comprehensive comparison in below. Thanks!

bigtop-tests/smoke-tests/logstash/TestLogstashSmoke.groovy
guyuqi: That's informative, thanks!
The `flush_interval => 0` makes the waiting interval reduce to 5 seconds on x86.
I will update this PR after verifying it on Arm64.

evans-ye:
>> If put ruby-2.4.0 in Bigtop_toolchain, ruby-2.4.0 is used for both build-time and runtime.
This is not correct. If put ruby-2.4.0 in bigtop_toolchain, It's used in build time only. Why you say so?

>> Install ruby-2.4.0 in do-component-build:
>> It's just for logstash build-time.
Even though we do this. We should remove the ruby installation after logstash is built. Otherwise the build env is tainted. Can we guarantee a clean removal?

I don't think dynamic dependency installation is a good way to go. We should make build env consistent. The way we do it is bigtop toolchain.

guyuqi:
>>_This is not correct. If put ruby-2.4.0 in bigtop_toolchain, It's used in build time only. Why you say so?_

Maybe I have misunderstanding here and please correct me:
In docker build scripts of bigtop-slaves:  https://github.com/apache/bigtop/blob/master/docker/bigtop-slaves/build.sh#L45-L79
The ruby-2.4.0 as the part of `bigtop_toolchain` would be contained in the docker images like` bigtop/slaves :xxx-ubuntu-18.04.
`
If we run the logstash against in this docker image,  ruby-2.4.0 would override the ruby-2.5 provided by Ubuntu-18.04.


>>_I don't think dynamic dependency installation is a good way to go. We should make build env consistent. The way we do it is bigtop toolchain._

For the reason above and the disscussion in #632, ruby would not be installed in bigtop_toolchain. 
I agree with you that dynamic dependency installation in `do-component-build` is not an elegant solution, but it may be a feasible approach just for this case.  Surely I would like to remove the ruby installation after logstash is built by [rvm remove](https://rvm.io/rubies/removing).

evans-ye:
>The ruby-2.4.0 as the part of bigtop_toolchain would be contained in the docker images like bigtop/slaves :xxx-ubuntu-18.04.
>If we run the logstash against in this docker image, ruby-2.4.0 would override the ruby-2.5 provided by Ubuntu-18.04.

bigtop/slaves is used for package building. bigtop/puppet is used for deployment. bigtop/slaves is based on bigtop/puppet so that it's able to execute puppet(for running bigtop toolchain).

> For the reason above and the discussion in #632, ruby would not be installed in bigtop_toolchain.
I'd like to clarify the understanding of #632 with you. CMIIW, what I get from #632 is we don't need to compile ruby 2.4 on our own if we just focus on the newer Distros. They've ruby 2.4 provided already. Therefore I think we can just install Distro. provided ruby by bigtop toolchain. What's the reason that we can't do Package {"ruby"}  directly in bigtop toolchain?

>Surely I would like to remove the ruby installation after logstash is built by rvm remove.
Thanks. If we have to do it in do-component-build anyway. I'm OK with it if removal is implemented.

guyuqi:
>>What's the reason that we can't do Package {"ruby"} directly in bigtop toolchain?

Because some dependency packages like minitar-0.5.4 would **not** be installed properly by leveraged ruby 2.4+ (2.4.10 or 2.5.8), it means ruby-2.4.0 is only one version that was verified to **build** logstash on all platform and Distros. 
It may be not guaranteed that the version of ruby installed by `Package {"ruby"}` is **2.4.0** , for ruby-2.4.0 package may be not available for all Distros in different platforms(arm/x86/ppc).
For runtime, logstash would work properly on every version of ruby 2.4.0+.

>>Thanks. If we have to do it in do-component-build anyway. I'm OK with it if removal is implemented.

I'll update it, thanks @evans-ye

evans-ye:
> minitar-0.5.4 would not be installed properly by leveraged ruby 2.4+ (2.4.10 or 2.5.8)

OK, so I'm very much confused by this. No other minor version is acceptable? What kind of problem it would be to prevent minitar-0.5.4 to install with ruby 2.4+? I'm very curious since I feel there's something unprofessional happening there...

Anyhow, the argument is for building Logstash, exactly version of ruby = 2.4.0 is required. If that's the constrain then I've no choice to agree with the solution you proposed.

guyuqi:
>>Can we use different version of minicar? 

Really thanks for you efforts and patience to help review this PR.
I once tried to use different versions of minitar. 
But in logstash, minitar dependency is defined as:
```
  task "archive-tar-minitar" do
    Rake::Task["gem:require"].invoke("minitar", "0.5.4")
  end # task archive-minitar
```
minitar-0.5.4 was deprecated and not be found in ruby-2.4+ (RubyGems packages manager).
If we install ruby by `Package {"ruby"}`, different versions of rubys would be installed in different Distros repectively. Then different ruby defines different versions of mintar. We need to adjust minitar version in logstash config corresponding to ruby. And there are many other packages that needs to be adjusted. 
IMHO, it's better to use the packages which logstash originally required and it's also efficient to use a ruby-2.4.0 to build logstash which was verifyied in almost all Distros and platforms.  Hope you will understand the reason of it, thanks again.

There are still some smoke tests issues in Arm64 about `flush_interval => 0`. Let me figure it out ,and then update this PR.

guyuqi:
@evans-ye Agree with your point that the basic packages should be installed in Bigtop_toolchain.
I 'd like to put `gnupg2` into Bigtop_toolchain not in `do-component-build`.

evans-ye:
Yes. gnupg2 should be in toolchain.
Back on minitar, so if 5.4.1's dependency is so hard to get it right or is stalled, why not use the newest version 7.7.0 or something else that is easier to fit into the our build environment?

guyuqi:
minitar is not installed by ourselves in our build environment. It's embedded in `logstash-core/logstash-core.gemspec`:
`gem.add_runtime_dependency "minitar", "~> 0.5.4"`.  
The minitar is installed accompanied by logstash build process triggered by RubyGems.

evans-ye:
Yup. I get that minitar is fixed on the Logstash side. I was thinking that:
1. Can we patch the Logstash on build time to set the dependency to minitar greater than 0.5.4?
2. Can we skip Logstash 5.4.1 and jump to 7.X so that we don't have to deal with the staled package issue.

We can still agree on the patch as is and iterate from there. But I was just trying to explore the options.

guyuqi:
The PR was updated:
1. Put gpg2 installation to bigtop_toolchain
2. Smoke tests: Reduce waitting time to 15s by setting 'flush_interval => 0'
3. Remove Ruby after build
4. RPM scripts refined

sekikn:
I applied the latest PR and ran `./gradlew toolchain logstash-pkg` on CentOS 7, but came across the following error, which is similar to the one that Evans reported before.

```
Installing RVM to /home/vagrant/.rvm/
    Adding rvm PATH line to /home/vagrant/.profile /home/vagrant/.mkshrc /home/vagrant/.bashrc /home/vagrant/.zshrc.
    Adding rvm loading line to /home/vagrant/.profile /home/vagrant/.bash_profile /home/vagrant/.zlogin.
Installation of RVM in /home/vagrant/.rvm/ is almost complete:

  * To start using RVM you need to run `source /home/vagrant/.rvm/scripts/rvm`
    in all your open shell windows, in rare cases you need to reopen all shell windows.
<warn>Thanks for installing RVM 🙏</warn>
Please consider donating to our open collective to help us maintain RVM.

👉  Donate: <code>https://opencollective.com/rvm/donate</code>


+ '[' -s /usr/local/rvm/scripts/rvm ']'
+ rvm install ruby-2.4.0
/home/vagrant/bigtop/build/logstash/rpm/SOURCES/do-component-build: line 41: rvm: command not found
+ rake bootstrap
/home/vagrant/bigtop/build/logstash/rpm/SOURCES/do-component-build: line 43: rake: command not found
```

The do-component-build script for logstash seems to install RVM into `$HOME/.rvm/scripts/rvm` rather than `/usr/local/rvm/scripts/rvm` in my environment.

guyuqi:
@sekikn 
Yep, the build and deployment were always triggered under `root`  in my scenario.
But the default directory of `root `is `/usr/local/rvm/scripts/rvm` , 
otherwise the directory of other `users `is `$HOME/.rvm ` as  shown in `https://get.rvm.io`:
```
    if
      (( UID == 0 ))
    then
      rvm_user_install_flag=0
      rvm_prefix="/usr/local"
      rvm_path="${rvm_prefix}/rvm"
    else
      rvm_user_install_flag=1
      rvm_prefix="$HOME"
      rvm_path="${rvm_prefix}/.rvm"
    fi
```
Thanks for your comments.
PR is updated.

bigtop-tests/smoke-tests/logstash/TestLogstashSmoke.groovy
```diff
@@ -0,0 +1,71 @@
...
+import static org.junit.Assert.assertFalse
+import org.junit.Test
+import org.apache.bigtop.itest.TestUtils
+import org.junit.runner.RunWith
+import static org.junit.Assert.assertEquals
+
+class TestLogstashSmoke {
+  static Shell sh = new Shell("/bin/bash -s");
+
+  static final String LOGSTASH_CMD = "/usr/lib/logstash/bin/logstash "
```
sekikn, line 32: Should we use `$LOGSTASH_HOME` instead of a hard-coded path?

bigtop-tests/smoke-tests/logstash/TestLogstashSmoke.groovy
```diff
@@ -0,0 +1,71 @@
...
+      + LOGSTASH_CMD
+      + "-f "
+      + TINYBENCH_CONFIG
+    );
+    def out = sh.out.join('\n');
+    assertFalse("Logstash tiny benchmark failed", (out =~ /ERROR/).find());
+  }
+
+  @Test
+  public void LogstashFilePipelineTest() {
```
sekikn, line 55: This test assumes /bigtop-home is the directory into which Bigtop is copied. It's always true for the docker provisioner, but I think it depends on the user when deploying a real cluster.
In addition, this test failed with the following error in my environment, for some reason (it looks like output file was not generated, but I didn't investigate in detail):

```
TestLogstashSmoke > LogstashFilePipelineTest FAILED
    java.lang.AssertionError: Logstash output did not match expected output. expected:<0> but was:<2>
```

How about removing LogstashFilePipelineTest for now (and adding it in the future if needed), because LogstashTinyBenchTest seems to be enough as a minimal smoke test if it also checks the status code of the logstash command?

guyuqi: Yes, it makes sense. I'll update it soon, thanks

sekikn: Thanks for the response Yuqi, one additional comment. I tried the latest PR on Ubuntu 16.04, and got the following error.

```
$ ./gradlew toolchain
$ ./gradlew clean logstash-pkg

(snip)

+ rake bootstrap
mkdir -p vendor
mkdir vendor/_
mkdir -p build
mkdir -p build/bootstrap
Installing minitar 0.5.4 because the build process needs it.
[bootstrap] Fetching and installing gem: minitar (0.5.4)
Successfully installed minitar-0.5.4
1 gem installed
Installing bundler ~> 1.9.4 because the build process needs it.
[bootstrap] Fetching and installing gem: bundler (~> 1.9.4)
Successfully installed bundler-1.9.10
1 gem installed
Invoking bundler install...
Error Bundler::InstallError, retrying 1/10
An error occurred while installing childprocess (4.0.0), and Bundler cannot continue.
Make sure that `gem install childprocess -v '4.0.0'` succeeds before bundling.

(snip)

Too many retries, aborting, caused by Bundler::InstallError
rake aborted!
Bundler::InstallError: An error occurred while installing childprocess (4.0.0), and Bundler cannot continue.
Make sure that `gem install childprocess -v '4.0.0'` succeeds before bundling.
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/installer.rb:145:in `install_gem_from_spec'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/installer.rb:269:in `block in install_sequentially'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/installer.rb:268:in `install_sequentially'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/installer.rb:90:in `run'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/installer.rb:18:in `install'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/cli/install.rb:80:in `run'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/cli.rb:155:in `install'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/vendor/thor/lib/thor.rb:359:in `dispatch'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/vendor/thor/lib/thor/base.rb:440:in `start'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/cli.rb:10:in `start'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:160:in `execute_bundler'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:131:in `block in execute_bundler_with_retry'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:129:in `block in execute_bundler_with_retry'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:203:in `capture_stdout'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:128:in `execute_bundler_with_retry'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:118:in `invoke!'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/rakelib/vendor.rake:122:in `block in (root)'
Gem::RuntimeRequirementNotMetError: childprocess requires Ruby version >= 2.4.0. The current ruby version is 2.3.0.
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/source/rubygems.rb:116:in `block in install'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/rubygems_integration.rb:131:in `preserve_paths'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/source/rubygems.rb:113:in `install'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/installer.rb:107:in `install_gem_from_spec'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/installer.rb:269:in `block in install_sequentially'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/installer.rb:268:in `install_sequentially'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/installer.rb:90:in `run'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/installer.rb:18:in `install'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/cli/install.rb:80:in `run'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/cli.rb:155:in `install'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/vendor/thor/lib/thor.rb:359:in `dispatch'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/vendor/thor/lib/thor/base.rb:440:in `start'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/vendor/bundle/jruby/2.3.0/gems/bundler-1.9.10/lib/bundler/cli.rb:10:in `start'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:160:in `execute_bundler'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:131:in `block in execute_bundler_with_retry'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:129:in `block in execute_bundler_with_retry'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:203:in `capture_stdout'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:128:in `execute_bundler_with_retry'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/lib/bootstrap/bundler.rb:118:in `invoke!'
/home/vagrant/bigtop/output/logstash/logstash-5.4.1/rakelib/vendor.rake:122:in `block in (root)'
Tasks: TOP => bootstrap => vendor:all => vendor:gems
(See full trace by running task with --trace)
debian/rules:31: recipe for target 'override_dh_auto_build' failed
make[1]: *** [override_dh_auto_build] Error 1
make[1]: Leaving directory '/home/vagrant/bigtop/output/logstash/logstash-5.4.1'
debian/rules:28: recipe for target 'build' failed
make: *** [build] Error 2
dpkg-buildpackage: error: debian/rules build gave error exit status 2
debuild: fatal error at line 1376:
dpkg-buildpackage -rfakeroot -D -us -uc -b
 failed

> Task :logstash-deb FAILED
```

The message says the ruby version is 2.3.0, but 2.4.0 seems to be enabled via RVM.

```
$ ruby -v
ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-linux]
$ rvm list
=* ruby-2.4.0 [ x86_64 ]

# => - current
# =* - current && default
#  * - default
```

Am I missing something?

guyuqi: The PR was updated for removing LogstashFilePipelineTest and using `$LOGSTASH_HOME `instead of a hard-coded path.

For ubuntu-16.04 build issue, let me look into it on my environment and will give you feddback later, thanks.

guyuqi: Sorry for the late response.
The root cause is: 
Childprocess-4.0.0 is newly released in **June 18, 2020**. 
But it is not compatible with 5.4.1.
Specify childprocess to 3.0.0.
Thanks!

sekikn:
+1, I confirmed that packaging, deploying and running the smoke test succeeded with the latest PR on CentOS 7 & 8, Debian 9 & 10 and Ubuntu 16.04 & 18.04, on x86_64 platform. Great work!

Just one comment, I noticed running a smoke test as a non-root user fails due to the lack of write permission to /usr/lib/logstash/data. I think most of (or all of, possibly?) other smoke tests are runnable as a non-root user, so it'd be nice if it's fixed as another issue.

evans-ye:
@sekikn Thanks  for carrying on with this. @guyuqi Thanks for keep up with this and finally get it merged.

PR #631, opened 2020-04-21
BIGTOP-3327: Fix Zeppelin build issue after upgrading Spark-2.4.5

guyuqi:
The PR is to fix Zeppelin build issue after upgrading Spark-2.4.5:

1. Bump Zeppelin to 0.8.2 which is the latest and stable release.

2. Fix the issues:
- Spark-2.4.5 dependency issues
- Arm64 native binaries missing( phantomjs / protoc / protoc-gen-grpc-java)
- Maven code coverage  issues


sekikn:
Great work @guyuqi! Will try your PR.

My only concern is, this build probably fails on ppc64le due to the problem @evans-ye [mentioned before](https://issues.apache.org/jira/browse/BIGTOP-3188?focusedCommentId=16793377&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16793377). protoc binary was published on the maven central by your effort and grpc-java as well, but they may be reluctant to do the same thing for ppc64le, that is relatively minor platform.

@evans-ye @oflebbe @JunHe77 Is it possible for us to publish the next release without Zeppelin only on ppc64le? Should we vote on dev-ml?

guyuqi:
@sekikn,  agree with you.
Google published protoc ppc64le binaries since [3.7.0](https://repo1.maven.org/maven2/com/google/protobuf/protoc/3.7.0/), 
but for protoc-gen-grpc-java, ppc64le binaries are still not available in the latest release [1.28.1](https://repo1.maven.org/maven2/io/grpc/protoc-gen-grpc-java/1.28.1/).

evans-ye:
I'm a +1 to @sekikn.
The rationale here should be if no one volunteer to work on it, we can vote to drop the support and move onward. As discussed previous on mailing list: our strategy is focus on core, user-demanded features instead of supporting many components across many platforms.
So yes. Let's bring this back to mailing list. We should give people who care about PPC the chance to chime in.

evans-ye:
@guyuqi could you help to arrange a discussion in bigtop dev list? Thanks!

guyuqi:
> @guyuqi could you help to arrange a discussion in bigtop dev list? Thanks!

Yes, of course. @evans-ye

guyuqi:
As disscussion on dev list, we decide to drop the Zeppelin ppc64le support in the next release.
Do you have any other comments on this PR? @evans-ye @sekikn

evans-ye:
Thanks for brought this up. Could you please conclude the voting result on the mailing list so that we spread out an explicitly conclusion to all the community? Once that's done we can proceed this one as discussed. You can search over the bigtop mailing list for [RESULT] tag and get to know how to do the tally. Thanks!

sekikn:
+1, thank you for closing the vote @guyuqi!

evans-ye:
Thanks @guyuqi @sekikn .

PR #632, opened 2020-04-27
BIGTOP-3346: Introduce Ruby to Bigtop Toolchain

guyuqi:
As discussion on [PR #620](https://github.com/apache/bigtop/pull/620) for BIGTOP-3319,
this PR is to introduce Ruby to Bigtop Toolchain for Logstash support.

### DIFF
--- a/bigtop_toolchain/manifests/ruby.pp
+++ b/bigtop_toolchain/manifests/ruby.pp
@@ -13,20 +13,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-class bigtop_toolchain::installer {
-  include bigtop_toolchain::jdk
-  include bigtop_toolchain::ruby
-  include bigtop_toolchain::maven
-  include bigtop_toolchain::ant
-  include bigtop_toolchain::gradle
-  include bigtop_toolchain::protobuf
-  include bigtop_toolchain::packages
-  include bigtop_toolchain::env
-  include bigtop_toolchain::user
-  include bigtop_toolchain::renv
+class bigtop_toolchain::ruby {
 
-  stage { 'last':
-    require => Stage['main'],
+  require bigtop_toolchain::packages
+
+  $url = "https://cache.ruby-lang.org/pub/ruby/2.4"
+
+  $ruby24 = "ruby-2.4.0.tar.gz"
+  $ruby24dir = "ruby-2.4.0"
+
+  exec { "download ruby":
+     cwd  => "/usr/src",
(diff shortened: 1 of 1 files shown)


guyuqi:
> I prefer to have ruby from distro or rvm.

@oflebbe really thanks for your comments.

We do need the ruby-related facilities such as ruby rumtime, rake, gem and bundle to build logstash. 
And for the Arm64 dependency issues, the ruby-2.4 is the latest version that works with logstash. 
But ruby version is 2.0 in Centos 7 and 2.3 in ubuntu16.04.

When leveraging rvm to install ruby, rvm would pull the ruby binaries and spend less time than  compiling ruby directly on x86_64.
But on Arm64, no ruby binaries are available and rvm would also download source code and compile ruby.

IMHO, our Bigtop should support different kind of distros and platforms(x86_64/arm64).
If ruby-2.4 were downloaded and compiled directly , it could guarantee that ruby-related facilities would be properly installed on the various distros and platforms though it may cost extra time just on x86 when building bigtop_toolchain. WDYT? 
Thanks!

oflebbe:
Regarding centos7 we should use https://www.softwarecollections.org/en/scls/rhscl/rh-ruby26/ . This is a redhat provided service for RHEL and Centos.
Ubuntu 16.04 should be removed, since it is outdated. We should support 18.04 instead.

evans-ye:
@guyuqi @oflebbe I think we can leverage Ruby on those platforms which already have ruby that meet the requirement. I'm guessing that most of the user just choose one arch X platform. If we compile our own dependencies, then user much do so in there env, which is not that ideal, does this make sense?

sekikn:
Just my two cents, how about we only support Logstash on the distros that provides Ruby 2.4+ natively and doesn't on relatively old ones, i.e., CentOS 7, Debian 9 (which provides Ruby 2.3.3) and Ubuntu 16.04?

oflebbe:
I agree @sekikn . All the platforms you are mentioning should be deprecated and therefore it should be ok to not provide new functionality.

guyuqi:
> I agree @sekikn . All the platforms you are mentioning should be deprecated and therefore it should be ok to not provide new functionality.

Yup, it makes sense. Thanks, @oflebbe @sekikn @evans-ye

guyuqi:
From the disscussion above, this PR would be abandoned. 
For logstash, the ruby native package would not be installed in bigtop_toolchain.
Just support Logstash on the distros that provides Ruby 2.4+ for runtime. (distros: Ubuntu-18.04, Debian-10, Centos-8 and Fedora-31)

PR #635, opened 2020-04-29
BIGTOP-3351: Fix sqoop build error on Ubuntu 18.04

guyuqi:
rsync package missing


JunHe77:
Verified with ubuntu-18.04 on both arm and x86. Thanks, @guyuqi

PR #633, opened 2020-04-27
BIGTOP-3347. Fix test failures of Hive ProxyFileSystem against Hadoop 2.10.0.

iwasakims:

PR #636, opened 2020-04-29
BIGTOP-3348. Fix Hive test failures caused by NoSuchMethodError of ApplicationId.fromString.

iwasakims:


sekikn:
I've already merged #633, so would you rename the new patch in this PR to a name starting with "patch4" to clarify their application order (though it doesn't matter which one is applied first in this case)?

iwasakims:
@sekikn Sure. I renamed the added file.

iwasakims:
`./docker-hadoop.sh -c 3 --smoke-tests` passed with the following configuration.
```
docker:
        memory_limit: "12g"
distro: centos
components: [hdfs, yarn, mapreduce, tez, hive]
enable_local_repo: true
smoke_test_components: [hive]
```

evans-ye:
+1. Thanks!

iwasakims:
Thanks, @evans-ye.

PR #637, opened 2020-04-30
BIGTOP-3352: Fix qfs build error on Ubuntu 18.04

guyuqi:
dh-python package missing.


JunHe77:
Verified on arm and x86 with bigtop/slaves:ubuntu-18.04. Thanks @guyuqi

evans-ye:
Thanks @guyuqi

PR #642, opened 2020-05-25
BIGTOP-3359. Fix failure of installing Hadoop due to conflicting .build-id entry on CentOS 8.

iwasakims:


iwasakims:
`./gradlew hadoop-rpm` worked on both CentOS 8 and CentOS 7. While CentOS 7 does not support _build_id_links, defining unused macros should be harmless.

sekikn:
+1, I also confirmed that Hadoop was successfully built and installed on CentOS 7 and 8 with this PR. Thanks!

PR #643, opened 2020-05-29
BIGTOP-3361: No libcrypt.so.1 in Fedora-31

guyuqi:
libcrypt.so.1 is not available in Fedora-31 for Arm64 (both in bigtop/puppet and bigtop/slaves) and x86( just in bigtop/puppet).
It would cause some dependency issues when some pacakges need to link it such as jffi.
Add libxcrypt-compat in Bigtop_toolchain to fix it.

### DIFF
--- a/bigtop_toolchain/bin/puppetize.sh
+++ b/bigtop_toolchain/bin/puppetize.sh
@@ -24,7 +24,7 @@ case ${ID}-${VERSION_ID} in
     fedora-31)
         dnf -y install yum-utils
         dnf -y check-update
-        dnf -y install hostname findutils curl sudo unzip wget puppet procps-ng
+        dnf -y install hostname findutils curl sudo unzip wget puppet procps-ng libxcrypt-compat
         # On Fedora 31, the puppetlabs-stdlib package provided by the distro installs the module
         # into /usr/share/puppet/modules, but it's not recognized as the default module path.
         # So we install that module in the same way as CentOS 7.


evans-ye:
+1 w/lo test. Is it compatible in x86?

oflebbe:
Second thought: I am not sure if this is the right way to fix. IIRC  JNI will load "libcrypt.so" rather a "libcrypt.so.1" at runtime.  @guyuqi did you check / put a reference to the source which actually tries to load libcrypt.so.1 rather libcrypt.so . 
IMO It might be better to install libxcrypt-devel to supply libcrypt.so.

If jffi really refers to libcrypt.so.1 than Debian would be broken, since it only supplies libcrypt.so to the user by libc-dev .

guyuqi:
>> since it initially confused me why this change should affect Arm64 only

Thanks for comments, @oflebbe 
For neither `libcrypt.so.1` nor `libcrypt.so` exist in `bigtop/puppet:trunk-fedora-31` in all platforms, 
agree with you to change the name of the JIRA issue. 

In Fedora-31, both of `libcrypt.so.1` and `libcrypt.so` do exist in x86 images: `bigtop/slaves:trunk-fedora-31`. 
However `libcrypt.so.1` doesn't exist in Arm images: `bigtop/slaves:trunk-fedora-31-aarch64`.
It misled me at first. I'll change it later. 

>>did you check / put a reference to the source /errors which actually tries to load libcrypt.so.1 rather libcrypt.so .

The errors in Fedora-31 like: 
```
Failed to load native POSIX impl; falling back on Java impl. Stacktrace follows.
java.lang.UnsatisfiedLinkError: libcrypt.so.1: cannot open shared object file: No such file or directory
        at jnr.ffi.provider.jffi.NativeLibrary.loadNativeLibraries(NativeLibrary.java:87)
        at jnr.ffi.provider.jffi.NativeLibrary.getNativeLibraries(NativeLibrary.java:70)
        at jnr.ffi.provider.jffi.NativeLibrary.getSymbolAddress(NativeLibrary.java:49)
        at jnr.ffi.provider.jffi.NativeLibrary.findSymbolAddress(NativeLibrary.java:59)
        at jnr.ffi.provider.jffi.AsmLibraryLoader.generateInterfaceImpl(AsmLibraryLoader.java:158)
        at jnr.ffi.provider.jffi.AsmLibraryLoader.loadLibrary(AsmLibraryLoader.java:89)
        at jnr.ffi.provider.jffi.NativeLibraryLoader.loadLibrary(NativeLibraryLoader.java:44)
        at jnr.ffi.LibraryLoader.load(LibraryLoader.java:325)
        at jnr.ffi.Library.loadLibrary(Library.java:127)
        at jnr.posix.POSIXFactory$DefaultLibCProvider$SingletonHolder.<clinit>(POSIXFactory.java:279)

```
In Debian-10 for x86 and Arm64 in our docker images (`bigtop/slaves:trunk-debian-10`),
Both libcrypt.so.1 and libcrypt.so are available:
```
/lib/aarch64-linux-gnu/libcrypt.so.1
/lib/x86_64-linux-gnu/libcrypt.so.1
```

```
/usr/lib/aarch64-linux-gnu/libcrypt.so
/usr/lib/x86_64-linux-gnu/libcrypt.so

```
But in `bigtop/puppet:trunk-debian-10` (x86 and Arm64): 
`libcrypt.so.1` exists but `libcrypt.so` doesn't.
Just for this case, becuase `libcrypt.so.1` is always available in `bigtop/puppet:trunk-debian-10` and `bigtop/slaves`,  no errors occurred in Debian-10 in all platforms.

I'm not sure there are any issues for missing `libcrypt.so` in Debian. How about to install both `libxcrypt-devel` and libxcrypt-compat in `bigtop/puppet` ? Or deal with it case by case?

oflebbe:
i am off the internet for some days from now. feel free to apply the proposed fix, i like to look into it when i am back on my computer. btw: what was the command triggering the messages you are referring to?

guyuqi:
> i am off the internet for some days from now. feel free to apply the proposed fix, i like to look into it when i am back on my computer. btw: what was the command triggering the messages you are referring to?

Build logstash: 
```
export JRUBY_OPTS="-Xnative.verbose=true"
rake bootstrap --trace
```

oflebbe:
Ok found out what's the root cause. logstash is using jruby. jruby depends on jnr-posix . jnr-posix for some reasons is tied to load libcrypt.so.1 
https://github.com/jnr/jnr-posix/blob/acbaf0fe6dd3667a125a77bc690a64ec250a49f5/src/main/java/jnr/posix/POSIXFactory.java#L315
However Fedora considers some of the interfaces of the libcrypt insecure since it is tied to the DES algorithm, which is considered insecure today.  
https://fedoraproject.org/wiki/Changes/FullyRemoveDeprecatedAndUnsafeFunctionsFromLibcrypt#Detailed_Description
However libcrypt's crypt() seems the only function used 
https://github.com/jnr/jnr-posix/blob/9f86fd649cf01ff22c7b8ccf17c1bf9590c6be23/src/main/java/jnr/posix/BaseNativePOSIX.java#L96
This function seems no to be deprecated.
To summarize: the workaround libxcrypt-compat is perfectly ok. And the source explains why Debian works as well.
So +1 from me

JunHe77:
Thanks @oflebbe for the detail investigation, and @guyuqi  for this fix.

PR #645, opened 2020-06-09
BIGTOP-3369: RHEL8 does not have PowerTools repo but CodeReady Builder

masatana:
Update puppetize.sh to use CodeReady repo when the environment is RHEL8.

I tested this change w/ RHEL8 on AWS EC2.

Note that if you want to run puppetize.sh, you may have to add /opt/puppetlabs/bin in secure_path as noted in [README.md](https://github.com/apache/bigtop/blob/master/README.md#for-developers-building-the-entire-distribution-from-scratch) and re-login.

(Though this is out-of-scope of this PR, should we automate this process?)

### DIFF
--- a/bigtop_toolchain/bin/puppetize.sh
+++ b/bigtop_toolchain/bin/puppetize.sh
@@ -47,7 +47,7 @@ case ${ID}-${VERSION_ID} in
         yum -y install hostname curl sudo unzip wget puppet
         puppet module install puppetlabs-stdlib --version 4.12.0
         ;;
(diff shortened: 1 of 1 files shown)


sekikn:
Though RHEL 8 is not our officially supported distro, this fix may be useful for some users.
I tested the PR and confirmed it worked (with adding /opt/puppetlabs/bin to secure_path in /etc/sudoers).
+1, thanks @masatana!

masatana:
Thank you @sekikn !

PR #646, opened 2020-06-15
BIGTOP-3364: Leverage pip3/python3 to install setuptools

guyuqi:
setuptools could not be installed by pip/pip2 on Ubuntu-16.04/18.04 and centos-7.
From https://packaging.python.org/tutorials/installing-packages/#requirements-for-installing-packages,  it suggests to leverage python3/pip3 to install setuptools package.

### DIFF
--- a/bigtop_toolchain/manifests/packages.pp
+++ b/bigtop_toolchain/manifests/packages.pp
@@ -237,30 +237,23 @@ class bigtop_toolchain::packages {
     }
   }
 
-  # Install Python packages using pip
-  case $operatingsystem{
-    /(?i:(centos|fedora|redhat))/: {
-      $pip = 'python2-pip'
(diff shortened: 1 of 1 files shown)


sekikn:
+1, I confirmed that your PR works at least for CentOS 7 and 8, and Ubuntu 18.04 on an x86_64 machine, using `./gradlew toolchain zeppelin-pkg -Dbuildwithdeps=true`. I'm also running the same command now on Debian 9 and 10, and Ubuntu 16.04 and it seems to work.
So go ahead and merge if you've already confirmed that it also works on aarch64 :)

JunHe77:
Merged. Thanks for the patch, @guyuqi .

evans-ye:
LGTM. +1

PR #648, opened 2020-07-03
BIGTOP-3372. Building Spark fails on Debian 9 and Ubuntu 16.04.

sekikn:


sekikn:
This PR removes "--as-cran" option from the command that checks the SparkR tarball to avoid an error which is fixed in the newer version of R.
I also added a new smoke test so that we can confirm SparkR works without that option.
Tested on CentOS 7, Debian 9 and Ubuntu 16.04.

guyuqi:
The PR works for ubuntu 16.04 and Debian 9 on Arm64.
+1

sekikn:
Thanks for confirming! Just merged.

PR #652, opened 2020-07-20
BIGTOP-3377. Fix deployment failures on Fedora 31

sekikn:


sekikn:
I tested all components which spec file was changed in this PR are successfully built and installed on Fedora 31.
For some components, I also confirmed that this PR works on CentOS 7, just in case.

sekikn:
In accordance with the CTR model, I've just merged this PR.

PR #653, opened 2020-07-22
BIGTOP-3378. Fix Livy deployment failure on Debian and Ubuntu.

sekikn:


sekikn:
I confirmed this fix worked with Debian 9/10 and Ubuntu 16.04/18.04.
(I installed Livy and its dependency (e.g., Spark) via puppet in one shot, and sometimes I had to execute the same `puppet apply` command a few times when timeout occurred, but that command succeeded at last)

sekikn:
Merged, thanks!

PR #656, opened 2020-07-28
BIGTOP-3371. Unable to build Apache Ambari on RHEL8.

sekikn:
@guyuqi @mattAndruff @masatana Would you review if possible? In this PR, I intended:

* Specify python package name explicitly as 'python2' in spec files so that they work with CentOS 8 and Fedora 31
* Replace Ambari functions that depend on the rpm-python package, which is not provided on CentOS 8, with an implementation using the `rpm` command
* Define the `python` command as an alias of `python2` on CentOS 8 and Fedora 31, which don't provide unversioned python command by default

I confirmed that Ambari, Livy and Zeppelin (the latter of two use Python3, in contrast to Ambari) are successfully built on CentOS 7, 8 and Fedora 31 with this PR.

### DIFF
--- a/bigtop-packages/src/common/ambari/patch4-AMBARI-CentOS8.diff
+++ b/bigtop-packages/src/common/ambari/patch4-AMBARI-CentOS8.diff
@@ -12,10 +12,10 @@ index 07b0b68319..9cfc2daeb5 100644
 +rpm.dependency.list=openssl,\nRequires: zlib,\nRequires: python2 >= 2.6
 +deb.dependency.list=openssl, zlibc, python (>= 2.6)
 diff --git a/ambari-common/src/main/python/resource_management/core/providers/package/__init__.py b/ambari-common/src/main/python/resource_management/core/providers/package/__init__.py
-index 799cf3eed0..ade58292cb 100644
+index 799cf3eed0..bc066e6318 100644
 --- a/ambari-common/src/main/python/resource_management/core/providers/package/__init__.py
 +++ b/ambari-common/src/main/python/resource_management/core/providers/package/__init__.py
-@@ -359,9 +359,8 @@ class RPMBasedPackageProvider(PackageProvider):
+@@ -359,17 +359,13 @@ class RPMBasedPackageProvider(PackageProvider):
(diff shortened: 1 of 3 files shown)


bigtop-packages/src/common/ambari/patch4-AMBARI-CentOS8.diff
```diff
@@ -0,0 +1,59 @@
...
++++ b/ambari-common/src/main/python/resource_management/core/providers/package/__init__.py
+@@ -359,9 +359,8 @@ class RPMBasedPackageProvider(PackageProvider):
+   INSTALLED_PACKAGE_VERSION_COMMAND = "rpm -q --queryformat '%{{version}}-%{{release}}' \"{0}\""
+ 
+   def rpm_check_package_available(self, name):
+-    import rpm # this is faster then calling 'rpm'-binary externally.
+-    ts = rpm.TransactionSet()
+-    packages = ts.dbMatch()
++    import os
++    packages = os.popen("rpm -qa --queryformat '%{name} '").read().split()
```
masatana, line 26: `packages` variable should contain a dict-like object as this is used like [this](https://github.com/apache/ambari/blob/release-2.6.1/ambari-common/src/main/python/resource_management/core/providers/package/__init__.py#L370)

masatana:
@sekikn Thanks for working on this! I haven't tested this patch yet, but I left one comment.
I'll test this patch on my environment later.

bigtop-packages/src/common/ambari/patch4-AMBARI-CentOS8.diff
sekikn: That's right, thanks! I fixed the succeeding for-loop.

masatana: +1. Thanks!

masatana:
LGTM (non-binding). I also built (`./gradlew toolchain` && `./gradlew ambari-pkg`) and installed on CentOS8 with success.
If we encounter any problems with running Ambari on CentOS8, let's fix it with another addendum patch.

sekikn:
Thank you for the review @masatana @iwasakims, just merged.

PR #658, opened 2020-08-16
BIGTOP-3387: Building QFS fails on CentOS 8

sekikn:


sekikn:
I confirmed that building qfs succeeded on CentOS 7 (as a regression test), 8 and Fedora 31 with this PR.

guyuqi:
The build command: 
`docker run --rm -v `pwd`:/ws --workdir /ws bigtop/slaves:trunk-centos-8 bash -c '. /etc/profile.d/bigtop.sh; qfs-pkg yum'`

It seems to encounter a dependencies issue of  `net-tools `.

```
> Task :qfs-rpm FAILED
Installing /ws/output/qfs/qfs-2.0.0-1.el8.src.rpm
error: Failed build dependencies:
        net-tools is needed by qfs-2.0.0-1.el8.x86_64

```

sekikn:
Merged into master. Thanks for the review @guyuqi!

PR #662, opened 2020-08-25
BIGTOP-3390. Building Alluxio fails on Fedora 31.

sekikn:
I confirmed that building alluxio succeeded on Fedora 31 and CentOS 7/8 (as regression test) with this PR.


guyuqi:
Verified it on Fedora-31 for Arm64.
+1

sekikn:
Thanks! Just merged.

PR #663, opened 2020-08-31
BIGTOP-3392. All packaging jobs are failing on Fedora 31 for aarch64 and ppc64le.

sekikn:
With this PR, we can pass some options from Jenkins to the `docker run` command for avoiding the problem reported as BIGTOP-3392.
I confirmed that the specified options were passed to the `docker run` command, as follows:

```
$ ./gradlew -Pprefix=trunk -POS=debian-9 -Pdocker-run-option="-v $HOME/.m2:/root/.m2" bigtop-utils-pkg-ind

(snip)

++ docker run -d -v /home/sekikn/.m2:/root/.m2 bigtop/slaves:trunk-debian-9 /sbin/init

(snip)

BUILD SUCCESSFUL in 6m 15s
1 actionable task: 1 executed
```

```
$ ./gradlew -Pprefix=trunk -POS=debian-9 -Pdocker-run-option="-v $HOME/.m2:/root/.m2" repo-ind

(snip)

++ docker run -d -v /home/sekikn/.m2:/root/.m2 bigtop/slaves:trunk-debian-9 /sbin/init

(snip)

BUILD SUCCESSFUL in 5m 14s
1 actionable task: 1 executed
```

### DIFF
--- a/bigtop-ci/build.sh
+++ b/bigtop-ci/build.sh
@@ -17,7 +17,7 @@
 BIGTOP_HOME=`cd $(dirname $0)/.. && pwd`
 
 usage() {
-    echo "usage build.sh --prefix trunk|1.3.0|1.2.1|... --os debian-8|centos-7|... --target hadoop|tez|... [--nexus]"
+    echo "usage build.sh --prefix trunk|1.4.0|1.3.0|... --os debian-9|centos-7|... --target hadoop|tez|... [--nexus] [--docker-run-option ...]"
     exit 1 # unknown option
 }
 
@@ -46,6 +46,10 @@ case $key in
     CONFIGURE_NEXUS="configure-nexus"
     shift
     ;;
+    --docker-run-option)
+    DOCKER_RUN_OPTION="$2"
+    shift
+    ;;
     *)
     usage
     ;;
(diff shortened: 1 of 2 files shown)


sekikn:
I'm going to merge this so as to try it on Jenkins. I'll revert immediately if it doesn't work as expected.

sekikn:
Confirmed this PR worked on both aarch64 and ppc64le on Jenkins.

evans-ye:
CMIIW, this is to leverage m2 caches to mitigate the network timeout problem which is due to huge amount of maven download IO?

sekikn:
Yes, I think we can leverage this feature to mounting the .m2 directory for avoiding the network timeout. It will also speed up the CI.
But I have a few concerns:

* When multiple containers share the same .m2 directory of their host, what happens if they try to write into the same file simultaneously?
* If a file in .m2 is broken by a network issue or something, do succeeding builds continue to fail unless the corrupted file is removed? (Maybe we need a new Jenkins job that finds and removes corrupted jars in .m2?)

I'll investigate the above concerns and try to enable the .m2 mount on Jenkins when I have some time :)

PR #664, opened 2020-09-03
BIGTOP-3391: Support GPDB on Perl 5.30+ environment

JunHe77:
GPDB is using a special variable $[ in its Perl script,
but it was deprecated from v5.30. So it can't be built
on some distros, e.g., Fedora 31 and Ubuntu 20.04.

This patch fixes indexs used in array and substr functions.

Change-Id: I92c793243c37ce57ffe146b71a5f732488f5e4a9


JunHe77:
This also fixes greenplum-db/gpdb#8989

sekikn:
Great work @JunHe77, I searched "[" and "substr" in parse.pl, and this PR seems to contain all required fixes.
Will try to build gpdb with it on some distros.

JunHe77:
Thanks for helping verifying this, @sekikn 
I just did quick test by issuing:
`"/usr/bin/perl" ./parse.pl . < ../../../../src/backend/parser/gram.y > preproc.y`
in <gpdb_root>/src/interfaces/ecpg/preproc/, with and without this fix to verify the generated prepro.y. So far it looks good as the outputs match.

sekikn:
+1, Confirmed building GPDB succeeded on CentOS 7, Debian 9 and Fedora 31. Thanks @JunHe77!

PR #666, opened 2020-09-04
BIGTOP-3395. Building Kafka is failing on Fedora 31 for ARM64.

sekikn:
This PR upgrades Groovy and Gradle to v2.5.4 and v5.6.4 respectively so that building Kafka on fedora-31-aarch64 succeeds.
I confirmed `./gradlew kafka-pkg` after `./gradlew toolchain` succeeded inside the fedora-31-aarch64 container with this PR.


sekikn:
Merging this to check if the problem is resolved on Jenkins.

PR #667, opened 2020-09-05
BIGTOP-3396. Building Ambari is failing on Fedora 31 for ARM64.

sekikn:


sekikn:
Merging this to check if the problem is resolved on Jenkins.

PR #668, opened 2020-09-07
BIGTOP-3398. Fix GPDB puppet manifest to work.

sekikn:
Tested on all target distros.

### DIFF
--- a/bigtop-deploy/puppet/modules/gpdb/manifests/init.pp
+++ b/bigtop-deploy/puppet/modules/gpdb/manifests/init.pp
@@ -95,21 +95,28 @@ class gpdb {
 
     class install_packages{
       case $operatingsystem{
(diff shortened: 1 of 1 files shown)


guyuqi:
+1

iwasakims:
@sekikn I got error on CentOS 8. The cause seems to be hard coded OS version in the URL of epel. Can we fix this here?
https://github.com/apache/bigtop/blob/master/bigtop-deploy/puppet/modules/gpdb/manifests/init.pp#L100

sekikn:
@guyuqi Thank you for reviewing!
@iwasakims Thank you for catching this, I'll update the PR.

sekikn:
Updated the PR, tested on CentOS 7 and 8.

iwasakims (approved):
+1. Thanks, @sekikn.

PR #671, opened 2020-09-14
BIGTOP-3401: Failed to build Centos-8 docker images for Arm64

guyuqi:
1. Enable Centos-8 docker image on Arm64.
2. Modify Puppet modulepath for puppetlabs-4.12 to fix issue :`'Unknown function: 'any2array'`
3. Fix docker provisioner on Arm64 Centos-8.
4. Fix `shellcheck `warning in `bash `scripts.

### DIFF
--- a/bigtop_toolchain/bin/puppetize.sh
+++ b/bigtop_toolchain/bin/puppetize.sh
@@ -56,10 +56,11 @@ case ${ID}-${VERSION_ID} in
         puppet module install puppetlabs-stdlib --version 4.12.0
         ;;
     centos-8*)
-        rpm -Uvh https://yum.puppet.com/puppet5-release-el-8.noarch.rpm
+        rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
         dnf -y check-update
-        dnf -y install hostname curl sudo unzip wget puppet-agent 'dnf-command(config-manager)'
-        puppet module install puppetlabs-stdlib
+        dnf -y install hostname curl sudo unzip wget puppet 'dnf-command(config-manager)'
+        # Install the module in the same way as Fedora 31 and CentOS 7 for compatibility issues.
+        puppet module install puppetlabs-stdlib --version 4.12.0
         # Enabling the PowerTools and EPEL repositories via Puppet doesn't seem to work in some cases.
         # As a workaround for that, enable the former here in advance of running the Puppet manifests.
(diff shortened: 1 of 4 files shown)


guyuqi:
Confirmed that Centos-8 docker images(`bigtop/puppet` and `bigtop/slaves`) could be successfully built both on x86 and Arm64.
It's also ok for provisioner depolyment on x86 and Arm64.

sekikn:
I tried the PR with CentOS 8 on x86_64 and came across the following error:

```
$ sudo bigtop_toolchain/bin/puppetize.sh 

(snip)

/etc/puppetlabs/code/modules
└── puppetlabs-stdlib (v4.12.0)
$ ./gradlew toolchain

(snip)

Error: Evaluation Error: Unknown function: 'concat'. (file: /home/vagrant/bigtop/bigtop_toolchain/manifests/packages.pp, line: 70, column: 17) on node centos8.localdomain

> Task :toolchain FAILED
```

I've not tried it on arm64 yet, but it seems to be required to add `/etc/puppetlabs/code/modules` to the modulepath in build.gradle (or replace `/etc/puppetlabs/code/environments/production/modules` with it) at least for x86_64.

guyuqi:
Please forgive that maybe I misunderstand the way how to run the `puppetize.sh`.
I run the `puppetize.sh` by invoking [build.sh](https://github.com/apache/bigtop/blob/master/docker/bigtop-puppet/build.sh).
It shows the `ENV PATH`sould be set in centos-8:
```
ENV_PATH=""
if [ ${OS} = "centos" -a ${VERSION} -ge 8 ]; then
  ENV_PATH='ENV PATH /opt/puppetlabs/bin:$PATH'
fi
```

guyuqi:
Let me try to add the modulepath in build.gradle.

guyuqi:
Update the puppetlibs-4.12's modulepath into build.gradle.
Confirmed that `./gradlew toolchain` was successfully built on Ubuntu18.04, Debian10, Centos8 and Fedora31 for both x86 and Arm64.

sekikn:
+1, I confirmed the following on an x86_64 machine:

* `./build.sh trunk-centos-8` (and centos-7 for regression purpose) in docker/bigtop-slaves succeeded.
* `sudo bigtop_toolchain/bin/puppetize.sh; ./gradlew toolchain` on CentOS 8 (and a few other distros) succeeded. In addition, packaging and deploying hadoop with the combination of puppet 6.17.0 and the existing manifest also succeeded.

PR #674, opened 2020-09-18
BIGTOP-3408. hadoop-docker.sh generates broken site.yaml from config.yaml containing multiple components.

iwasakims:
[Link to BIGTOP-3408](https://issues.apache.org/jira/browse/BIGTOP-3408)

### DIFF
--- a/provisioner/docker/docker-hadoop.sh
+++ b/provisioner/docker/docker-hadoop.sh
@@ -93,7 +93,7 @@ create() {
         repo=$(get-yaml-config repo)
     fi
     if [ -z ${components+x} ]; then
-        components="[`echo "$(get-yaml-config components)" | sed 's/ /, /g'`]"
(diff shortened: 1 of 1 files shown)


sekikn:
Nice catch, thanks!
+1 merged into master.

evans-ye:
@iwasakims Thanks for the fix. I'm just curious why this was not a problem before. Do we changed something and trigger this? Things like OS versions?

sekikn:
@evans-ye, this bug was recently introduced by #671 (I missed that...).

iwasakims:
@evans-ye  The issue was in the code reading components from config.yaml. Specifying components from cmdline like `./hadoop-docker.sh -c 3 --stack 'hdfs,yarn,mapreduce'` was not affected.

evans-ye:
OK, I see. So now we change it back. Thanks!

PR #675, opened 2020-09-22
BIGTOP-3394: Building Logstash is failing on CentOS 8, Debian 10 and Ubuntu 18.04

guyuqi:
when rvm install ruby-2.4.0 under user: jenkins,
it fails to "try_sudo" to install autolibs dependencies.
Disable unnecessary autolibs installation.

### DIFF
--- a/bigtop-packages/src/common/logstash/do-component-build
+++ b/bigtop-packages/src/common/logstash/do-component-build
@@ -43,6 +43,12 @@ then
 else
   [ -s $HOME/.rvm/scripts/rvm ] && source $HOME/.rvm/scripts/rvm
 fi
+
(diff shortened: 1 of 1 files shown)


sekikn:
+1, tested on CentOS 7/8 and Ubuntu 16.04/18.04. Thank you for the perseverant investigation, @guyuqi!
(To be precise, I had to replace the URL of keyserver with hkp://ipv4.pool.sks-keyservers.net in bigtop-packages/src/common/logstash/do-component-build only on CentOS 8 for some reason. Without that, the gpg2 command failed with a `gpg: keyserver receive failed: End of file` error. But it was a problem specific to my environment, I think)

PR #677, opened 2020-09-22
BIGTOP-3410. Fix bigtop-tomcat build failure due to downloading problem.

sekikn:


iwasakims:
I tried `./gradlew bigtop-tomcat-rpm`.  The added patch did not applied since bigtop-tomcat.spec does not contain  the `#BIGTOP_PATCH_FILES` line.

sekikn:
I totally forgot that, thanks @iwasakims! Just updated the PR.

iwasakims (approved):
+1.

iwasakims:
I would be nice if we can fallback to archive.apache.org for getting not-newest dependencies in follow-up JIRA.

sekikn:
Filed as https://issues.apache.org/jira/browse/BIGTOP-3412 :wink:

PR #679, opened 2020-09-24
BIGTOP-3414. Fix the default value of GPDB configuration for smoke test.

sekikn:


iwasakims:
I got smoke-tests failure on smoke-tests even with the patch applied. The cause looks irrevant to the fix. 
```
$ ./docker-hadoop.sh --smoke-tests
+ HCFS_USER=hdfs
+ SMOKE_TESTS=gpdb
+ '[' -z gpdb ']'
+ '[' -e /usr/lib/bigtop-utils/bigtop-detect-javahome ']'
+ echo -e '\nUNABLE TO DETECT JAVAHOME SINCE bigtop-utils NEEDS TO BE INSTALLED!\n'

UNABLE TO DETECT JAVAHOME SINCE bigtop-utils NEEDS TO BE INSTALLED!

+ exit 2
```

I installed only gpdb by following config. @sekikn Could you share your config? I think another components pulls the implicit dependencies.
```
docker:
        memory_limit: "8g"

distro: centos
components: [gpdb]
enable_local_repo: true
smoke_test_components: [gpdb]
```

sekikn:
> I think another components pulls the implicit dependencies.

That's right, gpdb requires bigtop-utils for detecting JAVA_HOME. It should be filed as an issue.

iwasakims (approved):
+1. Adding bigtop-utils to components of config.yaml for the smoke-tests worked for me. Thanks, @sekikn.

iwasakims:

PR #682, opened 2020-09-25
BIGTOP-3413. QFS doesn't work with Hadoop's TeraGen.

sekikn:
Tested with both mapreduce and qfs on Ubuntu 18.04.


iwasakims (approved):
+1. This worked for my CentOS 7 env.

PR #686, opened 2020-10-28
BIGTOP-3437. Building Logstash fails due to amazing_print version mismatch.

sekikn:


sekikn:
Tested with all supported distros on x86_64.

guyuqi:
+1,  verify it on Arm64 centos-7.

PR #689, opened 2020-11-04
BIGTOP-3439. Oozie's smoke test fails on CentOS and Fedora.

iwasakims:
jars are removed by install_oozie.sh on building RPM due to the fix of [BIGTOP-3330](https://issues.apache.org/jira/browse/BIGTOP-3330). The cause is that bigtop-packages/src/common/oozie/install_oozie.sh is written in deb centric way. I prefer quick-fix rather than overhaul here.


sekikn:
+1, I built oozie with this PR on all supported distros and confirmed the generated packages contained the dependencies in /usr/lib/oozie/lib. Thanks @iwasakims!

PR #696, opened 2020-11-11
BIGTOP-3429. QFS smoke test fails on Ubuntu 16.04.

sekikn:
This PR ensures that symlinks to qfs shared libraries are created after target directory creation. Tested with Ubuntu 16.04 and 18.04 on x86_64.


guyuqi:
+1, confirmed on Arm64 ubuntu16.04.

iwasakims (approved):
+1

PR #697, opened 2020-11-12
BIGTOP-3443: Building Spark fails on Ubuntu 16.04

guyuqi:
R version in Ubuntu16.04 is 3.2.3.
The packages 'digest' and 'htmltools' are not available (for R version 3.2.3).
There are some other dependency issues in R-3.2 which seems to be deprecated.
Upgrade the R version to 3.4 for ubuntu-16.04.

PR #698, opened 2020-11-13
BIGTOP-3446. Fix toolchain task failure due to undefined variable in renv.pp.

iwasakims:
This if follow-up of  [BIGTOP-3443](https://issues.apache.org/jira/browse/BIGTOP-3443) / #697 .


guyuqi:
+1, verified it on Ubuntu 16/18 and Fedora-31.

iwasakims:
Thanks, @guyuqi .

PR #699, opened 2020-11-13
BIGTOP-3447: Fix Unknown variable: 'rpkgs' in docker bigtop/slaves for Fedora/Centos/SLES

guyuqi:
'$rpkgs' should be defined outside of  'Distros detection'.


guyuqi:
Duplicated to #698.

PR #704, opened 2020-11-25
BIGTOP-3280. Bump Hadoop to 3.2.1.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3280


iwasakims:
This PR addresses packaging issues only. Contents of bundled configs might be updated based on the result of smoke-tests.

I think we should merge this after 1.5.0 is published for the ease of release process.

iwasakims:
We need protobuf built without `--disable-shared` for libhdfspp (C++ client library for HDFS). You need to reinstall protobuf before building hadoop with `-Pnative`.

```
sudo mv /usr/local/bin/protoc /tmp/
sudo mv /usr/src/protobuf-2.5.0 /tmp/
./gradelw toolchain
```

oflebbe:
Hi @iwasakims , having a shared library in an system incompatible version installed system visible is asking for trouble, AFAIK. 

Can we install the shared protobuf library in the libhdfspp RPATH as well ? 
* For instance packaging the protobuf dependency with libhfspp
 or 
* install shared protobuf in an private directory and extending the RPATH of libhfspp to include protobuf shared library ?

guyuqi (commented):
Hi @iwasakims , 
Could pls not remove `--disable-shared` in protobuff for libhdfspp?

The root cause of build issue for Hadoop-3.2.x libhdfspp is :
Protobuf was not build by "-fPIC", 
but the native lib libhdfspp was built by  "-fPIC".
So the issue occurred in the phase of  're-compile' and 'link'.

We can add `"CFLAGS=-fPIC"` in the build command of protocbuf:
`command => "/usr/src/$protobuf8dir/configure --prefix=/usr/local --disable-shared \"CFLAGS=-fPIC\" \"CXXFLAGS=-fPIC\" && /usr/bin/make install && /sbin/ldconfig",
`
I verify it in Hadoop-3.3.0 build, thanks.

iwasakims:
Thanks @oflebbe and @guyuqi. I update the PR based on @guyuqi 's suggestion.

iwasakims:
There is ongoing release process of Hadoop 3.2.2. We can remove patch0-HADOOP-16739-branch-3.2.diff if we can use hadoop-3.2.2. We need patch0-revert-HADOOP-16598.diff again instead..

guyuqi:
+1

iwasakims:
Thanks, @guyuqi. I'm merging this since bigtop-1.5.0 was released. Let me post a patch to upgrade to hadoop-3.2.2 in follow-up JIRA. 3.2.2 will be released in a week.

guyuqi:
@iwasakims 
As disscussed in [here](https://issues.apache.org/jira/browse/BIGTOP-3280?focusedCommentId=17238449&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17238449), we plan to upgrade hadoop to 3.3.3.
May I ask why we upgrade hadoop from 3.2.1 to 3.2.2 not directly to 3.3.3?

iwasakims:
> May I ask why we upgrade hadoop from 3.2.1 to 3.2.2 not directly to 3.3.3?

I'm assuming some downstream projects depending on hadoop are not ready for hadoop-3.3.0 which contains big changes in transitive dependencies. I think we can release the next bigtop (on hadoop-3.x) early by using 3.2.x. In addition, hadoop-3.2.2 should contain more bug fixes than 3.3.0 (released 9 months ago).

PR #707, opened 2020-12-02
BIGTOP-3459. Deploying GPDB fails on ARM64 CentOS 7.

sekikn:


sekikn:
Tested with CentOS 7, 8 and Fedora 31 on both aarch64 and x86_64.

bigtop-deploy/puppet/modules/gpdb/manifests/init.pp
```diff
@@ -137,6 +140,16 @@
             ensure   => latest,
           }
           package { ["paramiko"]:
+            ensure   => latest,
+            provider => pip,
+            require  => Package["setuptools"],
+          }
+          package { ["setuptools"]:
```
elukey, line 147: Totally ignorant about the environment but are those packages named python-setuptools or simply setuptools?

elukey:
LGTM 👍

sekikn:
Thanks! Merged into master and branch-1.5.

PR #708, opened 2020-12-03
BIGTOP-3460. Extend timeout for Ambari Server setup.

sekikn:


iwasakims (approved):
+1. Thanks, @sekikn.

elukey:
@sekikn feel free to discard this, but would a comment explaining why 600 has been chosen good in this case? Otherwise it looks good :)

sekikn:
@elukey Thanks! I added a comment about that.

sekikn:
Merged into master and branch-1.5. Thank you for reviewing @iwasakims @elukey!

PR #709, opened 2020-12-04
BIGTOP-3462. Deploying Livy fails if the SPARK_HOME directory doesn't exist.

sekikn:


iwasakims:
@sekikn I think the PR makes implicit dependency on spark::common which defines Package['spark-core']. Addressing [BIGTOP-3381](https://issues.apache.org/jira/browse/BIGTOP-3381) should be the right fix?

iwasakims:
I meant that specifying livy only does not work. We need `--stack spark,livy` always.

```
$ ./docker-hadoop.sh -c 1 --stack livy
...
Error: Could not find resource 'Package[spark-core]' in parameter 'require' (file: /bigtop-home/bigtop-deploy/puppet/modules/livy/manifests/init.pp, line: 46) on node 7fee5c6931a5.bigtop.apache.org
```

Including spark::common class in livy's manifests should fix this.

```
class livy {

  class deploy ($roles) {
    if ('livy-server' in $roles) {
      include livy::server
      include spark::common
    }
  }
```

sekikn:
> Addressing BIGTOP-3381 should be the right fix?

That's totally right, but it requires to rebuild packages for all distros/platforms and signing them. It will delay our release about a week again... So I'd like to address this issue within the puppet manifest as a workaround at this time. Is this OK for you?
If the RC0 vote fails and we have to rebuild the packages, I'll address BIGTOP-3381.

sekikn:
Updated the PR so that we don't have to specify Spark as a argument for the `--stack` option explicitly.

iwasakims (approved):
+1

sekikn:
Thank you for the review, merged into master and branch-1.5.

PR #713, opened 2021-01-19
BIGTOP-3476: Fix build Oozie issues

guyuqi:
Fail to build Oozie on all distros:
1. commons.lang was deprecated and need to upgrade to commons.lang3
2. For HADOOP-11356, it's needed to replace 'org.apache.hadoop.fs.permission.AccessControlException'
   with 'org.apache.hadoop.security.AccessControlException'.


iwasakims:
Thanks for working on this, @guyuqi. There is discussion about dropping Oozie or not in [BIGTOP-3471](https://issues.apache.org/jira/browse/BIGTOP-3471). If you still need Oozie, you should declare needs there.

bigtop-packages/src/common/oozie/patch5-HADOOP-11356.diff
```diff
@@ -0,0 +1,54 @@
+diff --git a/core/src/main/java/org/apache/oozie/action/hadoop/JavaActionExecutor.java b/core/src/main/java/org/apache/oozie/action/hadoop/JavaActionExecutor.java
+index 31d4817cb..c2233d1d8 100644
+--- a/core/src/main/java/org/apache/oozie/action/hadoop/JavaActionExecutor.java
++++ b/core/src/main/java/org/apache/oozie/action/hadoop/JavaActionExecutor.java
+@@ -45,13 +45,13 @@ import org.apache.hadoop.filecache.DistributedCache;
+ import org.apache.hadoop.fs.FileStatus;
+ import org.apache.hadoop.fs.FileSystem;
+ import org.apache.hadoop.fs.Path;
+-import org.apache.hadoop.fs.permission.AccessControlException;
```
iwasakims, line 9: This was fixed by [OOZIE-2733](https://issues.apache.org/jira/browse/OOZIE-2733). Upgrading Oozie should be the right fix?

bigtop-packages/src/common/oozie/patch4-commons-lang3.diff
```diff
@@ -0,0 +1,501 @@
+diff --git a/core/src/main/java/org/apache/oozie/CoordinatorEngine.java b/core/src/main/java/org/apache/oozie/CoordinatorEngine.java
+index cec634790..c5b1f091e 100644
+--- a/core/src/main/java/org/apache/oozie/CoordinatorEngine.java
++++ b/core/src/main/java/org/apache/oozie/CoordinatorEngine.java
+@@ -35,7 +35,7 @@ import java.util.Map;
+ import java.util.Set;
+ import java.util.StringTokenizer;
+ 
+-import org.apache.commons.lang.StringUtils;
++import org.apache.commons.lang3.StringUtils;
```
iwasakims, line 10: This was addressed by [OOZIE-3446](https://issues.apache.org/jira/browse/OOZIE-3446). Bumping Oozie to 5.2.0 could be an alternative fix.

guyuqi:
@iwasakims thanks for your comments.

As posted comments on BIGTOP-3471:
if we don't foresee any opposition, let me bump oozie to 5.2.0 rather than drop it.

But before that, let's merged this Pr to fix build Oozie issues which was reported in https://ci.bigtop.apache.org/job/Bigtop-trunk-packages/678/

elukey:
> Thanks for working on this, @guyuqi. There is discussion about dropping Oozie or not in [BIGTOP-3471](https://issues.apache.org/jira/browse/BIGTOP-3471). If you still need Oozie, you should declare needs there.

I'll follow up in there, thanks for the reminder! My team is an heavy user of Oozie, so we'd appreciate to have it fixed or upgraded for the next versions, even if we are planning to move away from it soon-ish in favor of Airflow. Not sure how many people still use Oozie, it could be an interesting thing to send to the users@ mailing list? (I doubt that all users are aware of BIGTOP-3471).

guyuqi:
@elukey 
Yes, it makes sense.  It's sensible that we would broadcast the definition of bom on `users@` and `dev@`  mailing lists before making the decision.

iwasakims:
The file name patch5-HADOOP-11356.diff implies that it is a diff of HADOOP-11356. Could you rename it to patch5-OOZIE-2733.diff implying the backport of OOZIE-2733.

guyuqi:
> The file name patch5-HADOOP-11356.diff implies that it is a diff of HADOOP-11356. Could you rename it to patch5-OOZIE-2733.diff implying the backport of OOZIE-2733.

Updated, also renamed patch4 to` patch4-OOZIE-3446.diff`, thanks.

iwasakims (approved):
+1.

PR #714, opened 2021-01-21
BIGTOP-3477: Fix puppet manifests for hadoop-3

guyuqi:
Fix puppet-deployment of Hadoop warning，PID patch mismatching，disable unnecessary dir-detection, etc.


iwasakims (approved):
+1. It worked on CentOS 8.

PR #711, opened 2021-01-18
BIGTOP-3474. Bump Hadoop to 3.2.2.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3474

* removed the [HADOOP-16739](https://issues.apache.org/jira/browse/HADOOP-16739) patch included in 3.2.2.
* added patch reverting [HADOOP-16598](https://issues.apache.org/jira/browse/HADOOP-16598) for the same reason of [BIGTOP-3449](https://issues.apache.org/jira/browse/BIGTOP-3449).

PR #715, opened 2021-01-25
BIGTOP-3283. Bump HBase to 2.2.6.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3283

* This is the rebased #585.
* I chose the newest in hbase-2.2 here as a first step since hbase-2.3.x is based on zookeeper 3.5.
* spotbugs-annotation must be excluded from transitive dependencies of zookeeper for the reason same of https://github.com/apache/bigtop/pull/711#issuecomment-762551679.
* I'm addressing only bigtop-package here. bigtop-deploy and bigtop-tests might be fixed in follow-up.


iwasakims:
HBase processes failed to start due to NoSuchMethodError when I launched cluster by docker provisioner. It looks like version mismatch of Guava.
```
Exception in thread "main" java.lang.NoSuchMethodError: com.google.common.base.Preconditions.checkArgument(ZLjava/lang/String;Ljava/lang/Object;)V
        at org.apache.hadoop.conf.Configuration.set(Configuration.java:1357)
        at org.apache.hadoop.conf.Configuration.set(Configuration.java:1338)
        at org.apache.hadoop.conf.Configuration.setBoolean(Configuration.java:1679)
        at org.apache.hadoop.util.GenericOptionsParser.processGeneralOptions(GenericOptionsParser.java:339)
        at org.apache.hadoop.util.GenericOptionsParser.parseGeneralOptions(GenericOptionsParser.java:572)
        at org.apache.hadoop.util.GenericOptionsParser.<init>(GenericOptionsParser.java:174)
        at org.apache.hadoop.util.GenericOptionsParser.<init>(GenericOptionsParser.java:156)
        at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:70)
        at org.apache.hadoop.hbase.util.ServerCommandLine.doMain(ServerCommandLine.java:149)
        at org.apache.hadoop.hbase.master.HMaster.main(HMaster.java:2960)
```

sekikn:
Tested with CentOS 8, Debian 10 and Ubuntu 18.04 on x86_64. Just merged, thanks @iwasakims!

iwasakims:
Thanks, @sekikn. I filed [BIGTOP-3478](https://issues.apache.org/jira/browse/BIGTOP-3478) as a follow-up.

PR #716, opened 2021-01-27
BIGTOP-3472. Building Hadoop 3.2.1 fails on CentOS 7 due to cmake version mismatch.

sekikn:


sekikn:
Tested on CentOS 7 and 8.

iwasakims (approved):
+1

iwasakims:

PR #717, opened 2021-01-28
BIGTOP-3479. Bump Tez to 0.10.0.

sekikn:


sekikn:
Tested with CentOS 8, Debian 10 and Ubuntu 18.04 on x86_64.

iwasakims (approved):
LGTM. [HIVE-23930](https://issues.apache.org/jira/browse/HIVE-23930) bumpted the Tez version without code change. We should move on to checking the effect on Hive side.

PR #718, opened 2021-01-28
BIGTOP-3480. Bump Hive to 3.1.2.

sekikn:


sekikn:
This should be merged after #717, because it contains a patch for addressing Tez upgrade.

iwasakims (approved):
+1. I tested this on CentOS 8. packaging and docker provisioner worked. I ran hive command and simple query in the container. Thanks, @sekikn.

tasanuma:
There's something I'd like to ask. Is there any reason that it had to apply HIVE-21569 and upgrade guava version of Hive?

iwasakims:
@tasanuma
removing the patch causes the following compilation error.

```
$ git checkout rel/3.0.0
$ rm bigtop-packages/src/common/hive/patch3-HIVE-21569.diff
$ ./gradlew hive-clean hive-pkg
...
+ mvn -Dhbase.version=2.2.6 -Dzookeeper.version=3.4.14 -Dhadoop.version=3.2.2 -DskipTests -Dtez.version=0.10.0 -Dspark.version=3.0.1 -Dscala.binary.version=2.12 -Dscala.version=2.12.13 -Dguava.version=27.0-jre clean install -Pdist
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.1:compile (default-compile) on project hive-llap-common: Compilation failure: Compilation failure:
[ERROR] /home/centos/srcs/bigtop/build/hive/rpm/BUILD/apache-hive-3.1.2-src/llap-common/src/java/org/apache/hadoop/hive/llap/AsyncPbRpcProxy.java:[173,16] method addCallback in class com.google.common.util.concurrent.Futures cannot be applied to given types;
[ERROR]   required: com.google.common.util.concurrent.ListenableFuture<V>,com.google.common.util.concurrent.FutureCallback<? super V>,java.util.concurrent.Executor
[ERROR]   found: com.google.common.util.concurrent.ListenableFuture<U>,org.apache.hadoop.hive.llap.AsyncPbRpcProxy.ResponseCallback<U>
[ERROR]   reason: cannot infer type-variable(s) V
[ERROR]     (actual and formal argument lists differ in length)
[ERROR] /home/centos/srcs/bigtop/build/hive/rpm/BUILD/apache-hive-3.1.2-src/llap-common/src/java/org/apache/hadoop/hive/llap/AsyncPbRpcProxy.java:[274,12] method addCallback in class com.google.common.util.concurrent.Futures cannot be applied to given types;
[ERROR]   required: com.google.common.util.concurrent.ListenableFuture<V>,com.google.common.util.concurrent.FutureCallback<? super V>,java.util.concurrent.Executor
[ERROR]   found: com.google.common.util.concurrent.ListenableFuture<java.lang.Void>,<anonymous com.google.common.util.concurrent.FutureCallback<java.lang.Void>>
[ERROR]   reason: cannot infer type-variable(s) V
[ERROR]     (actual and formal argument lists differ in length)
```

tasanuma:
@iwasakims Thank you for your quick reply. It seems that the `-Dguava.version=27.0-jre` option is the cause of the compilation failure. So, is `HIVE-21569.diff` required because of #763?

tasanuma:
@iwasakims  Thank you for the confirmation! Now I got it. Since it specifies `-Dhadoop.version=3.2.2`, it also needs to conform to the guava version of hadoop-3.2.2, or such problems will occur.

PR #719, opened 2021-01-29
BIGTOP-3482. Fix toolchain not to break cmake by creating unnecessary symlink.

sekikn:


sekikn:
Tested on CentOS 7, 8, Debian 10 and Ubuntu 18.04.

iwasakims (approved):
+1. Thanks for the follow-up, @sekikn.

PR #724, opened 2021-02-05
BIGTOP-3484. Bump Alluxio to 2.4.1.

sekikn:


sekikn:
Tested on CentOS 7 and Debian 10.

iwasakims (approved):
LGTM. packaging and smoke-tests worked on CentOS 8.

The alluxio jars like alluxio-assembly-server-2.4.1-jar-with-dependencies.jar and alluxio-assembly-client-2.4.1-jar-with-dependencies.jar contain dependency classes including Hadoop, ZooKeeper, Hive. We may need follow-up fix it the version mismatch causes issue, while existing smoke-tests did not complain.
* https://github.com/Alluxio/alluxio/blob/v2.4.1/pom.xml#L573-L584
* https://github.com/Alluxio/alluxio/blob/v2.4.1/integration/tools/hms/pom.xml#L34

PR #726, opened 2021-02-06
BIGTOP-3490. Bump Sqoop to 1.4.7.

sekikn:


sekikn:

iwasakims (approved):
+1. Thanks, @sekikn. I built rpm on CentOS 8 and ran smoke-test with docker provisioner successfully.

The puppet manifest kicked by `--stack hdfs,yarn,mapreduce,sqoop` tried to install sqoop2 too. We need to fix them on the JIRA dropping sqoop2.

PR #727, opened 2021-02-08
BIGTOP-3488. Bump Kafka to 2.4.1.

sekikn:


sekikn:

iwasakims:
@sekikn. Should we patch dependencies.gradle in order to set the zookeeper version to 3.4.14? It would be useful to find compilation issue and pull exact transitive dependencies of zookeeper-3.4.14.

```
diff --git a/gradle/dependencies.gradle b/gradle/dependencies.gradle
index 7597b9c52..dab290372 100644
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -114,7 +114,7 @@ versions += [
   spotbugs: "3.1.12",
   spotbugsPlugin: "1.6.9",
   spotlessPlugin: "3.23.1",
-  zookeeper: "3.5.7",
+  zookeeper: "3.4.14",
   zstd: "1.4.3-1"
 ]
```

sekikn:
Oh, that's right. I mistakenly thought that the zookeeper version was specified in do-component-build just like other components. Thanks!

sekikn:
Updated the PR. I chose rewriting dependencies.gradle in do-component-build rather than putting a patch in bigtop-packages/src/common/kafka, so that we can leverage `$ZOOKEEPER_VERSION` instead of a hard-coded value.
I confirmed the right version of ZK was used by checking the following message, which is output by install_kafka.sh.

```
+ rm -f debian/kafka//usr/lib/kafka/libs/zookeeper-3.4.14.jar
```

iwasakims (approved):
+1. Thanks, @sekikn.

PR #728, opened 2021-02-08
BIGTOP-3491. Add Ubuntu 20.04 support to the Bigtop toolchain.

sekikn:


sekikn:
I tested that `sudo bigtop_toolchain/bin/puppetize.sh`, `./gradlew toolchain`, and (after relogin) `/gradlew zookeeper-pkg -Dbuildwithdeps=true` worked with this PR on Ubuntu 20.04.

bigtop_toolchain/manifests/packages.pp
```diff
@@ -204,7 +204,6 @@
         "bison",
         "flex",
         "python-dev",
-        "python-pip",
```
iwasakims, line 207: Can we remove pip (for Python 2) from all distributions?

sekikn: Yes I think so. Python2 is still used in some components, but pip2 is not as far as I can see.

iwasakims: I think we should see the effect on Ubuntu/Debian first by committing this as is.

iwasakims (approved):
+1.

PR #730, opened 2021-02-08
BIGTOP-3382. Upgrade Flink to 1.11.3

okumurays:
I confirmed this patch worked on CentOS 8 at least.

The following log is smoke test running log.

```
../../../gradlew test -Psmoke.tests --stacktrace

> Task :bigtop-tests:smoke-tests:flink:test
Now testing...

BUILD SUCCESSFUL in 46s
6 actionable tasks: 6 executed
```


sekikn:
Thank you for the contribution @okumurays! I confirmed the PR worked on CentOS 7 and 8, Debian 10 and Ubuntu 18.04.
I noticed `testWordCountBatch` in the smoke tests sometimes failed due to `UnsupportedFileSystemSchemeException` for HDFS when Flink was deployed earlier than Hadoop, but once restarting flink services it succeeded.
We can address it later as another issue, so I'm +1 for this PR.

PR #731, opened 2021-02-09
BIGTOP-3494. Add MapReduce settings to yarn.nodemanager.env-whitelist in yarn-site.xml.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3494

MapReduce job does not work without adding HADOOP_MAPRED_HOME to the value of yarn.nodemanager.env-whitelist. testDistcpIntra and testDistcpIntra_MultipleSources reproducibly fails after upgrading hadoop to 3.2 due to this.

### DIFF
--- a/bigtop-deploy/puppet/modules/hadoop/templates/yarn-site.xml
+++ b/bigtop-deploy/puppet/modules/hadoop/templates/yarn-site.xml
@@ -156,6 +156,11 @@
   </property>
 
   <property>
+    <name>yarn.nodemanager.env-whitelist</name>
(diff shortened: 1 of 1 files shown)


iwasakims:
`./docker-hadoop.sh --smoke-tests hdfs,yarn,mapreduce` passed on 3-node docker provisioned cluster on CentOS 8.

sekikn:
+1, I also confirmed the above command succeeded with this PR on CentOS 7, Debian 10, and Ubuntu 18.04 using a three-node cluster. Thanks @iwasakims!

PR #735, opened 2021-02-16
BIGTOP-3400. Clean up or disable husky installed via packaging Kibana.

sekikn:


sekikn:
I ran `./gradlew kibana-clean kibana-pkg` with this PR locally and confirmed that the build succeeded and any hook such as pre-push wasn't installed under .git/hooks.

iwasakims (approved):
LGTM. The dependency on husky was removed in https://github.com/elastic/kibana/pull/24887. Removing it seems not to affect packaging of Bigtop.

PR #736, opened 2021-02-16
BIGTOP-3495. Optional tools of hadoop-tools should be optional.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3495

* [HADOOP-12857](https://issues.apache.org/jira/browse/HADOOP-12857) made tools such as hadoop-aws optional in order to reduce load lots of dependencies by default.
* Filesystem modules such as hadoop-aws and hadoop-azure should not belong to hadoop-mapreduce. Those does not depend on MR and usable via hadoop fs CLI or FileSystem API.


iwasakims:
I tested this on docker provisioned CentOS 8 env. hadoop-aws is loaded only if HADOOP_OPTIONAL_TOOLS contains the entry. HADOOP_OPTIONAL_TOOLS can be specified via command line or configuration files such as hadoop-env.sh.

```
# hadoop fs -ls s3a://iwasakims-test/
-ls: Fatal internal error
java.lang.RuntimeException: java.lang.ClassNotFoundException: Class org.apache.hadoop.fs.s3a.S3AFileSystem not found
        at org.apache.hadoop.conf.Configuration.getClass(Configuration.java:2638)
        at org.apache.hadoop.fs.FileSystem.getFileSystemClass(FileSystem.java:3341)


# HADOOP_OPTIONAL_TOOLS=hadoop-aws hadoop fs -ls s3a://iwasakims-test/
2021-02-16 08:45:07,177 INFO impl.MetricsConfig: Loaded properties from hadoop-metrics2.properties
2021-02-16 08:45:07,239 INFO impl.MetricsSystemImpl: Scheduled Metric snapshot period at 10 second(s).
2021-02-16 08:45:07,239 INFO impl.MetricsSystemImpl: s3a-file-system metrics system started
Found 4 items
-rw-rw-rw-   1 root root      15841 2020-10-21 23:35 s3a://iwasakims-test/NOTICE.txt
-rw-rw-rw-   1 root root       1366 2020-10-21 23:35 s3a://iwasakims-test/README.txt
drwxrwxrwx   - root root          0 2021-02-16 08:45 s3a://iwasakims-test/bar
drwxrwxrwx   - root root          0 2021-02-16 08:45 s3a://iwasakims-test/foo
2021-02-16 08:45:08,157 INFO impl.MetricsSystemImpl: Stopping s3a-file-system metrics system...
2021-02-16 08:45:08,157 INFO impl.MetricsSystemImpl: s3a-file-system metrics system stopped.
2021-02-16 08:45:08,157 INFO impl.MetricsSystemImpl: s3a-file-system metrics system shutdown complete.
```

iwasakims:
I got expected result on Ubuntu 18.04 too.

sekikn:
+1, also tested on Debian 10. Thanks @iwasakims!

PR #737, opened 2021-02-18
BIGTOP-3500. Bump Zeppelin to 0.9.0.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3500


sekikn:
I confirmed the followings with this PR:

* Building Zeppelin and its smoke tests succeeded, at least on CentOS 7 and 8 (it's also supposed to succeed on Debian 10 and Ubuntu 18.04).
* The following interpreters just worked:

  * %spark (though I had to specify both `spark.master` and `spark.submit.deployMode` separately for Spark 3. See https://zeppelin.apache.org/docs/0.9.0/interpreter/spark.html#configure-yarn-mode-for-spark-3x)
  * %livy.spark
  * %file (HDFS)
  * %python
  * %r (installing knitr is required, or disable the zeppelin.R.knitr flag)
  * %sh

For other components (Alluxio, Elasticsearch, Flink, HBase, Phoenix), Zeppelin doesn't work well only by specifying component versions via the -D option, due to version/API/protocol mismatch etc. But addressing them seems to require some time, so I'd like to postpone them as follow-up issues and submit this PR as the first step.

iwasakims (approved):
+1. packaging and smoke-tests worked on Ubuntu 18.04 too. Thanks, @sekikn.

PR #738, opened 2021-02-19
BIGTOP-3501. Update hadoop-kms resources based on hadoop-3.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3501

Resrouces for hadoop-kms such as init scripts and service configuration should be updated based on hadoop-3.

I'm addressing only packaging issue here. I manually tested that `systemctl start hadoop-kms` and `hadoop key create` work with primitive configuration on CentOS 8.

I'm going to fix puppet manifests in [BIGTOP-3505](https://issues.apache.org/jira/browse/BIGTOP-3505).


iwasakims:
> I manually tested that systemctl start hadoop-kms and hadoop key create work with primitive configuration on CentOS 8.

got expected result on Ubuntu 18.04 too.

sekikn:
+1, also tested on CentOS 7 and Debian 10.
I deployed the built package with docker provisioner ignoring path errors and had to rewrite the KMS port (16000 -> 9600) in core-site.xml, but they should be fixed in BIGTOP-3505.

PR #739, opened 2021-02-21
BIGTOP-3505. Update hadoop-kms puppet manifests based on hadoop-3.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3505

Current Puppet manifests of hadoop-kms assumes hadoop-2. It does not work for hadoop-3 without fix.


iwasakims:
I tested this on CentOS 8 by following cmdline:

```
./docker-hadoop.sh \
  --create 1 \
  --memory 16g \
  --disable-gpg-check \
  --stack hdfs,kms \
  --smoke-tests kms
```

sekikn:
+1, smoke test also succeeded on CentOS 7, Debian 10 and Ubuntu 18.04. Thanks @iwasakims!

PR #744, opened 2021-02-27
BIGTOP-3511. Drop Mahout.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3511

Drop Mahout based on the discussion in [BIGTOP-3471](https://issues.apache.org/jira/browse/BIGTOP-3471)

I intentionally left files under bigtop-bigpetstore untouched because it does not affect our CI and it contains contents about lots of already removed products. I will file a follow-up JIRA to clean up the bigtop-bigpetstore if I need.


sekikn:
+1, merged into master. @iwasakims Would you file a follow-up JIRA to fix bigtop-bigpetstore to work without the removed components such as Mahout and Pig?

iwasakims:
Thanks, @sekikn. I filed [BIGTOP-3518](https://issues.apache.org/jira/browse/BIGTOP-3518).

PR #745, opened 2021-02-28
BIGTOP-3508: drop /usr/share/java from sqoop's bin classpath

elukey:
On Debian 10 the package libslf4j-java deploys various jars to the
system, including log4j-over-slf4j.jar or slf4j-log4j12.jar that cannot
be added to the classpath together (see http://www.slf4j.org/codes.html#log4jDelegationLoop).
From my tests it seems that dropping /usr/share/java from the sqoop's
classpath is safe, I didn't notice any regression.

Co-authored-by: Kengo Seki <sekikn@apache.org>


iwasakims (approved):
+1. Thanks, @elukey and @sekikn.

I tested the patch on CentOS 8 to make sure there is no harm while there is nothing under /usr/share/java of CentOS 8 contianer.
```
./docker-hadoop.sh \
    --create 3 \
    --memory 8g \
    --disable-gpg-check \
    --stack hdfs,yarn,mapreduce,sqoop \
    --smoke-tests sqoop
```

PR #746, opened 2021-03-03
BIGTOP-3514. Bump GPDB to 5.28.5.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3514

### DIFF
--- a/bigtop-deploy/puppet/modules/gpdb/manifests/init.pp
+++ b/bigtop-deploy/puppet/modules/gpdb/manifests/init.pp
@@ -40,7 +40,7 @@ class gpdb {
       gp_home             => $gp_home,
       master_port         => $master_db_port,
       segment_port_prefix => $segment_db_port_prefix,
-      require             => [Class['gpdb::common::prepare_file_structure'], Package["gpdb"]]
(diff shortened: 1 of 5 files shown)


sekikn:

iwasakims (approved):
+1. It worked for me too.

The smoke-test of gpdb seems to (unnecessarily?) depend on Java. Installing products using Java like `--stack hdfs,gpdb` fixed the issue. This could be addressed in follow-up JIRAs.
```
$ ./docker-hadoop.sh \
    --create 1 \
    --memory 16g \
    --disable-gpg-check \
    --stack gpdb \
    --smoke-tests gpdb
...
+ HCFS_USER=hdfs
+ SMOKE_TESTS=gpdb
+ '[' -z gpdb ']'
+ '[' -e /usr/lib/bigtop-utils/bigtop-detect-javahome ']'
+ echo -e '\nUNABLE TO DETECT JAVAHOME SINCE bigtop-utils NEEDS TO BE INSTALLED!\n'

UNABLE TO DETECT JAVAHOME SINCE bigtop-utils NEEDS TO BE INSTALLED!

+ exit 2
```

PR #749, opened 2021-03-03
BIGTOP-3515 - Backport HIVE-19231

elukey:
This patch is contained in Hive 3.x, but not in the Hive 2.x branch. It
leads to subtle errors when executing Hive via Python or systemd, and
having it fixed in the 1.5 branch would allow people to rebuild their
packages if needed (we could update the docs about it).

### DIFF
--- a/bigtop-packages/src/common/hive/patch6-HIVE-19231.diff
+++ b/bigtop-packages/src/common/hive/patch6-HIVE-19231.diff
@@ -1,0 +1,14 @@
+diff --git a/bin/hive b/bin/hive
+index a7671c3950..508e73da75 100755
+--- a/bin/hive
++++ b/bin/hive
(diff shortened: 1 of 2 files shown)


elukey:
I didn't open a specific BIGTOP jira issue but I can if needed, and I am testing this patch in more detail right now (the build part I mean). The code works fine on the production bigtop 1.5 cluster :)

elukey:
I have adapted the Hive patch to the 2.x branch and built with:

docker run --rm  -v `pwd`:/ws --workdir /ws bigtop/slaves:1.5.0-debian-10 bash -c '. /etc/profile.d/bigtop.sh; ./gradlew allclean hive-pkg'

iwasakims (approved):
LGTM. This did not harm packaging and usual interactive session by hive and beeline.

PR #750, opened 2021-03-05
BIGTOP-3516. Remove dead HBase jar link for Hive.

jshmchenxi:
https://issues.apache.org/jira/browse/BIGTOP-3516


iwasakims (approved):
+1. Thanks, @jshmchenxi.

PR #754, opened 2021-03-11
(BIGTOP-3525) Replace hiera() with lookup()

elukey:
I noticed the following while trying to run the docker provisioner:

Warning: The function 'hiera' is deprecated in favor of using 'lookup'.
See https://puppet.com/docs/puppet/5.5/deprecated_language.html

I know that we need to keep compatibility with older OSes, but with Bigtop 3.x we may be able to have a more up to date puppet version available.

### DIFF
--- a/bigtop-deploy/puppet/hieradata/bigtop/cluster.yaml
+++ b/bigtop-deploy/puppet/hieradata/bigtop/cluster.yaml
@@ -71,15 +71,15 @@
 #hadoop::hadoop_security_authentication: "kerberos"
 #kerberos::krb_site::domain: "bigtop.apache.org"
 #kerberos::krb_site::realm: "BIGTOP.APACHE.ORG"
-#kerberos::krb_site::kdc_server: "%{hiera('bigtop::hadoop_head_node')}"
+#kerberos::krb_site::kdc_server: "%{lookup('bigtop::hadoop_head_node')}"
 #kerberos::krb_site::kdc_port: "88"
 #kerberos::krb_site::admin_port: "749"
 #kerberos::krb_site::keytab_export_dir: "/var/lib/bigtop_keytabs"
 
 # applies to hdfs, yarn, mapred, kms and httpfs
-hadoop::kerberos_realm: "%{hiera('kerberos::krb_site::realm')}"
+hadoop::kerberos_realm: "%{lookup('kerberos::krb_site::realm')}"
 
-hadoop::common_hdfs::hadoop_namenode_host: "%{hiera('bigtop::hadoop_head_node')}"
+hadoop::common_hdfs::hadoop_namenode_host: "%{lookup('bigtop::hadoop_head_node')}"
 # actually default but needed for hadoop_namenode_uri here
 hadoop::common_hdfs::hadoop_namenode_port: "8020"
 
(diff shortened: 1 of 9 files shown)


iwasakims:
@elukey Could you file a JIRA issue and put the id (BIGTOP-xxxx) on the title of this PR?
https://cwiki.apache.org/confluence/display/BIGTOP/How+to+Contribute

iwasakims:
I got following error on CentOS 7 with the patch. Could you take a look? Puppet 3.6.2 was install on CentOS 7 by bigtop_toolchain/bin/puppetize.sh.

```
$ ./docker-hadoop.sh --create 1 --image bigtop/puppet:trunk-centos-7 --memory 16g --repo file:///bigtop-home/output --disable-gpg-check --stack hdfs,mapreduce,yarn
...
Error: Function lookup() did not find a value for the name 'bigtop::jdk_preinstalled' on node 8ae2b15d58b0.bigtop.apache.org
Error: Function lookup() did not find a value for the name 'bigtop::jdk_preinstalled' on node 8ae2b15d58b0.bigtop.apache.org

$ ./docker-hadoop.sh --exec 1 puppet --version
3.6.2
```

elukey:
> @elukey Could you file a JIRA issue and put the id (BIGTOP-xxxx) on the title of this PR?
> https://cwiki.apache.org/confluence/display/BIGTOP/How+to+Contribute

Definitely: BIGTOP-3525

elukey:
> I got following error on CentOS 7 with the patch. Could you take a look? Puppet 3.6.2 was install on CentOS 7 by bigtop_toolchain/bin/puppetize.sh.
> 
> ```
> $ ./docker-hadoop.sh --create 1 --image bigtop/puppet:trunk-centos-7 --memory 16g --repo file:///bigtop-home/output --disable-gpg-check --stack hdfs,mapreduce,yarn
> ...
> Error: Function lookup() did not find a value for the name 'bigtop::jdk_preinstalled' on node 8ae2b15d58b0.bigtop.apache.org
> Error: Function lookup() did not find a value for the name 'bigtop::jdk_preinstalled' on node 8ae2b15d58b0.bigtop.apache.org
> 
> $ ./docker-hadoop.sh --exec 1 puppet --version
> 3.6.2
> ```

This is very strange, I can repro (but not on Debian 10). In theory bigtop::jdk_preinstalled has a default "false", no idea why it yields this error, will keep checking!

elukey:
With a more fresh mind, Debian 10 installs puppet 5.5, so lookup in there is probably different and not comparable. It seems as if the lookup in puppet 3.6.2 doesn't really accept the default_value, ending up in not finding anything. 

Debian 9 has also different problems (with puppet 4), see:

```
Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, Lookup of key 'hadoop::kerberos_realm' failed: DataBinding 'hiera': Invalid interpolation method 'lookup' at /bigtop-home/bigtop-deploy/puppet/modules/hadoop/manifests/init.pp:35:7  at /bigtop-home/bigtop-deploy/puppet/manifests/cluster.pp:230 on node b27e057e3765.bigtop.apache.org
```

Will  try to see if anything can be done, but maybe it is a task too difficult for the broad range of OSes supported, not sure :(

elukey:
I am going to discard this there were a lot of efforts to migrate to puppet 5.x but we decided to avoid it:
https://issues.apache.org/jira/browse/BIGTOP-3161
https://lists.apache.org/thread.html/26e14cf36e9cfd61e0de581ed83bf305565c2e65234f1ce3bfb97628@%3Cdev.bigtop.apache.org%3E

I hoped to be able to move to lookup but the function varies too much from puppet 3 to 5, and adding a ton of conditionals in the code to support both seems only polluting the code with no benefit..

PR #755, opened 2021-03-11
BIGTOP-3519. Drop Flume.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3519

Drop Flume based on the discussion in [BIGTOP-3471](https://issues.apache.org/jira/browse/BIGTOP-3471).

I intentionally left resources of juju, charm and ambari as is since they look not tightly coupled to Bigtop packages.


iwasakims:
I ran following to check that flume removal does not harm resources of other components on CentOS 8.

```
$ ./gradlew hadoop-clean hadoop-pkg repo
$ cd provisioner/docker
$ ./docker-hadoop.sh \
    --create 3 \
    --memory 8g \
    --disable-gpg-check \
    --stack hdfs,mapreduce,yarn \
    --smoke-tests hdfs,mapreduce,yarn
```

sekikn:
+1 merged into master. Thanks @iwasakims!

PR #756, opened 2021-03-14
BIGTOP-3520. Drop ignite-hadoop.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3520


iwasakims:
I tested Zeppelin packaging and smoke-tests after the ignite-hadoop removal.

```
./docker-hadoop.sh \
  --create 1 \
  --memory 16g \
  --disable-gpg-check \
  --stack hdfs,yarn,mapreduce,spark,zeppelin \
  --smoke-tests zeppelin
```

sekikn:
It seems that ignite-hadoop is still in the "component" section in bigtop.bom even with PR.

iwasakims:
> It seems that ignite-hadoop is still in the "component" section in bigtop.bom even with this PR.

Oops. Thanks, @sekikn . I addressed the comment and rebased.

sekikn:
+1, thanks @iwasakims!

PR #757, opened 2021-03-15
BIGTOP-3521. Drop QFS.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3521


sekikn:
I think we can also remove QFS-specific logic in provisioner/utils/smoke-tests.sh.

iwasakims:
Thanks, @sekikn. I updated the smoke-tests.sh and groovy scripts of mapreduce smoke-tests.

I tested the change by running mapreduce smoke-tests via docker-hadoop.sh.
```
./docker-hadoop.sh \
 --create 3 \
 --memory 8g \
 --disable-gpg-check \
 --stack hdfs,yarn,mapreduce \
 --smoke-tests hdfs,yarn,mapreduce
```

sekikn:
Thank you for updating the PR @iwasakims!
I've just noticed there's still QFS-specific logic in bigtop-tests/test-artifacts/hadoop/src/main/groovy/org/apache/bigtop/itest/hadoop/mapreduce/TestHadoopExamples.groovy.
Would you fix that too? Sorry for missing that in the first review.

iwasakims:
I updated the TestHadoopExamples.groovy and ran the smoke-tests again.

bigtop-tests/test-artifacts/hadoop/src/main/groovy/org/apache/bigtop/itest/hadoop/mapreduce/TestHadoopExamples.groovy
```diff
@@ -152,9 +143,7 @@ class TestHadoopExamples {
   public static LinkedHashMap<String, Object[]> generateTests() {
     LinkedHashMap<String, Object[]> res = [:];
     examples.each { k, v -> res[k] = [k.toString(), v.toString()] as Object[]; }
-    if (HADOOP_COMMAND != "hadoop-qfs") {
-      additional_examples.each { k, v -> res[k] = [k.toString(), v.toString()] as Object[]; }
-    }
+    additional_examples.each { k, v -> res[k] = [k.toString(), v.toString()] as Object[]; }
```
sekikn, line 146: We can merge `additional_examples` into `examples`, because the former was separated from the latter so as to addressing QFS test failure originally. See the comment at line 129.

iwasakims: Sure. I merged the maps.

sekikn:
+1, I also confirmed that YARN and MR smoke tests passed on CentOS 7. Thanks @iwasakims!

PR #759, opened 2021-03-16
BIGTOP-3526: Remove phantomjs dependency from yarn-ui

JunHe77:
Haoop Yarn-UI depends on phantomjs (optional) while there is
not prebuilt binary other than x86 platform. And this will
cause hadoop build failure as yarn-ui has been enabled by BIGTOP-3456.

Change-Id: I10d034d3172787f9667d53d44b1663a04c9fdf93

### DIFF
--- a/bigtop-packages/src/common/hadoop/patch7-remove-phantomjs-in-yarn-ui.diff
+++ b/bigtop-packages/src/common/hadoop/patch7-remove-phantomjs-in-yarn-ui.diff
@@ -1,0 +1,43 @@
+diff --git a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/src/main/webapp/yarn.lock b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/src/main/webapp/yarn.lock
+index a9b3ab758c3..94b6f4b5e04 100644
+--- a/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/src/main/webapp/yarn.lock
++++ b/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/src/main/webapp/yarn.lock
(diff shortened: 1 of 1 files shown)


bigtop-packages/src/common/hadoop/do-component-build
```diff
@@ -125,6 +125,8 @@ MAVEN_OPTS+="-DskipTests -DskipTest -DskipITs "
 . $(dirname ${0})/maven_deploy.sh
 
 # Build artifacts
+# remove phantomjs dependency
+sed -i '/^phantomjs-prebuilt/,+12d;N;/phantomjs/!P;D' hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/src/main/webapp/yarn.lock
```
iwasakims, line 129: Should we use `patch*.diff` file rather than ad hoc editing in the do-component-build? I think the sed scripting is possible to modify the file in unexpected way if the yarn.lock is updated (on Hadoop side).

iwasakims:
Since the PhantomJS is only used for testing, removing it would be ok. Yarn UI 2 built with the modified yarn.lock worked for me.

bigtop-packages/src/common/hadoop/do-component-build
JunHe77: That should be a better way. I'll upload a new patchset. Thanks, @iwasakims

iwasakims (approved):
+1. Thanks, @JunHe77.

jojochuang:
(FWIW, Ubuntu Bionic and Focal have official phantomjs binary packages for ARM) HADOOP-17723 / https://github.com/apache/hadoop/pull/3037

iwasakims:
Thanks, @jojochuang. The information is useful for Bigtop too.

PR #762, opened 2021-03-29
BIGTOP-3529. Drop bigtop-tomcat.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3529


iwasakims:
I left some configuration files of Oozie under bigtop-package as is. It should be updated in [BIGTOP-3531](https://issues.apache.org/jira/browse/BIGTOP-3531).

sekikn:
+1, I confirmed that bigtop-tomcat was not in the result of `./gradlew all-components` and `./gradlew oozie-pkg -Dbuildwithdeps=true` succeeded on CentOS 7, 8 and Ubuntu 20.04 with this PR. Thanks @iwasakims!

PR #763, opened 2021-04-02
BIGTOP-3531. Fix packaging of Oozie 5.2.1.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3531

I needed lots of fix than I expected just make oozie-server run.

* fixed layout based on the expectation of scripts under /usr/lib/oozie/bin.
* made /usr/lib/oozie/lib symlink to embedded-oozie-server/webapp/WEB-INF/lib.
* set guava version to 27.0-jre.
* set JETTY_PID_FILE  to /var/run/oozie/oozie.pid.
* fixed /etc/init.d/oozie.
* excluded log4j2 from transitive dependencies of hive3 artifacts.

Let me address smoke-tests issue on [BIGTOP-3532](https://issues.apache.org/jira/browse/BIGTOP-3532) after this is merged.

### DIFF
--- a/bigtop-packages/src/common/hive/do-component-build
+++ b/bigtop-packages/src/common/hive/do-component-build
@@ -25,7 +25,8 @@ HIVE_MAVEN_OPTS=" -Dhbase.version=$HBASE_VERSION \
 -Dtez.version=${TEZ_VERSION} \
 -Dspark.version=${SPARK_VERSION} \
 -Dscala.binary.version=${SCALA_VERSION%.*} \
--Dscala.version=${SCALA_VERSION}
+-Dscala.version=${SCALA_VERSION} \
+-Dguava.version=27.0-jre
 "
 
 # Include common Maven Deployment logic
--- a/bigtop-packages/src/common/oozie/install_oozie.sh
+++ b/bigtop-packages/src/common/oozie/install_oozie.sh
@@ -103,6 +103,7 @@ BIN_DIR=${CLIENT_PREFIX}/usr/bin
 
 install -d -m 0755 ${CLIENT_LIB_DIR}
 tar --strip-components=1 -zxf ${BUILD_DIR}/oozie-client-*.tar.gz -C ${CLIENT_LIB_DIR}/
+mv ${CLIENT_LIB_DIR}/lib ${CLIENT_LIB_DIR}/lib-client
 cp ${BUILD_DIR}/oozie-examples.tar.gz ${CLIENT_LIB_DIR}/
 install -d -m 0755 ${DOC_DIR}
 mv ${CLIENT_LIB_DIR}/*.txt ${DOC_DIR}/
@@ -179,6 +180,7 @@ DATA_DIR=${SERVER_PREFIX}/var/lib/oozie
 install -d -m 0755 ${SERVER_LIB_DIR}
 install -d -m 0755 ${SERVER_LIB_DIR}/bin
 install -d -m 0755 ${SERVER_LIB_DIR}/lib
+install -d -m 0755 ${CLIENT_LIB_DIR}/lib
 install -d -m 0755 ${DATA_DIR}
 for file in ooziedb.sh oozied.sh oozie-jetty-server.sh oozie-sys.sh oozie-setup.sh ; do
   cp ${BUILD_DIR}/bin/$file ${SERVER_LIB_DIR}/bin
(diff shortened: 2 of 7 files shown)


iwasakims:
This worked on CentOS 8 but not on Ubuntu-18.04. I'm looking into the cause.

iwasakims:
The cause turned out to be hive-exec.jar containing unshaded guava classes.
```
root@afa1d143f9b8:/# cat /tmp/oozie-cp.txt | sed -z -e 's/:/\n/g' | xargs -n 1 grep 'com/google/common/base/Preconditions.class'
Binary file /usr/lib/oozie/lib/avro-1.8.2.jar matches
Binary file /usr/lib/oozie/lib/guava-27.0-jre.jar matches
Binary file /usr/lib/oozie/lib/hbase-shaded-miscellaneous-1.0.1.jar matches
Binary file /usr/lib/oozie/lib/hive-exec-3.1.2.jar matches
Binary file /usr/lib/oozie/lib/jersey-guava-2.25.1.jar matches
grep: /usr/lib/oozie/libext/*.jar: No such file or directory

root@afa1d143f9b8:/# jar tvf /usr/lib/oozie/lib/avro-1.8.2.jar | grep 'com/google/common/base/Preconditions.class'
  5173 Mon Apr 05 04:43:28 GMT 2021 avro/shaded/com/google/common/base/Preconditions.class

root@afa1d143f9b8:/# jar tvf /usr/lib/oozie/lib/hive-exec-3.1.2.jar | grep 'com/google/common/base/Preconditions.class'
  5246 Mon Apr 05 04:43:28 GMT 2021 com/google/common/base/Preconditions.class

root@afa1d143f9b8:/# jar tvf /usr/lib/oozie/lib/hbase-shaded-miscellaneous-1.0.1.jar | grep 'com/google/common/base/Preconditions.class'
 20764 Mon Apr 05 04:43:28 GMT 2021 org/apache/hadoop/hbase/shaded/com/google/common/base/Preconditions.class

root@afa1d143f9b8:/# jar tvf /usr/lib/oozie/lib/jersey-guava-2.25.1.jar | grep 'com/google/common/base/Preconditions.class'
  5281 Mon Apr 05 04:43:28 GMT 2021 jersey/repackaged/com/google/common/base/Preconditions.class
```

iwasakims:
The situation is same for CentOS 8. If classes contained in guava-27.0-jre.jar are loaded before hive-exec-3.1.2.jar by chance, there is no error on starting up.

iwasakims:
Another issue specific to Ubuntu is that oozie-client put files /usr/lib/oozie/lib. oozie-server does not create symlink to /usr/lib/oozie/embedded-oozie-server/webapp/WEB-INF/lib as a result.

iwasakims:
Bigtop applies [the patch](https://github.com/apache/bigtop/blob/master/bigtop-packages/src/common/hive/patch3-HIVE-21569.diff) of [HIVE-21569](https://issues.apache.org/jira/browse/HIVE-21569) which bump guava version of Hive. We need to do `hive-pkg` before `oozie-pkg` to install patched Hive artifacts into local repo.

While both 27.0-jre and 28.1-jre have the `Precondition#checkArgument(boolean, String, Object)`, 27.0-jre which is used by hadoop-3.2.2 should be better.
* https://guava.dev/releases/27.0-jre/api/docs/com/google/common/base/Preconditions.html#checkArgument-boolean-java.lang.String-java.lang.Object-
* https://guava.dev/releases/28.1-jre/api/docs/com/google/common/base/Preconditions.html#checkArgument-boolean-java.lang.String-java.lang.Object-

iwasakims:
I tested the update on CentOS 8 and Ubuntu 18.04.

```
$ ./docker-hadoop.sh \
  --create 1 \
  --memory 16g \
  --disable-gpg-check \
  --stack hdfs,yarn,mapreduce,oozie
...
Notice: Applied catalog in 182.76 seconds

$ ./docker-hadoop.sh --exec 1 systemctl --no-pager status oozie
● oozie.service - LSB: Oozie server daemon
   Loaded: loaded (/etc/rc.d/init.d/oozie; generated)
   Active: active (running) since Mon 2021-04-05 15:40:36 UTC; 11min ago
     Docs: man:systemd-sysv-generator(8)
    Tasks: 59 (limit: 201543)
   Memory: 780.4M
   CGroup: /docker/f780b7e2bfcb9de3e471e9c36edbe7aad1caf7022e237f7540a398ebb24fe0b3/system.slice/oozie.service
           └─8841 /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.275.b01-1.el8_3.x86_64/…
...
```

iwasakims:
oozie-client RPM alone does not work since jars pointed by symbolic link are contained in oozie (sever) package. I'm fixing this.

iwasakims:
I verified that there is no class loading issue on `oozie` command even if only oozie-client pakcage is installed on CentOS 8 and Ubuntu 18.04.
```
$ ./docker-hadoop.sh --create 1 --image bigtop/puppet:trunk-centos-8 --memory 16g --repo file:///bigtop-home/output --disable-gpg-check --stack hdfs,yarn,mapreduce,oozie

$ ./docker-hadoop.sh --exec 1 systemctl --no-pager status oozie
● oozie.service - LSB: Oozie server daemon
   Loaded: loaded (/etc/rc.d/init.d/oozie; generated)
   Active: active (running) since Tue 2021-04-06 07:29:39 UTC; 39s ago
...

$ ./docker-hadoop.sh --exec 1 rpm -ev oozie
Preparing packages...
oozie-5.2.1-1.el8.noarch
warning: /etc/oozie/conf.dist/oozie-site.xml saved as /etc/oozie/conf.dist/oozie-site.xml.rpmsave

$ ./docker-hadoop.sh --exec 1 oozie
Invalid sub-command: missing sub-command

use 'help [sub-command]' for help details
```

sekikn:
+1, I also did the same check on CentOS 7, Debian 10, and Ubuntu 20.04. Thanks @iwasakims!

iwasakims:
Thanks, @sekikn.

PR #771, opened 2021-05-14
BIGTOP-3537. Fix puppetize.sh to work with the latest CentOS 8 Docker image.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3537

PR #772, opened 2021-05-15
BIGTOP-3537. Fix puppetize.sh to work with the latest CentOS 8 Docker image (addendum).

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3537


sekikn:
I'm going to merge this since the preceding PR (#771) was incomplete.

PR #773, opened 2021-05-16
BIGTOP-3538. Building GPDB fails on Fedora 33.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3538

This PR contains the following fixes so as to satisfy recent stricter compiler:

* Backport https://github.com/greenplum-db/gpdb/pull/10663
* Remove duplicate variable declarations

Tested on all distros we support using an x86_64 machine.

### DIFF
--- a/bigtop-packages/src/common/gpdb/do-component-configure
+++ b/bigtop-packages/src/common/gpdb/do-component-configure
@@ -16,4 +16,4 @@
 
 set -ex
 
-./configure --prefix=$1 --disable-orca
+./configure --prefix=$1 --disable-orca CFLAGS=-fcommon


iwasakims:
```
/usr/bin/ld: cdb/SUBSYS.o:(.data+0x3bac): multiple definition of `gp_encoding_check_locale_compatibility'; utils/SUBSYS.o:(.bss+0x4915c): first defined here
/usr/bin/ld: cdb/SUBSYS.o:(.bss+0x1207f): multiple definition of `Debug_resource_group'; utils/SUBSYS.o:(.bss+0x491dd): first defined here
```

@sekikn The cause seems to be the default value change from `-fcommon` to `-fno-common`. I think just adding `-fcommon` to CFLAGS could be safer than modifying code on our own.

```
$ git diff
diff --git a/bigtop-packages/src/common/gpdb/do-component-configure b/bigtop-packages/src/common/gpdb/do-component-configure
index dc362bb7..93b7b10d 100644
--- a/bigtop-packages/src/common/gpdb/do-component-configure
+++ b/bigtop-packages/src/common/gpdb/do-component-configure
@@ -16,4 +16,4 @@

 set -ex

-./configure --prefix=$1 --disable-orca
+./configure --prefix=$1 --disable-orca CFLAGS=-fcommon
```

sekikn:
@iwasakims Thanks! I confirmed your fix worked on all distros. Updated the PR.

iwasakims (approved):
Thanks. I'm +1.

PR #774, opened 2021-05-17
BIGTOP-3539. Building bigtop-groovy fails on all distros and platforms.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3539

### DIFF
--- a/bigtop.bom
+++ b/bigtop.bom
@@ -254,13 +254,13 @@ bigtop {
     }
     'bigtop-groovy' {
       name    = 'bigtop-groovy'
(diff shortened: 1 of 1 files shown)


iwasakims (approved):
+1. Thanks, @sekikn.

iwasakims:
I had to remove already downloaded tarball to reproduce the issue.  for my records.
```
$ rm dl/bigtop-groovy-2.5.4.tar.gz
$ ./gradlew bigtop-groovy-clean bigtop-grooby-pkg
```

PR #775, opened 2021-05-18
BIGTOP-3541: Smoke test on odpi-runtime failed

yoda-mon:
https://issues.apache.org/jira/browse/BIGTOP-3541

This was probably because https://conjars.org/repo/ stopped providing jars.

To avoid the issue, change the maven repository url from https://conjars.org/repo/ to https://public.nexus.pentaho.org/repository/proxy-public-3rd-party-release

Apache Beam had the same kinds of issue, and they switched to the pentaho's own repository.
https://issues.apache.org/jira/browse/BEAM-11689

### DIFF
--- a/bigtop-tests/smoke-tests/odpi-runtime/build.gradle
+++ b/bigtop-tests/smoke-tests/odpi-runtime/build.gradle
@@ -21,7 +21,7 @@ apply plugin: 'java'
 
 repositories {
   maven {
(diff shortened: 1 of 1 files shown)


sekikn:
+1, I could reproduce the problem on a machine that doesn't have gradle cache and made sure that your PR resolved it.
Thank you for the contribution, @yoda-mon!

PR #777, opened 2021-05-21
BIGTOP-3542. Fix build failure of Hadoop on ppc64le.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3542


sekikn:
+1 tested on CentOS 7/ppc64le. Thanks @iwasakims!

PR #783, opened 2021-06-09
BIGTOP-3548: Hive build failure caused by protoc-jar-maven-plugin mis…

JunHe77:
…sing aarch64 support

This is a follow-up patch of 0cabb0791e430ec0f710371cecd20b10c9ae815c
to solve the build failure on CentOS-7.
The reason is protoc included in com.github.os72::protoc doesn't work
with legacy centos-7 libc. so we use protoc built by bigtop_toolchain
to create a local jar.

Change-Id: I22b6158997adb946237121ed509159e21d7b6b45


bigtop-packages/src/common/hive/do-component-build
```diff
@@ -32,11 +32,21 @@ HIVE_MAVEN_OPTS=" -Dhbase.version=$HBASE_VERSION \
 # Include common Maven Deployment logic
 . $(dirname ${0})/maven_deploy.sh
 
+# workaround for BIGTOP-3548, as the protoc binary included in
+# com.github.os72::protoc doesn't work with legacy centos-7 libc
+# so we use protoc built by bigtop_toolchain to create a fake jar
+source /etc/os-release
+case ${ID}-${VERSION_ID} in
+    centos-7)
+        mvn install:install-file -DgroupId=com.github.os72 -DartifactId=protoc -Dversion=2.6.1-build3 \
```
iwasakims, line 41: protoc built by bigtop_toolchain is 2.5.0. If we can use local protoc artifact, how about set the version back to 2.5.0 for all distros/archs?

JunHe77: That should work. Actually, I'm a bit hesitating on which way we need to use: just create a local 2.5.0 artifact or backport upstream patch (HIVE-21939) while making a hack for CentOS-7 as well. I'm just wondering if there is a possibility that a user directly builds the source and patches in srpm/sdeb without bigtop toolchain. If that is a case it might be better to align with HIVE upstream as much as possible while creating a little hack here in the build script.

bigtop-packages/src/common/hive/do-component-build
```diff
@@ -32,11 +32,21 @@ HIVE_MAVEN_OPTS=" -Dhbase.version=$HBASE_VERSION \
...
+            -Dclassifier=linux-aarch_64 -Dpackaging=exe -Dfile=/usr/local/bin/protoc
```
JunHe77, line 46: "linux-aarch_64" is hardcoded. This should be changed to reflect certain non-x86 platform.

bigtop-packages/src/common/hive/do-component-build
iwasakims: Hadoop already depend on protoc built by bigtop_toolchain (on aarch64 and ppc64le). Phoenix will do so by #786. Asking users for using bigtop_toolchain on aarch64 and ppc64le should be reasonable. Since the fix version of HIVE-21939 is 4.0.0, it is not tested on the field. Removing patch6-HIVE-21939.diff should be safer if possible.

JunHe77: Ya. It makes sense to use the same solution to fix the same protoc problem across hive/hbase/phoenix. Will update patch soon.

iwasakims (approved):
+1 smoke-tests of hive succeeded on Ubuntu 18.04 (on EC2 A1). Thanks, @JunHe77 !

```
$ ./docker-hadoop.sh \
 --create 1 \
 --image bigtop/puppet:trunk-ubuntu-18.04 \
 --memory 16g \
 --disable-gpg-check \
 --stack hdfs,yarn,mapreduce,hive \
 --smoke-tests hive
...
:bigtop-tests:smoke-tests:hive:test (Thread[Execution worker for ':' Thread 4,5,main]) completed. Took 25.977 secs.

BUILD SUCCESSFUL in 1m 32s
35 actionable tasks: 7 executed, 28 up-to-date
Stopped 1 worker daemon(s).
+ rm -rf buildSrc/build/test-results/
```

PR #784, opened 2021-06-09
BIGTOP-3551: Phantomjs caused Ambari build failure on non-x86

JunHe77:
After ambari is bumped to 2.7.5, it cannot build on non-x86
platforms as phantomjs prebuilt is required.
This patch disables phantomjs reference as it is only used
for testing.

Change-Id: Ie23eae41441b201b31d6ed652ac203211b9482a7


martin-g:
With https://github.com/apache/ambari/pull/3312 I've replaced PhantomJS with headless Chromium but it is not merged yet.

iwasakims (approved):
+1. worked for me on Ubuntu 18.04 aarch64.

PR #787, opened 2021-06-09
BIGTOP-3555. Drop smoke-tests/odpi-runtime.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3555

It seem not to be maintained and used in release process of Bigtop now. We should drop unnecessary modules in order to avoid issues like [BIGTOP-3456](https://issues.apache.org/jira/browse/BIGTOP-3456).


sekikn:
+1, confirmed some existing gradlew commands worked with this PR.
Thanks @iwasakims!

PR #786, opened 2021-06-09
BIGTOP-3554: Phoenix build failure caused by protoc-2.5.0

JunHe77:
Change-Id: I7b153ae0c32b2d863488d3075c518e955d7cd9b2
Signed-off-by: Jun He <jun.he@arm.com>

PR #789, opened 2021-06-22
BIGTOP-3558. Fix build failure of HBase on ppc64le.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3558

ppc64le binary of protoc 3.5.1-1 is not available in maven repositories. Creating artifact from locally installed protobuf-3.5.1.1 should be fix. same as done on #785 and #786 .

**I did not test this on my local since I don't have ppc64le environment now.**

The same fix with replacement of `s/ppc64le/aarch64/g` worked on CentOS 7 aarch64.

### DIFF
--- a/bigtop-packages/src/common/hbase/do-component-build
+++ b/bigtop-packages/src/common/hbase/do-component-build
@@ -25,10 +25,13 @@ if [ "${BIGTOP_JDK:=8}" == "8" ]; then
 fi
 
 if [ $HOSTTYPE = "powerpc64le" ] ; then
-  sed  -i "s|<asciidoctor.plugin.version>.*</asciidoctor.plugin.version>|<asciidoctor.plugin.version>1.5.3</asciidoctor.plugin.version>|" pom.xml
+  sed -i "s|<asciidoctor.plugin.version>.*</asciidoctor.plugin.version>|<asciidoctor.plugin.version>1.5.3</asciidoctor.plugin.version>|" pom.xml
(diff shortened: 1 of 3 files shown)


afsanjar:
@iwasakims great job, but please check my comment in https://issues.apache.org/jira/browse/BIGTOP-3558

sekikn:
@iwasakims Would you add the following change to do-component-build? I tested the PR on the ppc64le worker node, but the default version of asciidoctorj seems to cause OOM when generating PDF.

```
-  sed  -i "s|<version>1.5.0-alpha.6</version>|<version>1.5.0-alpha.11</version>|" pom.xml
+  sed -i 's|<asciidoctorj.pdf.version>.*</asciidoctorj.pdf.version>|<asciidoctorj.pdf.version>1.5.0-alpha.11</asciidoctorj.pdf.version>|' pom.xml
```

iwasakims:
@afsanjar Thanks. The patch you mentioned ( https://github.com/ppc64le/build-scripts/blob/master/p/protoc/ubuntu-20.04/protoc-3.5.1.patch ) looks like backport of https://github.com/protocolbuffers/protobuf/commit/36ba04b5792c75d21be2ee91bfd5358980733fd1 .

I'm going to add the patch to bigtop_toolchain.

iwasakims:
@sekikn @afsanjar, I pushed the changes addressing your comments.

afsanjar:
+1

sekikn:
I'm also +1, tested on the ppc64le worker node that `./gradlew toolchain` and `./gradlew hbase-pkg` worked as expected.
Thanks @afsanjar @iwasakims!

PR #790, opened 2021-06-22
BIGTOP-3559. Fix the download URL for pip2 in the GPDB deployment manifest.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3559


sekikn:
Tested on CentOS 7.

iwasakims (approved):
+1. It worked on Ubuntu 18.04 (x86_64) too.

PR #793, opened 2021-06-28
BIGTOP-3564. Fix build failure of Flink on Cent OS 7 ppc64le.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3564

Build fails due to error related to node.js.
```
Running 'npm ci --cache-max=0 --no-save' in /bigtop/build/flink/rpm/BUILD/flink-1.11.3/flink-runtime-web/web-dashboard
/bigtop/build/flink/rpm/BUILD/flink-1.11.3/flink-runtime-web/web-dashboard/node/node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /bigtop/build/flink/rpm/BUILD/flink-1.11.3/flink-runtime-web/web-dashboard/node/node)
```

https://github.com/eirslett/frontend-maven-plugin/issues/852 reported that using v.12 node.js fixed similar issue.

**I did not test this on ppc64le.**

PR #794, opened 2021-06-29
BIGTOP-3565. Fix build failure of Alluxio on Cent OS 7 ppc64le.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3565

```
[INFO] Running 'npm install --production --no-save --no-package-lock --no-shrinkwrap' in /bigtop/build/alluxio/rpm/BUILD/alluxio-2.4.1/webui
[ERROR] /bigtop/build/alluxio/rpm/BUILD/alluxio-2.4.1/webui/node/node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /bigtop/build/alluxio/rpm/BUILD/alluxio-2.4.1/webui/node/node)
```

Upgrading node.js to 12 same as done for https://github.com/apache/bigtop/pull/793 should be the fix. We need to upgrade npm too to the version supporting node 12 here.

### DIFF
--- a/bigtop-packages/src/common/alluxio/do-component-build
+++ b/bigtop-packages/src/common/alluxio/do-component-build
@@ -18,4 +18,9 @@ set -ex
 
 . `dirname $0`/bigtop.bom
 
-mvn clean install -DskipTests -Dhadoop.version=3.2.2 -Phadoop-3 -Pyarn "$@"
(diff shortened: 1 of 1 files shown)


sekikn:
+1, tested on the ppc64le machine. Thanks @iwasakims!

PR #795, opened 2021-07-07
BIGTOP-3567. Building Zeppelin fails due to the dependency downloading problem.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3567


iwasakims (approved):
It looks like specific to Red Hat family on aarch64? I could not reproduce the issue on CentOS 8 x86_64 and Ubuntu 18.04 aarch64.

Anyway I'm +1 since `zeppelin-pkg` worked with the patch on those platforms.

iwasakims:
Oops. I could reproduce the issue by removing cached jars from my ~/.m2.
```
$ rm -rf ~/.m2/repository/org/jetbrains/pty4j
```

sekikn:
I think it's not specific for the RHEL family on aarch64, because I could reproduce it by removing pty4j from .m2/repository on CentOS 7/x86_64.
Thanks for the review anyways!

PR #798, opened 2021-07-10
BIGTOP-3570. Fix Spark smoke test failures.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3570


sekikn:
Tested on CentOS 7.

iwasakims (approved):
+1. It worked on CentOS 8 too. Thanks, @sekikn.

PR #797, opened 2021-07-10
BIGTOP-3569. Applying GPDB Puppet manifest fails on arm64.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3569

PR #800, opened 2021-07-12
BIGTOP-3569. Applying GPDB Puppet manifest fails on arm64 (addendum).

sekikn:
Addendum PR for https://issues.apache.org/jira/browse/BIGTOP-3569. See https://github.com/apache/bigtop/pull/797#issuecomment-877890859 for details.


iwasakims (approved):
+1. We should merge this and see if it's enough on CI environment.

PR #806, opened 2021-08-10
BIGTOP-3575. Fix Hive rpm build warning due to missing dependency.

woblerr:
https://issues.apache.org/jira/browse/BIGTOP-3575


iwasakims (approved):
+1. Thanks for the contribution, @woblerr.

PR #808, opened 2021-08-13
BIGTOP-3579. Spark smoke test fails on Debian 9 and Ubuntu 18.04.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3579

### DIFF
--- a/bigtop-deploy/puppet/modules/spark/manifests/init.pp
+++ b/bigtop-deploy/puppet/modules/spark/manifests/init.pp
@@ -136,6 +136,46 @@ class spark {
   }
 
   class sparkr {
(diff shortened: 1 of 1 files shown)


sekikn:
Tested on Debian 9 and CentOS 7 (as regression).

iwasakims (approved):
+1. I was able to reproduce the issue on Ubuntu 18.04 x86_64 and the patch worked.

iwasakims:
@sekikn It took over 800 seconds to apply the manifest in the docker provisioned container. Do we have enough long timeout on CI environment?

sekikn:
Thanks for the review @iwasakims. Timeout is set to 10h (!) on our CI environment, so it's enough. It may be too long conversely :)

PR #809, opened 2021-08-13
BIGTOP-3576. Deploying Ambari to Fedora 33 fails on ppc64le.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3576


sekikn:
I confirmed that Ambari deployment and smoke test succeeded on the ppc64le server with this PR.

iwasakims (approved):
+1. Thanks, @sekikn.

PR #810, opened 2021-08-13
BIGTOP-3578. Solr smoke test fails on ppc64le.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3578


sekikn:
Tested with CentOS 7 on ppc64le and Debian 9 on x86_64 (as a regression test).

iwasakims (approved):
+1. I got no regression on running smoke-tests of solr on Ubuntu 18.04 x86_64. Thanks, @sekikn.
```
$ ./docker-hadoop.sh \
    --create 1 \
    --image bigtop/puppet:trunk-ubuntu-18.04 \
    --memory 8g \
    --disable-gpg-check \
    --stack zookeeper,solrcloud \
    --smoke-tests solr
```

PR #816, opened 2021-09-03
BIGTOP-3583. Bump Zeppelin to 0.10.0.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3583

Packaging and smoke-tests are worked on CentOS 8 x86_64.


elukey:
Successfully built on Debian 10 x86_64. Worth noting that the build took several hours, a lot of time spent during `dh-strip-nondeterminism`. If it is not my docker config and happens to other people, we may need to add `override_dh_strip_nondeterminism` to the Debian rules file (but I am not super familiar with this step, will need to dig a bit more into it).

elukey:
Tried with:

```
diff --git a/bigtop-packages/src/deb/zeppelin/rules b/bigtop-packages/src/deb/zeppelin/rules
index 6ac4cd8a..a6d61ad0 100644
--- a/bigtop-packages/src/deb/zeppelin/rules
+++ b/bigtop-packages/src/deb/zeppelin/rules
@@ -40,3 +40,5 @@ override_dh_auto_install: $(svcs)
        --doc-dir=/usr/share/doc/zeppelin \
        --source-dir=debian \
        --prefix=debian/tmp
+
+override_dh_strip_nondeterminism:
\ No newline at end of file
diff --git a/bigtop.bom b/bigtop.b
```

And it took 33 minutes to complete..

elukey:
+1, build and smoke tests ran fine on Debian. Not sure if we should follow up on the `dh_strip_nondeterminism` step, it takes hours and we may not need it (but it shouldn't block this change).

iwasakims:
Thanks, @elukey. I filed [BIGTOP-3623](https://issues.apache.org/jira/browse/BIGTOP-3623) for dh_strip_nondeterminism. merging this.

PR #818, opened 2021-09-29
BIGTOP-3585. Remove obsolete text replacement from do-component-build of Ambari.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3585

We do not need the test replacement after https://github.com/apache/ambari/pull/3000.


sekikn:
+1, tested with Ubuntu 20.04 on x86_64.

PR #819, opened 2021-10-10
BIGTOP-3587. Deploying Ambari to Ubuntu 18.04 and 20.04 fails on ppc64le.

sekikn:
https://issues.apache.org/jira/browse/BIGTOP-3587


sekikn:
Tested with Debian 10, Ubuntu 18.04 and 20.04 on ppc64le.

PR #822, opened 2021-10-22
BIGTOP-3590. Hbase smoke tests failed on Ubuntu 20.04.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3590

The cause is non-deterministic order of test execution. The directory named as `/user/$USER/orig_*` is created by TestHBaseImportExport. If TestImportTsv is executed before the TestHBaseImportExport, TestImporptTsv#cleanUp raises exception since there is no such directory. Cleaning up of the `/user/$USER/orig_*` should be done in TestHBaseImportExport.


iwasakims:
I tested this on Ubuntu 20.04  (x86_64) and CentOS 7 (x86_64).

PR #824, opened 2021-11-03
BIGTOP-3591: Upgrading Bigtop-Mpack services deploying scripts aligne…

guyuqi:
After upgrading Ambari to 2.7.5, it failed to install the services from the Bigtop-Mpack.
The patch is is to upgrade Bigtop-Mpack services deploying scripts and fix the related issues.
Successfully install ambari-2.7.5 and deploy related services on Centos-7. 

All modified files are under module:
`bigtop-packages/src/common/bigtop-ambari-mpack/bgtp-ambari-mpack`

PR #825, opened 2021-11-03
BIGTOP-3592: Remove Ambari stack-select tool dependency from stack hooks

guyuqi:
The patch is related to PR #824.
It removed Ambari stack-select tool dependency from stack hooks when deploying mpack services.


guyuqi:
Successfully build Ambari on centos-7 and ubuntu20 and smoke tests also were passed.

vivostar:
I find it cannot build successfully Ambari on centos-7 
image
so I make an another patch to fix it.

PR #827, opened 2021-11-12
BIGTOP-3596. Bump ZooKeeper to 3.5.9.

iwasakims:
This is prerequisite of bumping products using ZooKeeper 3.5 like Hadoop 3.3 and HBase 2.4.

* followed the prefix change of contents of zookeeper-3.5.9-lib.tar.gz.
* addressed the change of output of `zkServer.sh status`.
* We can remove patch0-ZOOKEEPER-3079.diff since ZOOKEEPER-3719 covers the issue.
* We can remove patch1-ZOOKEEPER-3302.diff since 3.5.9 does not have compilation issue reported in ZOOKEEPER-3302.
* We can remove patch2-ZOOKEEPER-2654.diff since ZOOKEEPER-3034 obsoletes the fix.

PR #828, opened 2021-11-15
BIGTOP-3597. Fix build failure of Hadoop 3.2 against ZooKeeper 3.5.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3597

Backporting [YARN-9783](https://issues.apache.org/jira/browse/YARN-9783) fixes the issue.

While this is a follow-up of https://github.com/apache/bigtop/pull/827, merging this first should be harmless.


elukey:
The commit LGTM, and it seems an easy one (if upstream doesn't want a specific test it seems good to remove it!)

I tested a build with debian-10 and I see a zookeeper-related issue though:

```
docker run --rm  -v `pwd`:/ws --workdir /ws bigtop/slaves:trunk-debian-10 bash -c '. /etc/profile.d/bigtop.sh; ./gradlew allclean hadoop-pkg'
[..]
+ continue 2
# Forcing Zookeeper dependency to be on the packaged jar
ln -sf /usr/lib/zookeeper/zookeeper.jar debian/tmp/usr/lib/hadoop/lib/zookeeper*.jar
ln: target 'debian/tmp/usr/lib/hadoop/lib/zookeeper-jute-3.5.9.jar' is not a directory
make[1]: *** [debian/rules:57: override_dh_auto_install] Error 1
make[1]: Leaving directory '/ws/output/hadoop/hadoop-3.2.2'
make: *** [debian/rules:27: binary] Error 2
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2
debuild: fatal error at line 1182:
dpkg-buildpackage -us -uc -ui -b
 failed
```

+1 to merge this change (I double checked that it fixes the problem) if you want to proceed.

PR #836, opened 2021-11-28
BIGTOP-3607. Change the ambari metric repo to fix build error

vivostar:


iwasakims (approved):
LGTM. I was able to reproduce the issue by `./gradlew ambari-pkg` after removing my caches under `~/.m2`. The patch fixed the build failure.

PR #829, opened 2021-11-15
BIGTOP-3600 - Add Debian 11 support to the Docker provisioner

elukey:
Built locally:
```
bigtop/slaves   trunk-debian-11     16fa6cdec6a7   7 minutes ago       3.89GB
bigtop/puppet  trunk-debian-11     f2fb226b4e3c   2 hours ago         330MB
```

PR #837, opened 2021-12-04
BIGTOP-3600 - Deploy python*-dev deps on Debian images

elukey:
In https://github.com/apache/bigtop/pull/829 I caused some
failures in gpdb package builds, since python2.7-dev
dependencies are needed.

On Debian 9/10/11 python3-dev and python2.7-dev are available,
so better to be explicit and install both at the same time.

### DIFF
--- a/bigtop_toolchain/manifests/packages.pp
+++ b/bigtop_toolchain/manifests/packages.pp
@@ -151,7 +151,7 @@ class bigtop_toolchain::packages {
       "libffi-devel"
     ] }
     /(Ubuntu|Debian)/: {
-      $_pkgs = [
+      $pkgs = [
         "unzip",
         "curl",
         "wget",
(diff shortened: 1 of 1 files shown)


elukey:
Current image:

```
docker run --rm  -v `pwd`:/ws --workdir /ws bigtop/slaves:trunk-debian-10 bash -c 'dpkg -S Python.h'
libpython3.7-dev:amd64: /usr/include/python3.7m/Python.h
```

New image:

```
docker run --rm  -v `pwd`:/ws --workdir /ws bigtop/slaves:trunk-debian-10 bash -c 'dpkg -S Python.h'
libpython3.7-dev:amd64: /usr/include/python3.7m/Python.h
libpython2.7-dev:amd64: /usr/include/python2.7/Python.h
```

elukey:
With the new image I am able to build `gpdb` on Debian 10, should be the same on 11. We probably want to deprecated any usage of Python 2 in the future, maybe after 3.1?

iwasakims (approved):
+1. I was able to reproduce the issue on (non-container) Debian 11. The fix worked.

PR #838, opened 2021-12-05
BIGTOP-3309 - Update ca-certificates for CentOS 7

elukey:
Due to https://access.redhat.com/articles/6338021 ant and  maven classes
fail to download due to TLS certificate expiration check issues.
Since the issue is related to CentOS 7 and earlier, it seems sensible
to force the upgrade of the ca-certificates package (that in its latest
versions contain a fix).

While at it:
1) Add a specific dependency to 'ca-certificates' for CentOS 7.
2) Require bigtop_toolchain::packages in the varios ant/maven/groovy
   classes, so that we'll be able to rely on packages being
   installed when running the Execs. It may not be strictly required
   now but it seems good for future code.
3) Refactor a bit the Execs to be more readable when puppet runs
   (easier to track what's happening).

Tested rebuilding the puppet image and the slave image.

### DIFF
--- a/bigtop_toolchain/manifests/ant.pp
+++ b/bigtop_toolchain/manifests/ant.pp
@@ -16,26 +16,31 @@
 class bigtop_toolchain::ant {
 
   require bigtop_toolchain::gnupg
+  require bigtop_toolchain::packages
 
   # Ant version restricted to 1.9 because 1.10 supports Java>=8 only.
   $ant =  latest_ant_binary("1.9.[0-9]*")
   $apache_prefix = nearest_apache_mirror()
 
-  exec {"/usr/bin/wget $apache_prefix/ant/binaries/$ant-bin.tar.gz":
+  exec { 'Download Ant binaries':
+    command => "/usr/bin/wget $apache_prefix/ant/binaries/$ant-bin.tar.gz",
     cwd     => "/usr/src",
     unless  => "/usr/bin/test -f /usr/src/$ant-bin.tar.gz",
   } ~>
 
-  exec {"/usr/bin/wget https://www.apache.org/dist/ant/binaries/$ant-bin.tar.gz.asc":
+  exec { 'Download Ant binaries signature':
+    command => "/usr/bin/wget https://www.apache.org/dist/ant/binaries/$ant-bin.tar.gz.asc",
     cwd     => "/usr/src",
     unless  => "/usr/bin/test -f /usr/src/$ant-bin.tar.gz.asc",
   } ~>
 
-  exec {"/usr/bin/$bigtop_toolchain::gnupg::cmd -v --verify --auto-key-retrieve --keyserver hkp://keyserver.ubuntu.com:80 $ant-bin.tar.gz.asc":
+  exec { 'Verify Ant binaries':
+    command => "/usr/bin/$bigtop_toolchain::gnupg::cmd -v --verify --auto-key-retrieve --keyserver hkp://keyserver.ubuntu.com:80 $ant-bin.tar.gz.asc",
     cwd     => "/usr/src",
   } ->
 
-  exec {"/bin/tar xvzf /usr/src/$ant-bin.tar.gz":
+  exec { 'Extract Ant binaries':
+    command => "/bin/tar xvzf /usr/src/$ant-bin.tar.gz",
     cwd     => '/usr/local',
     creates => "/usr/local/$ant",
   } ->
--- a/bigtop_toolchain/manifests/groovy.pp
+++ b/bigtop_toolchain/manifests/groovy.pp
@@ -15,16 +15,20 @@
 
 class bigtop_toolchain::groovy {
 
+  require bigtop_toolchain::packages
+
   $groovy_version = '2.5.4'
   $groovy = "apache-groovy-binary-${groovy_version}"
 
-  exec {"/usr/bin/wget https://dl.bintray.com/groovy/maven/${groovy}.zip":
+  exec { 'Download Groovy':
+    command => "/usr/bin/wget https://dl.bintray.com/groovy/maven/${groovy}.zip",
     cwd     => "/usr/src",
     unless  => "/usr/bin/test -f /usr/src/${groovy}.zip",
   } ~>
 
-  exec {"/usr/bin/unzip -x -o /usr/src/${groovy}.zip":
-    cwd         => '/usr/local',
+  exec {'Extract Groovy':
+    command => /usr/bin/unzip -x -o /usr/src/${groovy}.zip":
+    cwd     => '/usr/local',
   } ->
 
   file {'/usr/local/groovy':
--- a/bigtop_toolchain/manifests/maven.pp
+++ b/bigtop_toolchain/manifests/maven.pp
@@ -16,31 +16,36 @@
 class bigtop_toolchain::maven {
 
   require bigtop_toolchain::gnupg
+  require bigtop_toolchain::packages
 
   $mvnversion = latest_maven_binary("3.6.[0-9]*")
   $mvn = "apache-maven-$mvnversion"
 
   $apache_prefix = nearest_apache_mirror()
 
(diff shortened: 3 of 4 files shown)


iwasakims:
@elukey
The fix looks intended for running puppetize.sh again to refresh existing environment. If it is the first time, the latest ca-certificates is installed regardless of your fix since it is required by (rubygems required by) puppet.

```
$ docker run bigtop/puppet:trunk-centos-7 rpm -q --whatrequires ca-certificates
openssl-libs-1.0.2k-19.el7.x86_64
rubygems-2.0.14.1-36.el7.noarch
```

I think `./gradlew toolchain` tends to be used to refresh already existing dev env rather than puppetize.sh. Should we add `yum update ca-certificates` in the manifests under [bigtop_toolchain/manifests](https://github.com/apache/bigtop/tree/master/bigtop_toolchain/manifests) too?

elukey:
@iwasakims thanks for the review! What I tried before this solution was:

- Add `require bigtop_toolchain::packages` in the maven/ant classes, with the aim to install ca-certificates (via dependency) before the Execs run, but it didn't work. The install packages step started before ant/maven but the Execs ran while it was still in progress, ending up in TLS validation failures.
- Add `ca-certificates` to the `bigtop_toolchain::packages` class, and add a `require => Package["ca-certificates"]` to the Execs in the ant/maven classes. This seems to work, but it affects all slave image builds. It seems correct in my opinion to have a require for wget and ca-certificates in those classes, since the Execs implicitly rely on those, but then I preferred to limit the scope to centos-7 and use puppetize.sh (that runs before puppet manifests are applied for sure, and also that installs wget and other utils as well).

Whatever we choose should guarantee, in my opinion, that the puppet Execs in the maven/ant/etc.. classes run when the ca-certificates packages is installed, otherwise dowloading maven/ant fails (and sometimes the image build proceeds anyway ending up in package build failures later on).

Let me know!

elukey:
I see what's happening, I was wrong, the ca-certificates package is installed but not at the right version. Testing an alternative approach in `bigtop_toolchain::packages` :)

elukey:
@iwasakims pushed a new version, lemme know what you think about it :)

iwasakims:
`ensure => latest` instead of `ensure => installed` should upgrade already installed packages. How about adding following?

```
  if $operatingsystem == 'CentOS' {
    ...(snip)

    if $operatingsystemmajrelease == 7 {
      package { ca-certificates:
        ensure => latest
      }
    }
```

While adding global constraints like `Package<||> -> Exec<||>` is handy, it causes following error.

```
Error: Could not apply complete catalog: Found 1 dependency cycle:
(Exec[Setuptools Installation] => Class[Bigtop_toolchain::Packages] => Class[Bigtop_toolchain::Renv] => Package[R] => Exec[Setuptools Installation])
Try the '--graph' option and opening the resulting '.dot' file in OmniGraffle or GraphViz
```

It would be nice if we can get rid of `require => Package[ca-certificates]` on every wget resources...

elukey:
Ah snap I tried to build the trunk-centos-7 puppet/slave image and it worked, where did you see that error? Anyway, will work on another solution, thanks for the feedback!

iwasakims:
> where did you see that error?

I think you can reproduce the issue by following diff.

```
$ git diff
diff --git a/bigtop_toolchain/manifests/packages.pp b/bigtop_toolchain/manifests/packages.pp
index 93e6c0a..fe0a3ec 100644
--- a/bigtop_toolchain/manifests/packages.pp
+++ b/bigtop_toolchain/manifests/packages.pp
@@ -250,8 +250,14 @@ class bigtop_toolchain::packages {
         ensure => latest
       }
     }
-  }

+    if $operatingsystemmajrelease == 7 {
+      package { ca-certificates:
+        ensure => latest
+      }
+      Package<||> -> Exec<||>
+    }
+  }

   # BIGTOP-3364: Failed to install setuptools by pip/pip2
   # on Ubuntu-16.04/18.04 and centos-7.
```
```
$ ./gradlew toolchain
...
Debug: /Stage[main]/Bigtop_toolchain::User/File[/var/lib/jenkins/.ssh]: Autorequiring Group[jenkins]
Info: Applying configuration version '1638800828'
Error: Could not apply complete catalog: Found 1 dependency cycle:
(Exec[Setuptools Installation] => Class[Bigtop_toolchain::Packages] => Class[Bigtop_toolchain::Renv] => Package[R] => Exec[Setuptools Installation])
Try the '--graph' option and opening the resulting '.dot' file in OmniGraffle or GraphViz
...
```

iwasakims:
Ordering between classes could be used instead of the `Package<||> -> Exec<||>` like

```
$ git diff
diff --git a/bigtop_toolchain/manifests/installer.pp b/bigtop_toolchain/manifests/installer.pp
index bcf9ee7..4ce5f67 100644
--- a/bigtop_toolchain/manifests/installer.pp
+++ b/bigtop_toolchain/manifests/installer.pp
@@ -25,6 +25,10 @@ class bigtop_toolchain::installer {
   include bigtop_toolchain::renv
   include bigtop_toolchain::grpc

+  Class['bigtop_toolchain::packages'] -> Class['bigtop_toolchain::maven']
+  Class['bigtop_toolchain::packages'] -> Class['bigtop_toolchain::ant']
+  Class['bigtop_toolchain::packages'] -> Class['bigtop_toolchain::gradle']
+
   stage { 'last':
     require => Stage['main'],
   }
diff --git a/bigtop_toolchain/manifests/packages.pp b/bigtop_toolchain/manifests/packages.pp
index 93e6c0a..b8e3be8 100644
--- a/bigtop_toolchain/manifests/packages.pp
+++ b/bigtop_toolchain/manifests/packages.pp
@@ -250,8 +250,13 @@ class bigtop_toolchain::packages {
         ensure => latest
       }
     }
-  }

+    if $operatingsystemmajrelease == 7 {
+      package { ca-certificates:
+        ensure => latest
+      }
+    }
+  }

   # BIGTOP-3364: Failed to install setuptools by pip/pip2
   # on Ubuntu-16.04/18.04 and centos-7.
```

iwasakims:
Hmm.. `wget` failed even with the latest ca-certificates on my CentOS 7 environment.

```
$ rpm -q ca-certificates
ca-certificates-2021.2.50-72.el7_9.noarch

$ /usr/bin/wget https://dlcdn.apache.org//maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz
--2021-12-06 23:38:30--  https://dlcdn.apache.org//maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz
Resolving dlcdn.apache.org (dlcdn.apache.org)... 151.101.2.132, 2a04:4e42::644
Connecting to dlcdn.apache.org (dlcdn.apache.org)|151.101.2.132|:443... connected.
ERROR: cannot verify dlcdn.apache.org's certificate, issued by ‘/C=US/O=Let's Encrypt/CN=R3’:
  Issued certificate has expired.
To connect to dlcdn.apache.org insecurely, use `--no-check-certificate'.
```

elukey:
```

+  Class['bigtop_toolchain::packages'] -> Class['bigtop_toolchain::maven']
+  Class['bigtop_toolchain::packages'] -> Class['bigtop_toolchain::ant']
+  Class['bigtop_toolchain::packages'] -> Class['bigtop_toolchain::gradle']
+
```
This bit in theory is the same as adding requires in specific classes right? Not sure if there is a preference in Bigtop, I am more used to the explicit require where needed vs the include + -> ordering.

iwasakims:
> This bit in theory is the same as adding requires in specific classes right?

Now I think so. I gueesed that `require bigtop_toolchain::packages` did not work but the `wget` failure seems to have different cause.

iwasakims:
Replacing files by `sudo yum reinstall ca-certificates` fixed the issue on non-docker environment. Something was broken on my side.

elukey:
@iwasakims tried to add your suggestions, let me know how it looks now :)

Thanks for the patience!

(tested for puppet/slave images, no issues)

iwasakims (approved):
+1. The latest patch worked for both docker and non-docker CentOS 7. Thanks, @elukey.

PR #840, opened 2021-12-12
BIGTOP-3604 - Upgrade jsvc to 1.2.4

elukey:
Users are starting to test Bigtop packages with Openjdk 11,
and it was reported that the HDFS Datanote is not able to run
correctly without a more up-to-date version of bigtop-jsvc.


iwasakims (approved):
LGTM.

iwasakims:
We are keeping bigotp-jsvc  just for compatibility. Hadoop >= 2.6.0 leverages SASL for DataNode security. JSVC is only used for old configuration. We should deprecate bigtop-jsvc. Products other than Hadoop seems not to use bigtop-jsvc.

https://hadoop.apache.org/docs/r3.2.2/hadoop-project-dist/hadoop-common/SecureMode.html#Secure_DataNode

PR #845, opened 2021-12-21
BIGTOP-3622: replace $(PWD) with $(CURDIR) in Debian rule files

elukey:
The only remaining one is the Oozie's rules file.

### DIFF
--- a/bigtop-packages/src/deb/oozie/rules
+++ b/bigtop-packages/src/deb/oozie/rules
@@ -36,7 +36,7 @@ override_dh_auto_build:
 	tar cf - --exclude=debian/\* . | (cd debian/tmp && tar xf -)
 
 override_dh_auto_install:
-	sh -x debian/install_oozie.sh --extra-dir=debian/ --build-dir=$(PWD) --server-dir=./debian/oozie --client-dir=./debian/oozie-client --docs-dir=./debian/oozie-client/usr/share/doc/oozie --initd-dir=./debian/oozie/etc/init.d --conf-dir=./debian/oozie/etc/oozie/conf.dist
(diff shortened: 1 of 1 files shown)


oflebbe (approved):

elukey:
@oflebbe thanks for the review! This would be my first commit to merge, and I have to admit that I have never used github/gitbox so I don't know the workflow. I checked https://cwiki.apache.org/confluence/display/BIGTOP/How+to+Contribute but it doesn't seem up to date afaics, what is the best procedure?
- Can we merge from the Github mirror UI? (If so I don't find a way to do it)
- Should we merge from Gitbox's UI?
- Alternatively, merge + push from the gitbox local repo, but I suspect that there is a simpler way.
In this case I'd like to merge this change on the master branch, and if it applies cleanly, also on branch-1.5 (people will need to rebuild oozie for https://issues.apache.org/jira/browse/BIGTOP-3621).

Thanks in advance!

oflebbe:
Sorry I was in "At Work Mode" when I did the Github review, I was supposed to use JIRA for Reviews.
However AFAIK the gitbox / github integration should work the same as committing the usual way directly on the ASF infra, but I have to admit I used the "usual way" (as in the How to Contribute) before.  If unsure simply ignore my review here and use git as you would normally do.

oflebbe:
Ah did I get the question wrong -- as in you don't see the "Rebase and merge" button on the PR ?

elukey:
@oflebbe exactly I don't see it, so I was wondering if I had to use gitbox etc..

oflebbe:
I see it , and I was looking for some UI at github to enable/disable . 
Maybe you are missing the steps mentioned in https://blogs.apache.org/infra/entry/apache-and-github-a-friendly ?

elukey:
@oflebbe I am currently trying to sort out my account, if you have time would you mind to merge the change on master and (if possible) also on branch-1.5?

oflebbe:
Patch applies not cleanly to branch-1.5 , sorry.

elukey:
Thanks a lot @oflebbe! I updated https://github.com/apache/bigtop/pull/846 to include this bit, it should be good (so when the other one will be merged we'll include the CURDIR fix as well :)

PR #849, opened 2021-12-31
BIGTOP-3626: Upgrade ycsb and its log4j dependencies

elukey:
At the time of writing upstream didn't release any official fix
for the lo4j CVEs, but brianfrankcooper/YCSB#1583 seems taking care of it.
Credits for the upstream fix: Filipe Oliveira <filipecosta.90@gmail.com>

This change also bumps ycsb to its latest upstream, to allow the log4j
patch to be applied cleanly.


iwasakims (approved):
LGTM. While we should care about compatibility on bumping product versions in PATCH release, YCSB has no dependant. smoke-tests of YCSB passed on my local.

PR #850, opened 2022-01-03
BIGTOP-3627. Pinning the version of puppetlabs-stdlib

vivostar:
When I deploy components according to the Readme doc, I find an error occur
image
Then, I find such a issue to discribe it
https://issues.apache.org/jira/browse/BIGTOP-3088
The issue has been fixed, but the doc haven't updated, so I make this commit to fix the gap.

### DIFF
--- a/bigtop_toolchain/manifests/puppet_modules.pp
+++ b/bigtop_toolchain/manifests/puppet_modules.pp
@@ -17,7 +17,7 @@ class bigtop_toolchain::puppet_modules {
 
   exec { 'install-puppet-stdlib':
     path    => '/usr/bin:/bin',
-    command => 'puppet module install puppetlabs-stdlib',
+    command => 'puppet module install puppetlabs-stdlib --version 4.12.0',
     creates => '/etc/puppet/modules/stdlib',
   }
 


iwasakims:
@vivostar 
The cause seems to be pinning the version of puppetlabs-stdlib is insufficient in [bigtop_toolchain::puppet_modules](https://github.com/apache/bigtop/blob/master/bigtop_toolchain/manifests/puppet_modules.pp). You should fix the manifest rather than README.md.

Since the puppetize.sh belonging to bigtop_toolchain covers installation of Puppet itself, the README of puppet-deploy does not assume to use it. [bigtop_toolchain::puppet_modules](https://github.com/apache/bigtop/blob/master/bigtop_toolchain/manifests/puppet_modules.pp) is the class for installing only puppetlabs-stdlib to the environment in which puppet is already installed.

iwasakims (approved):
Thanks for the update @vivostar. +1.

iwasakims:
for my record:

We are pinning puppetlabs-stdlib to 4.12.0 only for CentOS 7, CentOS 8 and Fedora 33 in [puppetize.sh](https://github.com/apache/bigtop/blob/master/bigtop_toolchain/bin/puppetize.sh).

I got following error on Debian 11 on which puppet-module-puppetlabs-stdlib is already installed. If puppet-module-puppetlabs-stdlib is not installed, it works.

```
$ puppet apply --modulepath=~/srcs/bigtop -e "include bigtop_toolchain::puppet_modules" --debug --verbose
...
Debug: Executing: 'puppet module install puppetlabs-stdlib --version 4.12.0'
Notice: /Stage[main]/Bigtop_toolchain::Puppet_modules/Exec[install-puppet-stdlib]/returns: Notice: Preparing to install into /home/admin/.puppet/etc/code/modules ...
Notice: /Stage[main]/Bigtop_toolchain::Puppet_modules/Exec[install-puppet-stdlib]/returns: Error: Could not install module 'puppetlabs-stdlib' (v4.12.0)
Notice: /Stage[main]/Bigtop_toolchain::Puppet_modules/Exec[install-puppet-stdlib]/returns:   Module 'puppetlabs-stdlib' (v5.0.0) is already installed
Notice: /Stage[main]/Bigtop_toolchain::Puppet_modules/Exec[install-puppet-stdlib]/returns:     Use `puppet module upgrade` to install a different version
Notice: /Stage[main]/Bigtop_toolchain::Puppet_modules/Exec[install-puppet-stdlib]/returns:     Use `puppet module install --force` to re-install only this module
Error: 'puppet module install puppetlabs-stdlib --version 4.12.0' returned 1 instead of one of [0]
Error: /Stage[main]/Bigtop_toolchain::Puppet_modules/Exec[install-puppet-stdlib]/returns: change from 'notrun' to ['0'] failed: 'puppet module install puppetlabs-stdlib --version 4.12.0' returned 1 instead of one of [0]
```

If puppetlabs-stdlib 4.12.0 cause isuses for recent Debian/Ubuntu, I will file a follow-up JIRA.

PR #851, opened 2022-01-03
BIGTOP-3614: allow to change the docker-compose cfg in the provisioner

elukey:
This is a simple change to allow docker-compose to use a different
config file when creating instances. This is useful when testing
new settings and also to keep multiple versions of docker-compose.yml
(if the same version cannot be re-used across multiple OS etc..).

For example, from my recent tests the actual docker-compose.yml
config seems not to work with cgroup v2 and systemd on Debian 10/11.

### DIFF
--- a/provisioner/docker/docker-hadoop.sh
+++ b/provisioner/docker/docker-hadoop.sh
@@ -16,8 +16,9 @@
 # limitations under the License.
 
 usage() {
-    echo "usage: $PROG [-C file] args"
+    echo "usage: $PROG [-C file] [-F file] args"
     echo "       -C file                                   - Use alternate file for config.yaml"
+    echo "       -F file, --docker-compose-yml file        - Use alternate file for docker-compose.yml"
     echo "  commands:"
     echo "       -c NUM_INSTANCES, --create NUM_INSTANCES  - Create a Docker based Bigtop Hadoop cluster"
     echo "       -d, --destroy                             - Destroy the cluster"
@@ -76,7 +77,7 @@ create() {
     export MEM_LIMIT=${memory_limit}
 
     # Startup instances
-    docker-compose -p $PROVISION_ID up -d --scale bigtop=$1 --no-recreate
+    $DOCKER_COMPOSE_CMD -p $PROVISION_ID up -d --scale bigtop=$1 --no-recreate
     if [ $? -ne 0 ]; then
         log "Docker container(s) startup failed!";
         exit 1;
@@ -84,7 +85,7 @@ create() {
 
     # Get the headnode FQDN
     # shellcheck disable=SC2207
-    NODES=(`docker-compose -p $PROVISION_ID ps -q`)
+    NODES=(`$DOCKER_COMPOSE_CMD -p $PROVISION_ID ps -q`)
     # shellcheck disable=SC1083
     hadoop_head_node=`docker inspect --format {{.Config.Hostname}}.{{.Config.Domainname}} ${NODES[0]}`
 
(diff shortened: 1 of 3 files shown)


iwasakims:
@elukey All invocation of `docker-compose` command should be replaced with `docker-compose -f $docker_compose_yamlconf`? I got following error since `docker-compose` without `-f` seeks the file named docker-compose.yml or docker-compose.yaml.

```
$ mv docker-compose.yml docker-compose-renamed.yml
$ ./docker-hadoop.sh \
   --docker-compose-yml docker-compose-renamed.yml \
   --create 1 \
   --memory 8g \
   --disable-gpg-check \
   --stack zookeeper
Environment check...
Check docker:
Docker version 20.10.7, build f0df350
Check docker-compose:
docker-compose version 1.26.2, build eefe0d31
Check ruby:
ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-linux]
Creating 20220104_011930_r8500_bigtop_1 ... done
ERROR:
        Can't find a suitable configuration file in this directory or any
        parent. Are you in the right directory?

        Supported filenames: docker-compose.yml, docker-compose.yaml
...
```

elukey:
@iwasakims great point, thanks a lot! Uploaded a new version of the script, currently under testing :)

iwasakims (approved):
Thanks for the update @elukey. +1.

elukey:
@iwasakims I added a few changes from your last review:
1) A new file called `docker-compose-cgroupsv2.yml`
2) An entry in the README file
3) The possibility to use `-F` in the command line (as opposed to only `--docker-compose-yml` that is really long).

Let me know if it is still ok :)

provisioner/docker/docker-hadoop.sh
```diff
@@ -284,12 +293,19 @@ log() {
...
         docker exec $node bash -c "cd /bigtop-home; ./gradlew -PNEXUS_URL=$1 configure-nexus"
     done
     wait
 }
 
+get_nodes() {
+    if [ -n "$PROVISION_ID" ]; then
+        # shellcheck disable=SC2207
+        NODES=(`$DOCKER_COMPOSE_CMD -p $PROVISION_ID ps -q`)
+    fi
```
evans-ye, line 307: Shall we error out earlier and explicitly log what's the problem when getting nodes?

elukey: I am open to any suggestion, I have basically wrapped what was there in a function.. What do you have in mind? I can incorporate it in the change :)

iwasakims:
@elukey I'm +1 on updated patch. It worked on Debian 11 and Fedora 33.

```
./docker-hadoop.sh \
 --docker-compose-yml docker-compose-cgroupv2.yml \
 --create 1 \
 --image bigtop/puppet:trunk-debian-11 \
 --memory 8g \
 --disable-gpg-check \
 --stack zookeeper
```
```
./docker-hadoop.sh \
 -F docker-compose-cgroupv2.yml \
 --create 1 \
 --image bigtop/puppet:trunk-fedora-33 \
 --memory 8g \
 --disable-gpg-check \
 --stack zookeeper
```

iwasakims:
I cherry-picked this to branch-3.0 since this is useful for testing maintenance release (3.0.1).

provisioner/docker/docker-hadoop.sh
evans-ye: My bad for not being clear. I was thinking that for the else case (PROVISION_ID is not set), we can error out earlier with a clear log message instead of being silence . But this is a minor comment as we didn't do this kind of early error out in other place, plus you're just wrapping what originally written ;)

elukey: @evans-ye ah yes definitely! I can follow up with another commit :)

PR #853, opened 2022-01-08
BIGTOP-3629: Drop Debian 9 support

elukey:


iwasakims (approved):
+1. I ran `./gradlew toolchain` on Debian 11 to test the changes of jdk.pp.

PR #865, opened 2022-02-20
 BIGTOP-3644: Improve Gradle download resiliency

elukey:
This change bumps up the version of the de.undercouch.download
Gradle download plugin to add the `retries` parameter to the
download function used to retrieve upstream artifacts.

The bump from 3.x to 4.x should be supported, afaics, from our
current setup:
https://github.com/michel-kraemer/gradle-download-task#migrating-from-version-3x-to-4x

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ buildscript {
 }
 
 plugins {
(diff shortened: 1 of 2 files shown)


oflebbe:
LGTM +1

PR #869, opened 2022-02-27
BIGTOP-3648. Change download site to dlcdn.apache.org for installing Ant and Maven.

sekikn:
### Description of PR

During installing toolchain, Ant and Maven is downloaded from downloads.apache.org.
But requesting to that site too many times causes IP ban, so this PR replaces it with dlcdn.apache.org,
which isn't supposed to have such a limitation.

I confirmed that `./gradlew toolchain` succeeded with this PR on the node in question, which was still banned from the ASF site.
Also, building and testing Ambari with this PR succeeded as before on another node, which was not blacklisted.

### For code changes:


elukey:
Checked all the new links and binaries available, I think it is a great change, +1 LGTM

sekikn:
Thanks! Merged into master and branch-3.0.

PR #870, opened 2022-03-03
BIGTOP-3649: Fix Nexus local mirror functionality

cheelio:
This PR fixes local Nexus mirror functionality by:
* Adding new maven mirrors to repos list in build.gradle.
* Fixing <mirrors> configuration in settings.xml

I use a comma separated list of repo ids  this XML fragment instead of a just a `*` to prevent the build from breaking when a new repository is added somewhere in the build without adding it to the repos list in build.gradle


cheelio:
On my local machine, build times are dramatically reduced.

sekikn:
@cheelio Great improvement, thank you for the contribution!
Just a quick question, is there any handy way to check if IDs defined in the repos variable covers all repositories used by the Bigtop components?
I'm worried that if some components change their repository definition with their upgrade, or are added/removed with their repo definition, we are likely to forget to update the value of the repos variable.

cheelio:
Good question. First of all: If the repository variable isn't updated, it won't result in a broken build, but just in a less 'optimal' one. I tested that (locally) by removing all the Horton works repos from the variable.
Maven will then fallback to the remote repo.

My best suggestion for keeping this list updated will be grepping the build log for any non-local downloads afterwards. Maybe some periodic check in Jenkins?

cheelio:
Epic oneliner (to run in separate jenkins alert job or something):
```
[[ $(grep "Downloading from" ~/bigtop/build.txt | grep -v "Downloading from default" | wc -l ) = "0" ]] && echo "Build contains external repositories that are not added to 'repos'-variable in buid.gradle. Please consider updating the variable."
```

elukey:
@cheelio I tested your change with a debian-11 Docker image and it works great, I can see the nexus container showing up logs of requests and its cache storage growing as the builds proceed. I tried with a build of hive (note: I overrode the dh_strip_determinism step in the rules files since it is a step that takes ages) and it took ~12 mins on my laptop with an empty cache, ~6 mins with cache warmed up (basically an identical hive-clean hive-pkg executed afterwards). So this is really great!

I am trying to understand the nexus caching code in more details, so apologies if this question is trivial - is there a reason to have `configure-nexus-public-group` and `generate_nexus_tasks` initializing the same set of `repos` ?

cheelio:
The `generate_nexus_tasks` function creates `configure-nexus-xxxxxx` gradle tasks for each `repo` in `repos`. These tasks create a proxy repository in Nexus.

The (new) `configure-nexus-public-group` gradle task is executed after the individual nexus tasks and create/update the `public` repository group in nexus so that this group consists of all the individual proxy's in `repos`. 

Finally, the `configure-nexus` task is the entry point. Because of the dependants, it will always first create/update the proxy repo followed by updating the group repo.

The task itself will write a settings.xml file pointing to the `public` group repo.  The `mirrorOf`-attribute contains a  comma separated string with repo IDs. These will match with the repo IDs in the individual `pom.xml` files.

 In this way, when a new repository is added somewhere in a in a (new) module that isn't (yet) included in the `repos` variable, it will not be proxied by nexus to make sure the build will not fail.

elukey:
@cheelio thanks a lot for the detailed explanation, would it be ok for you to add comments to the code that you modified so everybody will be on the same page when reading `build.gradle`? As far as I can see there is not a lot of knowledge about how Nexus works, and your experience would be really appreciated by future readers if written down properly :)

cheelio:
Something like that?

elukey:
Yep LGTM! @oflebbe @sekikn I think that this change would be great even for the upcoming 3.0.1 release, lemme know what are your thoughts!

oflebbe:
Hi @elukey  , please do what makes sense. The usage of the private repository  nexus-private raises airbrows  for me.
shouldn't we use http://repo.hortonworks.com/content/groups/public instead? .

sekikn:
The nexus-private repository seems to [come from Ambari](https://github.com/apache/ambari/blob/release-2.7.5/pom.xml#L135), and I'm not sure if Ambari is successfully built if we replace it with https://repo.hortonworks.com/content/groups/public/. I'll try it later.

cheelio:
The nexus-private repo was actually introduced in bigtop in  https://github.com/apache/bigtop/commit/ebbf2f763149da2132a055c11d00d41a4ca01eff.

I don't understand why, because in upstream ambari, it was changed to `https://repo.hortonworks.com/content/groups/public/` in https://github.com/apache/ambari/pull/3063/files/20caaf1404b8ed438c9a743e6dec03a36f2cb63f.

sekikn:
I tried to build Ambari 2.7.5 with a clean Maven local repository without private-nexus in any pom.xml and it succeeded.
As Olaf said, I also think we generally shouldn't use corporate private repository if it's not necessary, so would you update the PR with the following items @cheelio?

* Add a diff that replaces private-nexus with https://repo.hortonworks.com/content/groups/public/ into bigtop-packages/src/common/ambari/patch9-remove-not-working-plugin-repos.diff
* Remove the existing bigtop-packages/src/common/ambari/patch9-ambari-repo-change.diff introduced in BIGTOP-3607 (I think it may have been a temporary problem caused by a intermittent failure of the Hortonworks public repository, and its fix is not necessary now)
* Remove the private-nexus entry from the repos variable in build.gradle

Once updated, I'll merge this into branch-3.0 and master.

vivostar:
BIGTOP-3607 is a temporary solution for Ambari dependencies download error. I think we should remove the patch file.

cheelio:
@sekikn @oflebbe @elukey   
Ok, I got two flavours:

First commit has everything you asked for.
Second commit removes all 'dependencies' to Hortonworks repositories. I think I sleep much better at night, knowing that this build will always succeed, even if Cloudera decides to shutdown all their public Nexus repositories. I guess that could be a serious scenario.
What I had to do to make that work is removing the `ambari-metrics-storm-sink-legacy` module from the project. This module contains references 'backtype.storm.metric.IClusterReporter' which only seems available in `org.apache.storm:storm-core:0.10.0.2.3.0.0-2557`. Sources of this interface aren't available anywhere online. By removing `ambari-metrics-storm-sink-legacy` (I think the word 'legacy' tells me that it won't hurt that much), as a module from the build, I can eliminiate *all* references to Hortonworks Nexus repositories from ambari.

Let me know what you think!

Please tell me what you think.

cheelio:
Oh, and another thing that I had to do was changing the pom version of af bunch of modules from from 1.0.0.0-SNAPSHOT to 2.7.5.0.

sekikn:
@cheelio Thanks for updating! I agree to remove the ambari-metrics-storm-sink-legacy module and dependencies on the Hortonworks repository, because that module seems to support before Storm 1.0, which was released more than five years ago. In addition, Bigtop doesn't provide Storm as its component. Give me a few days to try to build and run Ambari with your PR.

bigtop-packages/src/common/ambari/patch9-remove-not-working-plugin-repos.diff
```diff
@@ -0,0 +1,70 @@
...
+   </dependencies>
+
+-  <pluginRepositories>
+-    <pluginRepository>
+-      <id>oss.sonatype.org</id>
+-      <name>OSS Sonatype Staging</name>
+-      <url>https://oss.sonatype.org/content/groups/staging</url>
+-    </pluginRepository>
+-  </pluginRepositories>
+-
```
sekikn, line 33: After applying this patch, references to oss.sonatype.org still seem to remain in ambari-infra/pom.xml and ambari-logsearch/pom.xml. Should we remove them (and the entry from the repos variable in build.gradle) too?

bigtop-packages/src/common/ambari/patch9-z_bonus.diff
```diff
@@ -0,0 +1,446 @@
...
+@@ -1789,7 +1789,7 @@
+     <dependency>
+       <groupId>org.apache.ambari</groupId>
+       <artifactId>ambari-utility</artifactId>
+-      <version>1.0.0.0-SNAPSHOT</version>
++      <version>2.7.5.0.0</version>
+       <scope>provided</scope> <!-- for @ApiIgnore -->
+       <exclusions>
+         <exclusion>
+diff --git a/ambari-server/pom.xml.orig b/ambari-server/pom.xml.orig
```
sekikn, line 291: This patch seems unnecessary. Also, would you rename this file to a more explict/descriptive one? (and when renaming, note that Bigtop sorts patches in alphabetical order before applying)

bigtop-packages/src/common/ambari/patch9-z_bonus.diff
```diff
@@ -0,0 +1,446 @@
...
+@@ -27,7 +27,7 @@
+   <groupId>org.apache.ambari</groupId>
+   <artifactId>ambari-serviceadvisor</artifactId>
+   <name>Ambari Service Advisor</name>
+-  <version>1.0.0.0-SNAPSHOT</version>
++  <version>2.7.5.0.0</version>
+   <description>Service Advisor</description>
+ 
+   <dependencies>
+diff --git a/ambari-serviceadvisor/pom.xml.orig b/ambari-serviceadvisor/pom.xml.orig
```
sekikn, line 335: Same as the patch for ambari-server/pom.xml.orig.

bigtop-packages/src/common/ambari/patch9-z_bonus.diff
```diff
@@ -0,0 +1,446 @@
...
+diff --git a/ambari-serviceadvisor/pom.xml b/ambari-serviceadvisor/pom.xml
+index dbfbafad..68d24a97 100644
+--- a/ambari-serviceadvisor/pom.xml
++++ b/ambari-serviceadvisor/pom.xml
+@@ -16,7 +16,7 @@
+    limitations under the License.
+ -->
+ <project xmlns="http://maven.apache.org/POM/4.0.0"
+-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
++         xmlns:xsi="http://www  .w3.org/2001/XMLSchema-instance"
```
sekikn, line 322: Unintended whitespaces.

sekikn:
@cheelio I left some comments on the latest PR. Would you address them during my trial?

bigtop-packages/src/common/ambari/patch9-remove-not-working-plugin-repos.diff
cheelio: Agreed, removed it!

bigtop-packages/src/common/ambari/patch9-z_bonus.diff
cheelio: Yes, done.

bigtop-packages/src/common/ambari/patch9-z_bonus.diff
cheelio: done

bigtop-packages/src/common/ambari/patch9-z_bonus.diff
cheelio: done

cheelio:
Now that Ambari is building smoothly, I will start working on creating a bigtop 3.0.0 mpack.

sekikn:
+1, I ran `./gradlew allclean ambari-pkg-ind -Pnexus=true` twice with a nexus container set up in the same node and confirmed

* the package was successfully built
* .m2/settings.xml was properly configured
* the second run was about 2.3 times faster than the first (22m45s -> 9m43s)

sekikn:
Merged into master and branch-3.0. Thanks a lot @cheelio!

PR #852, opened 2022-01-05
Ambari Mpack fixes for debian/ubuntu

zlokomatic:
Hi,

with these commits i got the .deb for bigtop-ambari-mpack to build and i could install the mpack components in ambari

PR #876, opened 2022-03-26
BIGTOP-3656. Fix Mpack's build and deployment problems on Ubuntu.

sekikn:
### Description of PR

Currently, building Mpack fails on Ubuntu due to an unwanted binary file, and deployment with it also fails due to a wrong repository definition. This PR fixes them.

Built Mpack with this PR and deploy a cluster via Ambari locally.

### For code changes:


sekikn:
Oh, I've just noticed the same fix had already been submitted as #852. I'll close this PR and review that instead.

PR #861, opened 2022-02-08
BIGTOP-3639: Alluxio  build fails on ppc64le

afsanjar:
Alluxio  build fails on ppc64le

PR #877, opened 2022-03-27
BIGTOP-3639. Alluxio build fails on ppc64le (addendum).

sekikn:
### Description of PR

This is a follow-up for #861. Building Alluxio for Debian or Ubuntu fails on a ppc64le machine because fakeroot tries to use the root user's home directory instead of jenkins' to create local maven repository for some reason. This PR avoids this problem by leveraging the maven.repo.local property passed from bigtop-packages/src/deb/alluxio/rules.

Ran `./gradlew allclean alluxio-pkg` in a ubuntu-18.04 container on a ppc64le machine and confirmed it succeeced.

### For code changes:


iwasakims (approved):
Thanks, @sekikn.

sekikn:
Thanks! Merged into master and branch-3.0.

PR #879, opened 2022-03-28
BIGTOP-3657. Revert the download site to archive.apache.org partially

sekikn:
### Description of PR

We change the ASF download site to dlcdn.apache.org in BIGTOP-3648 for avoiding IP ban,
but the old version of Phoenix seems to be removed from that site and it causes build failures.
This PR fixes the problem above.

I confirmed the replaced link was alive via browser.

### For code changes:


iwasakims (approved):
+1

PR #880, opened 2022-03-30
BIGTOP-3658. Add Fedora 35 support and drop 33.

sekikn:
### Description of PR

This PR upgrades supported Fedora version from 33 to 35.

I confirmed that several components were successfully built on Fedora 35 with this PR, including ZK, Hadoop, HBase, Hive, Kafka, etc.

### For code changes:

### DIFF
--- a/bigtop-packages/src/common/hadoop/patch10-MAPREDUCE-7373.diff
+++ b/bigtop-packages/src/common/hadoop/patch10-MAPREDUCE-7373.diff
@@ -1,0 +1,13 @@
+diff --git a/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-nativetask/src/CMakeLists.txt b/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-nativetask/src/CMakeLists.txt
+index ae3b9c6029e..ed03a1b127f 100644
+--- a/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-nativetask/src/CMakeLists.txt
++++ b/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-nativetask/src/CMakeLists.txt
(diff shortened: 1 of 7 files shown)


iwasakims:
@sekikn `./gradlew hadoop-pkg-ind` did not work. The cause seems to be that there is no /sbin/init in the image. Can you address the issue here?

```
$ ./gradlew hadoop-pkg-ind -POS=fedora-35 -Pprefix=trunk
...
++ docker run -d bigtop/slaves:trunk-fedora-35 /sbin/init
+ CONTAINER_ID=f84bd35ba9b3042770ad93fda27dafbf4bd823b9fe7424d6ab963f6199777c85
+ trap 'docker rm -f f84bd35ba9b3042770ad93fda27dafbf4bd823b9fe7424d6ab963f6199777c85' EXIT
+ docker cp /home/fedora/srcs/bigtop f84bd35ba9b3042770ad93fda27dafbf4bd823b9fe7424d6ab963f6199777c85:/bigtop
+ docker cp /home/fedora/srcs/bigtop/bigtop-ci/entrypoint.sh f84bd35ba9b3042770ad93fda27dafbf4bd823b9fe7424d6ab963f6199777c85:/bigtop/entrypoint.sh
+ docker exec f84bd35ba9b3042770ad93fda27dafbf4bd823b9fe7424d6ab963f6199777c85 bash -c 'chown -R jenkins:jenkins /bigtop'
Error response from daemon: Container f84bd35ba9b3042770ad93fda27dafbf4bd823b9fe7424d6ab963f6199777c85 is not running
...
```

sekikn:
Thanks for the comment @iwasakims. I tested the PR on a VM instead of a Docker container, so I missed it. I updated the PR.

Currently, we have to run the *-ind tasks with the following options in case of Fedora 35, which systemctl requires them. I'd like to address it as another issue.

```
./gradlew allclean zookeeper-pkg-ind -POS=fedora-35 -Pdocker-run-option="--tmpfs /run --tmpfs /tmp --cgroupns=host -v /sys/fs/cgroup:/sys/fs/cgroup:rw"
```

iwasakims (approved):
+1. packaging Hadoop succeeded within and without docker. Thanks, @sekikn.

PR #888, opened 2022-04-21
BIGTOP-3671. Add a patch for CVE-2021-22569 to bigtop_toolchain.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3671

https://github.com/protocolbuffers/protobuf/pull/9371 is the fix for vulnerability of protobuf-java. This PR adds backported patch for protobuf-2.5.0. Users can locally install patched protobuf-java from the source code set up by bigtop_toolchain (under /usr/src/protobuf-2.5.0/java) for their own build.


sekikn:
+1, thanks @iwasakims!

iwasakims:
Thanks, @sekikn.

PR #889, opened 2022-04-22
BIGTOP-3672: Always use HTTPS when accessing repos and not plain HTTP.

LeviticusMB:
### Description of PR
Use HTTPS instead of HTTP when building the images.

I was able to use the new Docker image to build HBase for CentOS 8 and Rocky 8.

### For code changes:

- [X] Does the title or this PR starts with the corresponding JIRA issue id (e.g. 'BIGTOP-3638. Your PR title ...')?
- [X] Make sure that newly added files do not have any licensing issues. When in doubt refer to https://www.apache.org/licenses/


sekikn:
+1, tested with some distros and confirmed their images were successfully built.
Thank you for the contribution, @LeviticusMB!

PR #895, opened 2022-04-29
BIGTOP-3675. Building Tez fails due to broken dependency.

sekikn:
### Description of PR

This PR upgrades Tez to 0.10.1 and addresses [BIGTOP-3675](https://issues.apache.org/jira/browse/BIGTOP-3675).
I newly added a patch for TEZ-4214 in addition to TEZ-4411, because without that, shading tez-aux-services failed due to the old version of maven-shade-plugin.

I ensured that Tez was successfully built with this PR on CentOS 7.

### For code changes:


iwasakims (approved):
+1. The issue is reproducible on Rocky Linux 8 too. The patch worked. Thanks, @sekikn.

PR #899, opened 2022-05-06
BIGTOP-3679. Building Phoenix for CentOS 7 fails on aarch64.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3679

protoc provided by org.openlabtesting.protobuf is not portable enough. We should use our own patched protoc.

### DIFF
--- a/bigtop-packages/src/common/phoenix/do-component-build
+++ b/bigtop-packages/src/common/phoenix/do-component-build
@@ -17,20 +17,24 @@
 set -ex
 . `dirname $0`/bigtop.bom
 
+MVN_ARGS="-DskipTests "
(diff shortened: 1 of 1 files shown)


iwasakims:
tested on my CentOS 7 aarch64 VM.

```
$ ./gradlew phoenix-clean phoenix-pkg
...
> Task :phoenix-pkg

BUILD SUCCESSFUL in 22m 11s
7 actionable tasks: 7 executed
```

sekikn:
+1, tested on aarch64 and ppc64le.

sekikn:
Merged into master and branch-3.1. Thanks @iwasakims!

PR #903, opened 2022-05-16
BIGTOP-3685: make docker-provisioner to be used both in docker-compose v1 and v2

yoda-mon:
### Description of PR
Currently docker-provisioner could not used in docker-compose v2.
This PR
- Add `services` level to docker-compose.yml.
  Bigtop is still using deprecated format v1.
  https://docs.docker.com/compose/compose-file/compose-versioning/#version-1-deprecated
  (It is complicated, about yaml format, there are v1, v2, and v3, and v1 is now deprecated.)
  By adding `services` level, we can use the provisioner both in v1 and v2.
- Fix provision_id's name.
  docker-compose v2 seems case sensitive when using -p option.

Check that `./gradlew docker-provisioner` get succeeded in docker-compose `v1.29.2` and `v2.4.0`.

### For code changes:

### DIFF
--- a/provisioner/docker/docker-compose-cgroupv2.yml
+++ b/provisioner/docker/docker-compose-cgroupv2.yml
@@ -13,7 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-bigtop:
+services:
+  bigtop:
     image: ${DOCKER_IMAGE}
     command: /sbin/init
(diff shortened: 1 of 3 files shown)


yoda-mon:
I have to fix the use of `docker inspect --format '{{ .NetworkSettings.IPAddress }}'` in `docker-hadoop.sh`, which only works on the old yaml format.

PR #908, opened 2022-06-02
BIGTOP-3691. Bump Puppet to Puppet7 for Centos7

vivostar:
### Description of PR
bump puppet to puppet7 for centos

1. build a bigtop-puppet docker image 
```shell
cd $BIGTOP_HOME/docker/bigtop-puppet
./buid trunk-centos-7
```
2. create bigtop docker container 
```shell
cd $BIGTOP_HOME/provisioner/docker
./docker-hadoop.sh -c 1
```
### For code changes:


iwasakims:
@vivostar Thanks for the contribution.

This patch can not be merged because it breaks the deployment on distros other than CentOS 7. We must take care of all supported distros at the same time in order to upgrade Puppet.

vivostar:
@iwasakims thanks for your advice, so I made another some patch for debian, ubuntu, and centos8(rocky linux). The advantage of ugdating puppet deploy is that we can use puppet new features suche as hiera v5.  these work are all tested on amd64 machines. there will be some tests to do.

iwasakims:
The patch worked for me on Rocky Linux 8 (x86_64), CentOS 7 (aarch64), Ubuntu 18.04 (aarch64).

> https://lists.apache.org/thread/gbgox39m9wks50hjl986h0syb41b7n3k.

I don't know the context of the thread. If you know the specific failure scenario (distro/arch), let me know. @sekikn

vivostar:
hi, @iwasakims ,the patch now is work well on debian.

sekikn:
@vivostar @iwasakims

> I don't know the context of the thread. If you know the specific failure scenario (distro/arch), let me know.

I was not involved in the discussion in question directly, but as far as I investigated and tried, puppetlabs doesn't seem to support ppc64le.

Steps to reproduce are as follows. I ran them on Ubuntu 22.04.

1. Download CentOS 7 image for ppc64le from any mirror site listed in http://isoredirect.centos.org/altarch/7/isos/ppc64le/.
2. Create VM using the downloaded image (configure parameters in accordance with your environment):

```
$ sudo apt-get install -y libvirt-clients virtinst libvirt-daemon-system qemu-utils qemu-system-ppc
$ sudo virt-install --name=c7ppc --ram=8192 --vcpus=4 --network=bridge=virbr0,model=virtio --os-variant=rhel7.9 --arch=ppc64le --machine=pseries-2.12 --disk=/var/lib/libvirt/images/c7ppc.img,format=qcow2,size=20 --cdrom=CentOS-7-ppc64le-Minimal-2009.iso --serial=pty --console=pty --boot=hd --graphics=none
```

3. Test this PR in the created VM:

```
$ sudo yum update -y
$ sudo yum install -y curl git unzip wget
$ git clone https://github.com/apache/bigtop.git
$ cd bigtop
$ curl -sL https://github.com/apache/bigtop/pull/908.patch | git apply
$ sudo bigtop_toolchain/bin/puppetize.sh

...

failure: repodata/repomd.xml from puppet7: [Errno 256] No more mirrors to try.
http://yum.puppetlabs.com/puppet7/el/7/ppc64le/repodata/repomd.xml: [Errno 14] HTTP Error 404 - Not Found
bigtop_toolchain/bin/puppetize.sh: line 60: /opt/puppetlabs/bin/puppet: No such file or directory
```

vivostar:
hi, @sekikn , I find ppc64 puppet pakages on the url 
[https://nightlies.puppetlabs.com/](url)
so I update the patch. please help test it on the ppc64 machine, thanks.

bigtop_toolchain/bin/puppetize.sh
```diff
@@ -20,51 +20,74 @@ if [ -f /etc/os-release ]; then
     . /etc/os-release
 fi
 
+RPM_DOWNLOAD_URL="https://yum.puppet.com/puppet7"
+APT_DOWNLOAD_URL="https://apt.puppet.com/puppet7"
+NIGHTLY=""
+
+if [ "${ARCH}" != "aarch64" -a "${ARCH}" != ""];then
```
iwasakims, line 27: needs space before `]`.

```
/tmp/puppetize.sh: line 27: [: too many arguments
```

vivostar: thanks, I will fix this.

vivostar:
now the patch is work well on bigtop-slaves images.

sekikn:
@vivostar I can try the latest PR on ppc64le, but I still have a few concerns...

* I couldn't find the ppc64le package for Fedora 35, Debian or Ubuntu, either on the official or nightly repo.
* The CentOS 7 ppc64le package seems to be unattended. Its timestamp is 2020-10-27.
* Though I'm not sure how the nightly repo is operated actually, its name sounds like it's unstable and not suitable for production use.

All of the advantages of upgrading Puppet you mentioned above are certainly reasonable, but considering these circumstances, upgrading to Puppet 7 looks a bit too early to me.
If the puppet community has a plan to support ppc64le and the packages will be provided in the official repository, how about putting off the upgrade after that?

vivostar:
> @vivostar I can try the latest PR on ppc64le, but I still have a few concerns...
> 
> * I couldn't find the ppc64le package for Fedora 35, Debian or Ubuntu, either on the official or nightly repo.
> * The CentOS 7 ppc64le package seems to be unattended. Its timestamp is 2020-10-27.
> * Though I'm not sure how the nightly repo is operated actually, its name sounds like it's unstable and not suitable for production use.
> 
> All of the advantages of upgrading Puppet you mentioned above are certainly reasonable, but considering these circumstances, upgrading to Puppet 7 looks a bit too early to me.
> If the puppet community has a plan to support ppc64le and the packages will be provided in the official repository, how about putting off the upgrade after that?

That's right, updating puppet is non-trivial.

bigtop_toolchain/bin/puppetize.sh
```diff
@@ -20,51 +20,74 @@ if [ -f /etc/os-release ]; then
...
+        yum update -y
+        yum -y install hostname curl sudo unzip wget puppet-agent
+        # bump puppet to puppet7
+        /opt/puppetlabs/bin/puppet module install puppetlabs-stdlib
         ;;
     centos-8*)
         sed -i -e 's/^\(mirrorlist\)/#\1/' -e 's,^#baseurl=http://mirror.centos.org,baseurl=https://vault.centos.org,' /etc/yum.repos.d/CentOS-Linux-*
         ;&
     rocky-8*)
         rpm -ivh https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
```
brahmareddybattula, line 76: why we need still have this..? we can remove right as puppet is not there in epel.?

vivostar: yes, the puppet is not there in epel now

bigtop_toolchain/bin/puppetize.sh
```diff
@@ -20,51 +20,74 @@ if [ -f /etc/os-release ]; then
...
 fi
 
+RPM_DOWNLOAD_URL="https://yum.puppet.com/puppet7"
+APT_DOWNLOAD_URL="https://apt.puppet.com/puppet7"
+NIGHTLY=""
+
+if [ "${ARCH}" = "ppc64le" ];then
+RPM_DOWNLOAD_URL="https://nightlies.puppetlabs.com/yum/puppet7-nightly"
+APT_DOWNLOAD_URL="https://nightlies.puppetlabs.com/apt/puppet7-nightly"
+NIGHTLY="-nightly"
```
pan3793, line 30: does it[1][2] mean puppet7 is ready for ppc64le now?

[1] https://apt.puppet.com/dists/focal/puppet7/index.html
[2] https://yum.puppet.com/puppet7/el/8/index.html

PR #913, opened 2022-06-14
BIGTOP-3701: Remove deprecated Hadoop fs-image after uninstalling HDF namenode

guyuqi:
### Description of PR
When uninstalled Hadoop-2.10 and reinstalled Hadoop-3.2, it would fail to format namenode and The user have to restart NameNode with the "-rollingUpgrade started" option if a rolling upgrade is already started; or restart NameNode with the "-upgrade" option to start a new upgrade. 
It would confuse some new BIgtop Hadoop users. 

It's better to remove deprecated Hadoop fs-image file after uninstalling HDFS-namenode when uninstalling package.

Build Hadoop by Bigtop and install/remove Hadoop package.
```
e.g. :
rpm -e hadoop-hdfs-3.2.3-1.el7.x86_64.rpm
or
rpm -qa | grep hadoop | sudo xargs yum -y remove
```

### For code changes:

### DIFF
--- a/bigtop-packages/src/rpm/hadoop/SPECS/hadoop.spec
+++ b/bigtop-packages/src/rpm/hadoop/SPECS/hadoop.spec
@@ -652,6 +652,13 @@ if [ "$1" = 0 ]; then
   %{alternatives_cmd} --remove %{name}-conf %{etc_hadoop}/conf.empty || :
 fi
 
+%preun hdfs
+if [ $1 = 0 ]; then
+  if [ -d "/hadoop/hdfs" ]; then
+      rm -rf /hadoop/hdfs
+  fi
+fi
+
 %preun httpfs
 if [ $1 = 0 ]; then
   service %{name}-httpfs stop > /dev/null 2>&1


iwasakims:
-1. Users should do this dangerous operation on their own. We usually expect user-generated data to be preserved even after uninstalling (rpm/deb) package. If users want to format HDFS overwriting existing fsimage, `hdfs namenode -format -force` can be used on their responsibility.

guyuqi:
@iwasakims Thanks for your comments.
Just as you mentioned,  users could overwrite the existing fsimage by `hdfs namenode -format -force` or `-upgrade`, but not all new comers would know how to handle it when the issues occurred like: 
```
Failed to start namenode.
java.io.IOException:
File system image contains an old layout version -63.
An upgrade to version -65 is required.
.....
...
```

Hdfs is the distributed file system to make data safe.  `Data replication `and `secondarynamenode` would help recover one node from the other nodes of the same cluster if fsimage and edit files were deleted.   (Is it really dangerous?)
From my perspective, if users remove(`$1 = 0` here, not upgrade) the Hadoop packages from one node, it seems there is no necessary to retain the deprecated data. 

On the other hand, it's convennient for users to deployment different Hadoop verisons by Mpack (or other automatic deployment tools) if we removed deprecated Hadoop fs-image after RPM/deb uninstallation.

Is it the institutionalized acceptance that we don't touch `/hadoop/hdfs` even if there would be a obstacle to new comers?

iwasakims:
The paths for containing HDFS data are not defined by rpm/deb. Users arbitrarily specify them (by `dfs.namenode.name.dir` and `dfs.datanode.data.dir` in hdfs-site.xml). If the `/hadoop/hdfs` is defined in Puppet manifests, it shoud be cared by them. Modifying .spec to remove them does not make sense at all.

PR #929, opened 2022-07-05
BIGTOP-2971. RPM packaging for Ranger.

sekikn:
This PR is based on #326 submitted by @Xingwd (much appreciated!) and
includes the following updates:

* Upgrade Ranger version to 2.2.0

* Change the download site from ASF to GitHub, because the source
  tarball distributed on the former is unnecessarily large

* Add package dependencies to bigtop.bom

* Remove the assembly:assembly target from the mvn command since it
  doesn't work in recent releases

* Add the version info of other components to the mvn command

* Fix the lib paths in install_ranger.sh and the RPM spec file

* Add the #BIGTOP_PATCH_FILES and #BIGTOP_PATCH_COMMANDS comments to the
  RPM spec so that the patches provided by Bigtop are applied during
  building RPM

* Add patches for addressing RANGER-3373 and RANGER-3459

* Remove symlink to non-existent ojdbc6.jar in install_ranger.sh

* Replace the hard-coded paths to /usr/lib/rpm/redhat in the RPM spec
  with %{_rpmconfigdir} (cf. BIGTOP-1344)

* Define debug_package=nil in the RPM spec to avoid build failure on
  Fedora 35

### Description of PR

See https://issues.apache.org/jira/browse/BIGTOP-2971.

Ran `./gradlew allclean ranger-pkg repo -Dbuildwithdeps=true` on CentOS 7, Rocky Linux 8, and Fedora 35 (x86_64).

### For code changes:

### DIFF
--- a/bigtop-packages/src/common/ranger/do-component-build
+++ b/bigtop-packages/src/common/ranger/do-component-build
@@ -1,0 +1,35 @@
+#!/bin/sh
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


guyuqi:
Successfully build rpm packages on Arm64 Rockylinux-8.
+1, LGTM.

It would be appreciated if Deb packaging scripts and smoke tests were provided. :)

PR #933, opened 2022-07-06
BIGTOP-3717: Update MapReduce service configuration in Mpack

guyuqi:
Mapred configuration macros are deprected (HADOOP_MAPRED_ROOT_LOGGER,
HADOOP_JOB_HISTORYSERVER_OPTS, HADOOP_MAPRED_NICENESS, etc.).

Update them in Mpack aligned to
https://github.com/apache/hadoop/blob/trunk/hadoop-mapreduce-project/bin/mapred-config.sh#L38-L54

PR #932, opened 2022-07-06
BIGTOP-3718: Fix MapReduce2 service-check in Amabri Mpack

guyuqi:
The patch is to fix the issues of checking MapReduce2 service and
"JAR does not exist".

PR #931, opened 2022-07-06
BIGTOP-3719: Decouple Mpack HDFS services from Ambari Core

guyuqi:
Decouple HDFS services from Ambari Core and fix the issues of HDFS
deployment from Bigtop 3.x repo. It's convenient for us to modify/upgrade
Mpack without re-building and re-deploying Ambari respectively.

PR #934, opened 2022-07-06
BIGTOP-3716: Rename the Yarn/MapReduce historyserver PID file

guyuqi:
The naming rule of Yarn/MapReduce historyserver PID file
is deprecated in Hadoop 3.x.


guyuqi:
The PRs ( https://github.com/apache/bigtop/pull/931 https://github.com/apache/bigtop/pull/932 https://github.com/apache/bigtop/pull/933 https://github.com/apache/bigtop/pull/934 ) make Mpack support Bigtop 3.1:

iwasakims (approved):
LGTM.

I think the second part of the pid file name is not a user name but the value of HADOOP_IDENT_STRING.
https://github.com/apache/hadoop/blob/rel/release-3.3.3/hadoop-common-project/hadoop-common/src/main/bin/hadoop-functions.sh#L2699

HADOOP_IDENT_STRING is defined in the default file of the service unless Ambari overrides it.
https://github.com/apache/bigtop/blob/47ee254af1fb7ff131bdc36bed12d66b85ac63b9/bigtop-packages/src/common/hadoop/hdfs.default#L20

Anyway the prefix is always `hadoop-` and the second parts are `hdfs`, `yarn`, `mapred` unless Ambari overrides it.

PR #964, opened 2022-07-23
[WIP]BIGTOP-3760. Bump Hadoop to 3.3.4.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3760

Hadoop 3.3.4 RC0 is under vote.
https://lists.apache.org/thread/r41ftyrs4b6g67sf8nz8xypfwohov7qb

PR #970, opened 2022-07-29
BIGTOP-3763: Fix Hive build issues about protobuf 2.5.0 on Arm64 and PPC64le

guyuqi:
### Description of PR

1.Fix Hive build issues about protobuf 2.5.0 on Arm64 and
PPC64le when upgrading hadoop to 3.3.3/3.3.4,
2.Expand Java heap size.
### For code changes:
Build Hive and run smoke tests

### DIFF
--- a/bigtop-packages/src/common/hive/do-component-build
+++ b/bigtop-packages/src/common/hive/do-component-build
@@ -36,12 +36,12 @@ HIVE_MAVEN_OPTS=" -Dhbase.version=$HBASE_VERSION \
 # for non-x86 platforms
 if [ $HOSTTYPE = "powerpc64le" ] ; then
   mvn install:install-file -DgroupId=com.google.protobuf -DartifactId=protoc -Dversion=2.5.0 \
-    -Dclassifier=linux-ppcle_64 -Dpackaging=exe -Dfile=/usr/local/bin/protoc
+    -Dclassifier=linux-ppcle_64 -Dpackaging=exe -Dfile=/usr/local/protobuf-2.5.0/bin/protoc
 elif [ $HOSTTYPE = "aarch64" ] ; then
   mvn install:install-file -DgroupId=com.google.protobuf -DartifactId=protoc -Dversion=2.5.0 \
-    -Dclassifier=linux-aarch_64 -Dpackaging=exe -Dfile=/usr/local/bin/protoc
+    -Dclassifier=linux-aarch_64 -Dpackaging=exe -Dfile=/usr/local/protobuf-2.5.0/bin/protoc
 fi
-export MAVEN_OPTS="${MAVEN_OPTS} -Xmx1500m -Xms1500m"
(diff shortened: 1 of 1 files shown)


guyuqi:
Tested it on Arm64 rockylinux8:
```
 ./docker-hadoop.sh --create 3 \
--memory 16g \
--disable-gpg-check --enable-local-repo \
--stack hdfs,yarn,mapreduce,hbase,hive \
--smoke-tests hive
```

bigtop-packages/src/common/hive/do-component-build
```diff
@@ -36,12 +36,12 @@ HIVE_MAVEN_OPTS=" -Dhbase.version=$HBASE_VERSION \
...
   mvn install:install-file -DgroupId=com.google.protobuf -DartifactId=protoc -Dversion=2.5.0 \
-    -Dclassifier=linux-ppcle_64 -Dpackaging=exe -Dfile=/usr/local/bin/protoc
+    -Dclassifier=linux-ppcle_64 -Dpackaging=exe -Dfile=/usr/local/protobuf-2.5.0/bin/protoc
 elif [ $HOSTTYPE = "aarch64" ] ; then
   mvn install:install-file -DgroupId=com.google.protobuf -DartifactId=protoc -Dversion=2.5.0 \
-    -Dclassifier=linux-aarch_64 -Dpackaging=exe -Dfile=/usr/local/bin/protoc
+    -Dclassifier=linux-aarch_64 -Dpackaging=exe -Dfile=/usr/local/protobuf-2.5.0/bin/protoc
 fi
-export MAVEN_OPTS="${MAVEN_OPTS} -Xmx1500m -Xms1500m"
+export MAVEN_OPTS="${MAVEN_OPTS} -Xmx2500m -Xms2500m"
```
iwasakims, line 44: How about removing `-Xms1500m` to make JVM allocate heap on demand? We can not expect much gain by allocating maximum heap from the beginning for short term task like building products. I got no OOM with current settings on my x86_64 environment.

guyuqi: It makes sense.  Also tested it on my Arm64 environment.

iwasakims: While `-Xmx2500m` might be still useful to allocate enough heap on machine with memory less than 5GB (because the default maximum heap of JVM is capped at half of the machine's memory IIRC), we can not practically use such a small machine for Bigtop development. I think dropping `-Xmx` too should be ok.

iwasakims (approved):
LGTM. tested on CentOS 7 aarch64 .

While I got [the error I commented to BIGTOP-3760](https://issues.apache.org/jira/browse/BIGTOP-3760?focusedCommentId=17572800&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17572800), applying the patch of #964 fixed the issue.

PR #971, opened 2022-07-30
BIGTOP-3765: Remove cmake conditional for Hadoop rpm spec

guyuqi:
### Description of PR
The cmake3 would be always installed on centos-7:
https://github.com/apache/bigtop/blob/master/bigtop_toolchain/manifests/packages.pp#L69-#L73

We can remove the cmake conditional for cmake3 was always provided in CentOS 7.

Builf Hadoop against on Centos-7.
Run smoke tests.

### For code changes:

### DIFF
--- a/bigtop-packages/src/rpm/hadoop/SPECS/hadoop.spec
+++ b/bigtop-packages/src/rpm/hadoop/SPECS/hadoop.spec
@@ -176,11 +176,6 @@ Source31: kms.default
 #BIGTOP_PATCH_FILES
 Buildroot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id} -u -n)
 BuildRequires: fuse-devel, fuse
-%if %{?el7}0
-BuildRequires: cmake3
-%else
-BuildRequires: cmake
-%endif
 Requires: coreutils, /usr/sbin/useradd, /usr/sbin/usermod, /sbin/chkconfig, /sbin/service, bigtop-utils >= 0.7, zookeeper >= 3.4.0
 Requires: psmisc, %{netcat_package}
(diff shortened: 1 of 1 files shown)


iwasakims:
I think I misunderstood the intent of the conditional. Only CentOS 7 needs `BuildRequires: cmake3` since CMake 2 is provided as **cmake** and CMake 3 is provided as **cmake3**. Newer RHEL family provides CMake 3 as **cmake**. The conditional is working as expected.

CentOS 7:
```
$ rpm -q cmake
cmake-2.8.12.2-2.el7.x86_64

$ rpm -q cmake3
cmake3-3.17.5-1.el7.x86_64
```

Rocky Linux 8:
```
$ rpm -q cmake
cmake-3.20.2-4.el8.x86_64
```

iwasakims:
@guyuqi Do you still want to remove the conditional at all? I think it makes sence if we need to install CMake from source code in the future.

guyuqi:
Hi @iwasakims ,
In Centos7 (`bigtop/slaves:trunk-centos-7` ` x86_64/arm64`):
`cmake3` was pre-installed and default linked to `cmake`:
```
[root@a06bcc19ce22 /]# whereis cmake
cmake: /usr/bin/cmake
[root@a06bcc19ce22 /]# ls /usr/bin/cmake -al
lrwxrwxrwx 1 root root 15 May 13 01:55 /usr/bin/cmake -> /usr/bin/cmake3
```

IMHO,  it makes sense to remove cmake conditional here.

iwasakims (approved):
+1. We can not set exhaustive BuildRequires anyway. Thanks, @guyuqi.

PR #978, opened 2022-08-10
BIGTOP-3774: Drop Docker Sandbox and Vagrant Provisioner

evans-ye:
### Description of PR
As title :)

No test as the PR is to remove existing modules.


guyuqi:
LGTM,  +1.

And what about also removing: 
 ` "By applying the snippet, Vagrant will be installed(the Docker installation will be added soon)."`  in [bigtop_toolchain/README](https://github.com/apache/bigtop/blob/master/bigtop_toolchain/README.md)

evans-ye:
@guyuqi good point! Thanks for discovering it. I've add a commit to address it.

guyuqi:
Merged, thanks @evans-ye

PR #981, opened 2022-08-12
BIGTOP-3770: Remove Sqoop from Bigtop stack

guyuqi:
https://issues.apache.org/jira/browse/BIGTOP-3770
Apache Sqoop was retired last year.
https://lists.apache.org/thread/pkylvs0qrpmpxdlgfdmdj32rjk9q24x4


iwasakims (approved):
LGTM. I built Oozie and ran smoke-tests to see if any regression. Thanks, @guyuqi.

PR #987, opened 2022-08-25
 [WIP]BIGTOP-3741:Add Flink support for Bigtop 3.1.0 Mpack

smallyao:
### Description of PR
Add flink support for Bigtop 3.1.0 Mpack

Tested in centos 7.6 
Flink version is 1.15.0

1 summary
image

2 configs
image

3 service check
image


kevinw66:
We should add more configuration template from `/etc/flink/conf`, like `log4j.properties`

PR #1008, opened 2022-09-13
BIGTOP-3814. Deploying Spark ThriftServer fails because it's launched too early.

sekikn:
### Description of PR

- This PR defers Spark ThriftServer's launch after initializing HDFS so that the launch succeeds.

- Ran `./docker-hadoop.sh -d -C config_rockylinux-8.yaml -r file:///bigtop-home/output -G -k hdfs,yarn,spark -c 3` locally.

### For code changes:

### DIFF
--- a/bigtop-deploy/puppet/modules/spark/manifests/init.pp
+++ b/bigtop-deploy/puppet/modules/spark/manifests/init.pp
@@ -63,7 +63,8 @@ class spark {
       hasrestart => true,
       hasstatus  => true,
     }
-    
+
+    Exec<| title == "init hdfs" |> -> Service["spark-thriftserver"]
   }
 
   class client {


iwasakims:
@sekikn Hmm. Deployment by the following cmdline stucked.

```
$ ./docker-hadoop.sh \
       --create 1 \
       --memory 16g \
       --disable-gpg-check \
       --stack hdfs,yarn,mapreduce,spark
```

How about using dependency between classes following existing convention? like

```
diff --git a/bigtop-deploy/puppet/modules/spark/manifests/init.pp b/bigtop-deploy/puppet/modules/spark/manifests/init.pp
index 06d804e5..bce16cff 100644
--- a/bigtop-deploy/puppet/modules/spark/manifests/init.pp
+++ b/bigtop-deploy/puppet/modules/spark/manifests/init.pp
@@ -42,6 +42,7 @@ class spark {

     if ('spark-thriftserver' in $roles) {
       include spark::spark_thriftserver
+      Class['Hadoop::Init_hdfs'] -> Class['Spark::Spark_thriftserver']
     }

   }
```

iwasakims:
Sorry, @sekin. It looks installing spark-sparkr is stucked in the container process.

```
/usr/libexec/platform-python /usr/bin/dnf -d 0 -e 1 -y install spark-sparkr
```

While the patch should not be the cause, I think `Class['Hadoop::Init_hdfs'] -> Class['Spark::Spark_thriftserver']` is easier to understand the dependency between modules.

iwasakims (approved):
Thanks for the explanation, @sekikn. +1 on the current patch if the manifest is expecting deployment without Hadoop.

sekikn:
Merged into master. Thank you for the review @iwasakims!

PR #1009, opened 2022-09-14
BIGTOP-3812: Building bigtop-ambari-mpack fails on Debian/Ubuntu due to unwanted binary files

guyuqi:
https://issues.apache.org/jira/browse/BIGTOP-3812

This PR is to fix the building issues of bigtop-ambari-mpack on Debian/Ubuntu due to unwanted binary files. 

After the discussion with Zhiguo Gu, ` bigtop-ambari-mpack` will be selected as the default stack for Ambari and it will be moved back to Ambari core after Bigtop-3.3 and Ambari-2.8.
So we move 'dev-support' files to Bigtop provisioner and rename it to bigtop-stack-provisioner. It is used for stack provisioner which is the test tool for Mpack developers. 
As mentioned before,  it still will be used for Ambari UI test framework and Ambair-bigtop-demo after ` bigtop-ambari-mpack` moved back to Ambari core.


guyuqi:
Merged.

PR #1007, opened 2022-09-13
BIGTOP-3810: Fix wrong home configuration for HTTPFS on build scripts

kevinw66:
### For code changes:

- [ ] Does the title or this PR starts with the corresponding JIRA issue id (e.g. 'BIGTOP-3638. Your PR title ...')?
- [ ] Make sure that newly added files do not have any licensing issues. When in doubt refer to https://www.apache.org/licenses/

PR #1011, opened 2022-09-15
BIGTOP-3815. Fix failure of installing hadoop-httpfs deb due to change of httpfs home dir.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3815

This is a follow-up of #1007 causing following error.

```
Setting up hadoop-httpfs (3.3.4-1) ...
chown: cannot access '/var/run/hadoop-httpfs': No such file or directory
dpkg: error processing package hadoop-httpfs (--configure):
```


guyuqi:
+1

guyuqi:
Merged.

PR #1012, opened 2022-09-15
BIGTOP-3623. dh_strip_nondeterminism spends too much time on deb packaging.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3623

strip-nondeterminism on deb packaging of Hadoop and Spark takes several hours after upgrading to Hadoop 3.x due to expanded dependencies and shaded client. Since building Bigtop packages are not so [reproducible](https://wiki.debian.org/ReproducibleBuilds/About) by its nature, we should be ok to skip it now.

The same fix is already applied for Ambari, Hive and Oozie already.


guyuqi:
It makes sense, shall we apply it to all other components for Debian packaging?

elukey:
LGTM!

iwasakims:
> It makes sense, and shall we apply it to all other components for Debian packaging?

I will file another JIRA for time consuming products if any after merging this. @guyuqi

iwasakims:
I merged this to master and branch-3.1. Thanks, @guyuqi and @elukey.

PR #1033, opened 2022-10-12
BIGTOP-3840. Enable short circuit local read in HDFS deployment by default.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3840

Short circuit local read is important feature for data store like HBase. It should be enabled by Puppet provisioner for testing the feature of libhadoop.so.


iwasakims:
Tested this by running smoke-tests of hdfs, yarn, mapreduce, hbase on Rocky Linux 8 x86_64 and Ubuntu 18.04 aarch64.

sekikn:
+1, tested on CentOS 7.

```
$ curl -sL https://github.com/apache/bigtop/pull/1033.diff | git apply 
$ ./gradlew allclean hadoop-pkg repo -Dbuildwithdeps=true
$ cd provisioner/docker
$ ./docker-hadoop.sh -d -L -k hdfs -c 1
$ ./docker-hadoop.sh -e 1 grep -C2 shortcircuit /etc/hadoop/conf/hdfs-site.xml

  <property>
    <name>dfs.client.read.shortcircuit</name>
    <value>true</value>
  </property>
```

iwasakims:

PR #1035, opened 2022-10-13
BIGTOP-3845: Add the command to create the /usr/bigtop folder in the bigtop-select rpm package

timyuer:
BIGTOP-3845: Add the command to create the /usr/bigtop folder in the bigtop-select rpm package

### For code changes:

### DIFF
--- a/bigtop-packages/src/common/bigtop-select/install_select.sh
+++ b/bigtop-packages/src/common/bigtop-select/install_select.sh
@@ -67,10 +67,12 @@ LIB_DIR=${LIB_DIR:-/usr/lib/bigtop-select}
 BIN_DIR=${BIN_DIR:-/usr/bin}
 CONF_DIR=${CONF_DIR:-/etc/bigtop-select/conf.dist}
 
+STACK_ROOT_DIR=/usr/bigtop
 STACK_SELECTOR=distro-select
 CONF_SELECTOR=conf-select
 
 # Install packages
 install -d -p -m 755 $PREFIX${LIB_DIR}/
+install -d -p -m 755 $PREFIX${STACK_ROOT_DIR}/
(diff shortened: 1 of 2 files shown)


kevinw66:
+1 LGTM, tested with prefixed version of Zookeeper and Kafka

iwasakims (approved):
LGTM. Thanks, @timyuer and @kevinw66.

sekikn:
@timyuer @kevinw66 Thanks for the contribution! Just a question, `STACK_ROOT_DIR` is a constant in this PR, but is it still valid if setting another value for the `parentDir` option instead of /usr/bigtop? e.g., `./gradlew allclean pkgs -Dbuildwithdeps=true -PparentDir=/usr/local/bigtop`

kevinw66:
> @timyuer @kevinw66 Thanks for the contribution! Just a question, `STACK_ROOT_DIR` is a constant in this PR, but is it still valid if setting another value for the `parentDir` option instead of /usr/bigtop? e.g., `./gradlew allclean pkgs -Dbuildwithdeps=true -PparentDir=/usr/local/bigtop`

Hi @sekikn , thanks for your reminder, but we have one more question here, currently the version info in parentDir need to be added explicitly like `-PparentDir=/usr/bigtop/3.2.0`, should we change the behavior to read from bigtop version and remove `-SNAPSHOT` automatically?

sekikn:
> should we change the behavior to read from bigtop version and remove `-SNAPSHOT` automatically?

@kevinw66 I think that's OK since it doesn't affect the default (without the parentDir option) behavior at all.

kevinw66:
> > should we change the behavior to read from bigtop version and remove `-SNAPSHOT` automatically?
> 
> @kevinw66 I think that's OK since it doesn't affect the default (without the parentDir option) behavior at all.

Automatically add version info might be convenience if we wants to use bigtop-select without Ambari in the future

PR #1041, opened 2022-10-17
BIGTOP-3850: Conflict occurs when installing both ranger-hdfs-plugin …

eubnara:
…and ranger-yarn-plugin

### Description of PR

`ranger-hdfs-plugin` and `ranger-yarn-plugin` install jars on the same directory, `/usr/lib/hadoop/lib`. If both NameNode and ResourceManager are installed on the same machine, it could be a problem when installing or upgrading plugins respectively. It should be separated.

Manually tested on private on-premise hadoop cluster.
NameNode and ResourceManager are located on the same machine.
Try install & upgrade `ranger-hdfs-plugin` and `ranger-yarn-plugin` respectively.

### For code changes:

- [x] Does the title or this PR starts with the corresponding JIRA issue id (e.g. 'BIGTOP-3850. Your PR title ...')?


eubnara:
@JiaLiangC @iwasakims 
Could you review it please?

PR #1044, opened 2022-10-24
BIGTOP-3856: Add hive-hcat-client in distro-select

timyuer:
### For code changes:

### DIFF
--- a/bigtop-packages/src/common/bigtop-select/distro-select
+++ b/bigtop-packages/src/common/bigtop-select/distro-select
@@ -55,6 +55,7 @@ leaves = {
            "hive-metastore": "hive",
            "hive-server2": "hive",
            "hive-webhcat": "hive-hcatalog",
+           "hive-hcat-client": "hive-hcatalog",
            "kafka-broker": "kafka",
            "oozie-client": "oozie",
            "oozie-server": "oozie",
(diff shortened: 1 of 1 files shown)


bigtop-packages/src/common/bigtop-select/distro-select
```diff
@@ -55,6 +55,7 @@ leaves = {
            "hive-metastore": "hive",
            "hive-server2": "hive",
            "hive-webhcat": "hive-hcatalog",
+           "hive-hcat-client": "hive-hcatalog",
```
iwasakims, line 58: hive-hcatalog-client (without abbreviation) sounds consistent with other names. Is the term `hive-hcat` already used in Ambari? @timyuer.

timyuer: Yes, this name looks better, but Ambari has already used the name Hcat. Please give me some time to evaluate the amount of work needed to change it. @iwasakims

timyuer: Oops, in the early Ambari version Hcat Client and Webhcat Server used the same Link `hive-webhcat`, so this PR doesn't seem to make sense, I'm going to close this PR. Thanks for your review.

PR #1050, opened 2022-11-09
BIGTOP-3868: Add an error message file check to avoid the error exit

yoda-mon:
### Description of PR
If users use docker provisioner from gradle, it fails. This is because `docker-hadoop.sh` returns 1 even if the provisioning was secceeded.

```sh
$ ./gradlew -Pstack=zookeeper docker-provisioner
...
Notice: Finished catalog run in 74.78 seconds
> Task :docker-provisioner FAILED
```

This PR add an error message check before printing the file, to avoid the error exit.

```sh
$ ./gradlew -Pstack=zookeeper docker-provisioner
...
Notice: Finished catalog run in 65.20 seconds

BUILD SUCCESSFUL in 1m 29s
```

### For code changes:

### DIFF
--- a/provisioner/docker/docker-hadoop.sh
+++ b/provisioner/docker/docker-hadoop.sh
@@ -196,7 +196,7 @@ provision() {
         ) &
     done
     wait
-    if [ -f .error_msg_* ]; then
+    if [ `find . -maxdepth 1 -type f -name ".error_msg_*" | wc -l` -gt 0 ]; then
       cat .error_msg_* 2>/dev/null && exit 1
     fi
 }
--- a/provisioner/docker/docker-hadoop.sh
+++ b/provisioner/docker/docker-hadoop.sh
@@ -196,7 +196,9 @@ provision() {
         ) &
     done
     wait
-    cat .error_msg_* 2>/dev/null && exit 1
+    if [ -f .error_msg_* ]; then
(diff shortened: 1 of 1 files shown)


provisioner/docker/docker-hadoop.sh
```diff
@@ -196,7 +196,9 @@ provision() {
         ) &
     done
     wait
-    cat .error_msg_* 2>/dev/null && exit 1
+    if [ -f .error_msg_* ]; then
```
iwasakims, line 199: `test -f` takes only one argument. If there are multiple files matching `.error_msg_*`, this results in syntax error?

```
$ touch foo.1 foo.2
$ if [ -f foo.* ] ; then echo foo ; fi
-bash: [: foo.1: binary operator expected
```

yoda-mon: Thank you,  I forgot considering about that. I'll fix this with another way.

yoda-mon: I changed to use `find` to be able to pick up multiple files.

```sh
touch .error_msg_1
touch .error_msg_0
if [ `find . -maxdepth 1 -type f -name ".error_msg_*" | wc -l` -gt 0 ]; then echo "true"; fi

true
```

iwasakims (approved):
LGTM. Thanks, @yoda-mon.

PR #1052, opened 2022-11-15
BIGTOP-3869: Decouple mpack Zookeeper service from Ambari Core

guyuqi:
https://issues.apache.org/jira/browse/BIGTOP-3869

- Decouple mpack Zookeeper service from Ambari Core
- Fix zookeeper service alert.


iwasakims:
@guyuqi Mpack looks moved to Ambari by [AMBARI-25732](https://issues.apache.org/jira/browse/AMBARI-25732). Why do we still need this in Bigtop??

guyuqi:
> @guyuqi Mpack looks moved to Ambari by [AMBARI-25732](https://issues.apache.org/jira/browse/AMBARI-25732). Why do we still need this in Bigtop??

Hi @iwasakims, 
Yes, Ambari-2.8 would take Bigtop Mpack as the default stack, but there are still much work in Ambari-2.8 development and it hasn't been released yet. 
There also maybe some work to support Ambari-2.8 in Bigtop and we would remove Mpack after Ambari-2.8 is released.

Currently, Bigtop offers the support for Ambari-2.7.5 (the latest stable version),  and its Mpack including more components and visual metrics (Ambari metrics service) are the main features of the comming 3.2 release which would provide the whole Bigdata management platform and software stack to make management of Hadoop easier and simpler for the developers and administrators.  

This PR is to fix the ZK connection issues: Expected response imok, Actual response ruok is not executed because it is not in the whitelist.

guyuqi:
The ZK Mpack service works well on Bigtop stack cluster:

PR #1053, opened 2022-11-15
BIGTOP-3870: Adjust Phoenix dependency of phoenix-hbase-compat for HBase-2.13 and Hadoop-3.3.4

guyuqi:
https://issues.apache.org/jira/browse/BIGTOP-3870

To fix the issues of Ambari metrics, Phoenix should be algined to Hadoop and Hbase version which are conformance to Bigtop stack version.  (For example, we should update HBase `2.4.1+` to `2.4.13` for `phoenix-hbase-compat`)

PR #1054, opened 2022-11-15
BIGTOP-3872: Adjust Ambari-2.7.5 dependency to fix the deployment issues of Ambari metrics

guyuqi:
https://issues.apache.org/jira/browse/BIGTOP-3872

Adjust Ambari-2.7.5 dependency to fix the deployment issues of Ambari metrics. When users deploy Hadoop components with Bigtop stack,  Ambari mertrics dependency should be updated to fix the related issues.

### DIFF
--- a/bigtop-packages/src/common/ambari/patch12-fix-ambari-metrics-grafana-service-issues.diff
+++ b/bigtop-packages/src/common/ambari/patch12-fix-ambari-metrics-grafana-service-issues.diff
@@ -1,0 +1,44 @@
+diff --git a/ambari-metrics/ambari-metrics-grafana/conf/unix/ambari-metrics-grafana b/ambari-metrics/ambari-metrics-grafana/conf/unix/ambari-metrics-grafana
+index dba2f5ff55..edd3ec1a09 100644
+--- a/ambari-metrics/ambari-metrics-grafana/conf/unix/ambari-metrics-grafana
++++ b/ambari-metrics/ambari-metrics-grafana/conf/unix/ambari-metrics-grafana
+@@ -118,7 +118,7 @@ case "$1" in
+     then
+       for i in {1..10}
+       do
+-        sleep 2
++        sleep 12
+         # check if pid file has been written to
+         if [ -s $PID_FILE ]; then
+           echo " $(date) pid_file has been written to" >> $LOG_FILE
+@@ -131,18 +131,19 @@ case "$1" in
+         echo " $(date) Start FAILED because daemon did not write pid in pid_file after 20 seconds" >> $LOG_FILE
+         exit 1
+       fi
+-      i=0
+-      timeout=60
++      #i=0
++      #timeout=60
+       # Wait for the process to be properly started before exiting
+-      until { cat "$PID_FILE" | xargs kill -0; } >/dev/null 2>&1
+-      do
+-        sleep 1
+-        i=$(($i + 1))
+-        if [ $i -gt $timeout ]; then
+-          echo "FAILED"
+-          exit 1
+-        fi
+-      done
++      sleep 120
++      #until { cat "$PID_FILE" | xargs kill -0; } >/dev/null 2>&1
++      #do
++      #  sleep 1
++      #  i=$(($i + 1))
++      #  if [ $i -gt $timeout ]; then
++      #    echo "FAILED"
++      #    exit 1
++      #  fi
++      #done
+     fi
+ 
+     echo "OK" >> $LOG_FILE
--- a/bigtop-packages/src/common/ambari/patch5-refine-AMBARI-25599-for-Hadoop-3.4.diff
+++ b/bigtop-packages/src/common/ambari/patch5-refine-AMBARI-25599-for-Hadoop-3.4.diff
@@ -43,7 +43,7 @@ index f8423b3806..e37072e9f8 100644
                    </target>
                  </configuration>
 diff --git a/ambari-metrics/pom.xml b/ambari-metrics/pom.xml
-index a0a11b8e6e..6770d86080 100644
+index a0a11b8e6e..909a297c1c 100644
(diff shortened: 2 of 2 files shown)


guyuqi:
This PR is dependent on https://github.com/apache/bigtop/pull/1056, https://github.com/apache/bigtop/pull/1053.
TODO: Put the related Binary TARs and JARs on Bigtop Public Repo and update the url in `ambari-metrics/pom.xml`

guyuqi:
Update url link for downloading jars/tars to S3 buckets.
Combining with https://github.com/apache/bigtop/pull/1056, it would sucessfully deploy Bigtop-3.2(Hadoop-3.3.4) + Mpack + Ambari metrics:

guyuqi:
Also successfully Build ambari on centos7/rocky8/Debian-11 and smoke tests were passed.
```
./docker-hadoop.sh --create 3 \
--image bigtop/puppet:trunk-rockylinux-8 --memory 16g \
--repo  file:///bigtop-home/output --disable-gpg-check --enable-local-repo \
--stack ambari,bigtop-utils  \
--smoke-tests ambari
```

```
Successfully started process 'Gradle Test Executor 2'

Gradle Test Executor 2 started executing tests.

> Task :bigtop-tests:smoke-tests:ambari:test

org.apache.bigtop.itest.ambari.TestAmbariSmoke > testBlueprints STANDARD_OUT
    [href:http://localhost:8080/api/v1/blueprints, items:[]]

org.apache.bigtop.itest.ambari.TestAmbariSmoke > testStackNameVersion STANDARD_OUT
    [versions:[[Versions:[stack_version:2.0, stack_name:ODPi], href:http://localhost:8080/api/v1/stacks/ODPi/versions/2.0]], Stacks:[stack_name:ODPi], href:http://localhost:8080/api/v1/stacks/ODPi]

Gradle Test Executor 2 finished executing tests.

> Task :bigtop-tests:smoke-tests:ambari:test
Finished generating test XML results (0.013 secs) into: /bigtop-home/bigtop-tests/smoke-tests/ambari/build/test-results/test
Finished generating test html results (0.032 secs) into: /bigtop-home/bigtop-tests/smoke-tests/ambari/build/reports/tests/test
Now testing...
:bigtop-tests:smoke-tests:ambari:test (Thread[Execution worker for ':' Thread 7,5,main]) completed. Took 4.428 secs.

BUILD SUCCESSFUL in 1m 35s
30 actionable tasks: 7 executed, 23 up-to-date
```

guyuqi:
> This PR is dependent on #1056, #1053. TODO: Put the related Binary TARs and JARs on Bigtop Public Repo and update the url in `ambari-metrics/pom.xml`

The link was updated to s3 buckets.

iwasakims:
@guyuqi  [The issue you raised](https://github.com/apache/bigtop/pull/1053#discussion_r1031008728) was not related to embedded Hadoop and HBase. Do you still need Bigtop tarball?

iwasakims:
@guyuqi How about cherry-picking [AMBARI-25780](https://github.com/apache/ambari-metrics/pull/77) instead of forking Ambari?

guyuqi:
> @guyuqi How about cherry-picking [AMBARI-25780](https://github.com/apache/ambari-metrics/pull/77) instead of forking Ambari?

1.  The [issue](https://github.com/apache/bigtop/pull/1053#discussion_r1031008728) I mentioned is irrelevant , so I closed the PR #1053. 

2. It doesn't work by just only cherry-picking https://github.com/apache/ambari-metrics/pull/77. There are other related PRs : AMBARI-25717, AMBARI-25668, AMBARI-25599, AMBARI-25349, AMBARI-25405.....

3. The Ambari metrics is a dependent project, not in Ambari project after Ambari-2.8. Ambari guys are still working on Ambari metrics for Ambari-2.8 (not done).

4. Acctually this PR is for Ambari-2.7.5 + Mpack + Hadoop-3.3.4, I don't think we should take more efforts on cherry-picking those PRs which are based on Ambari-2.8. And Ambari guys  still didn't perform any tests on Bigtop stack + Ambari-2.8 deployment.
IMHO, it makes sense to include this minimum modification for Ambari-2.7.5 and spend time on 3.2 release work.

Thanks. 
@iwasakims

iwasakims:
> It doesn't work by just only cherry-picking AMBARI-25780 Ambari metrics should upgrade hadoop and hbase for known CVE issues ambari-metrics#77. There are other related PRs : AMBARI-25717, AMBARI-25668, AMBARI-25599, AMBARI-25349, AMBARI-25405.....

AMBARI-25349 and AMBARI-25405 are already resolved in 2.7.5. You don't need AMBARI-25599 if you resolve conflicts on back-porting AMBARI-25780. While you need backport of AMBARI-25717 and AMBARI-25668, those are small patch. Using Bigtop top tarball is not solution for these.

Anyway it is not sure that JIRAs listed here are sufficient to migrate ambari-metrics to upstream Hadoop/HBase as you mentioned. I appreciate your effor moving the next Bigtop relase forward. @guyuqi

iwasakims:
@guyuqi I reverted this. You should not merge PR with explict +1.

bigtop-packages/src/common/ambari/patch12-fix-ambari-metrics-grafana-service-issues.diff
```diff
@@ -0,0 +1,44 @@
...
++      sleep 120
++      #until { cat "$PID_FILE" | xargs kill -0; } >/dev/null 2>&1
++      #do
++      #  sleep 1
++      #  i=$(($i + 1))
++      #  if [ $i -gt $timeout ]; then
++      #    echo "FAILED"
++      #    exit 1
++      #  fi
++      #done
```
iwasakims, line 41: You should add comment explaining why this should be commented out.

PR #1062, opened 2022-11-30
BIGTOP-3877: Add support for hadoop component

MacChen02:
### Description of PR
Add support for hadoop component, the point modifed is:
      1. redhat-rpm-config should be instead of openEuler-rpm-config in openEuler.
      2. The node-gyp 3.8.0 only support python2, but openEuler using the python3.

### For code changes:

- [ ] Does the title or this PR starts with the corresponding JIRA issue id (BIGTOP-3877)


guyuqi:
Duplicated to https://github.com/apache/bigtop/pull/1128

PR #1069, opened 2022-12-05
BIGTOP-3889: Update S3 bucket links of Jars downloading for Metrics in Mpack

guyuqi:
https://issues.apache.org/jira/browse/BIGTOP-3889

For Metrics in Mpack,  the `bigtop-snapshot` bucket would be refreshed every new build. Then the Jars would be cleaned in the [related path](http://bigtop-snapshot.s3.amazonaws.com/centos-7/x86_64/bigtop-stack-binary/) and files would be not found.

Update the link to the bucket: `repo.bigtop.apache.org `which would be permanently reserved for 3.2.0.


JunHe77:
LGTM. +1

PR #1086, opened 2023-02-07
BIGTOP-3907: Fix Spark History Server fails to start on aarch64

kevinw66:
### For code changes:

### DIFF
--- a/bigtop-packages/src/common/spark/do-component-build
+++ b/bigtop-packages/src/common/spark/do-component-build
@@ -18,6 +18,68 @@ set -ex
 
 . `dirname $0`/bigtop.bom
 
+# BIGTOP-3907: Build native library/jars for aarch64
+if [ $HOSTTYPE = "aarch64" ] ; then
+        # download leveldbjni 1.8 from original repo
+        git clone https://github.com/chirino/leveldb.git
+        git clone https://github.com/fusesource/leveldbjni.git --branch leveldbjni-1.8 --single-branch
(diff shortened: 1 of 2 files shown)


bigtop-packages/src/common/spark/do-component-build
```diff
@@ -18,6 +18,68 @@ set -ex
...
+        # aarch64 version of leveldbjni.
+        cd ${LEVELDBJNI_HOME}
+        mvn clean install -DskipTests -P download -Plinux64,all
+        cd ..
+        #cleanup
+        rm -rf ${LEVELDBJNI_HOME}
+        rm -rf ${LEVELDB_HOME}
+fi
+## BIGTOP-3907
+
```
kevinw66, line 82: These can be removed if we already built Hadoop

guyuqi: This issue was brought about by the `leveldbjni `dependency as you mentioned.
You mean if we build Hadoop first on the same node, it appears there are no leveldbjni dependency issues with Spark, right?

kevinw66: Yes, but we still need to delete profile `aarch64` for Spark

pan3793:
AFAIK, Spark is moving to rocksdb in recently version, and leveldb is not a hard dependency, instead of a workaround for leveldb, how do you think switching to rocksdb by change the default configurations?

PR #1087, opened 2023-02-12
BIGTOP-3908 Upgrade Spark Rpm Packages for PySpark Requres Python3

vivostar:
### Description of PR
Upgade spark rpm packages for pyspark requires python3, according to the [doc](https://spark.apache.org/docs/3.2.3/api/python/getting_started/quickstart_df.html).

```bash
./docker-hadoop.sh \
       -d \
       -dcp\
       --create 1 \
       --memory 8g \
       -L \
       --disable-gpg-check \
       --stack hdfs,yarn,mapreduce,spark,hive
```

### DIFF
--- a/bigtop-packages/src/rpm/spark/SPECS/spark.spec
+++ b/bigtop-packages/src/rpm/spark/SPECS/spark.spec
@@ -116,11 +116,7 @@ Server for Spark worker
 %package -n %{spark_pkg_name}-python
 Summary: Python client for Spark
 Group: Development/Libraries
-%if 0%{?rhel} >= 8
-Requires: %{spark_pkg_name}-core = %{version}-%{release}, python2
-%else
-Requires: %{spark_pkg_name}-core = %{version}-%{release}, python
-%endif
+Requires: %{spark_pkg_name}-core = %{version}-%{release}, python36
 
 %description -n %{spark_pkg_name}-python
 Includes PySpark, an interactive Python shell for Spark, and related libraries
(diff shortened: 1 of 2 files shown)


bigtop-packages/src/rpm/spark/SPECS/spark.spec
```diff
@@ -116,11 +116,7 @@ Server for Spark worker
 %package -n %{spark_pkg_name}-python
 Summary: Python client for Spark
 Group: Development/Libraries
-%if 0%{?rhel} >= 8
-Requires: %{spark_pkg_name}-core = %{version}-%{release}, python2
-%else
-Requires: %{spark_pkg_name}-core = %{version}-%{release}, python
-%endif
+Requires: %{spark_pkg_name}-core = %{version}-%{release}, python36
```
pan3793, line 119: Currently, bigtop uses Spark 3.2, but since Spark 3.3, the minimal python version is 3.7, so can we use a higher version to make it more future-proofing?
https://spark.apache.org/docs/3.3.2/api/python/getting_started/install.html

vivostar: yes, we can.

kevinw66:
PySpark stucked in CentOS7.

```
[root@d84cd50ac3f6 /]# pyspark
Python 3.6.8 (default, Nov 16 2020, 16:55:22) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)] on linux
Type "help", "copyright", "credits" or "license" for more information.
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/usr/lib/spark/jars/slf4j-log4j12-1.7.30.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/usr/lib/hadoop/lib/slf4j-reload4j-1.7.36.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Log4jLoggerFactory]
Setting default log level to "WARN".
To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
23/06/15 13:34:41 WARN Utils: Service 'SparkUI' could not bind on port 4040. Attempting port 4041.
23/06/15 13:34:42 WARN Client: Neither spark.yarn.jars nor spark.yarn.archive is set, falling back to uploading libraries under SPARK_HOME.
### stucked here
```

BTW, could you also upgrade python for Spark deb packages?

PR #1090, opened 2023-03-09
BIGTOP-3913:The path information is incorrect in the zookeeper-client script file by building with -PparentDir

guluo2016:
### Description of PR

fix bug, see:  [BIGTOP-3913](https://issues.apache.org/jira/browse/BIGTOP-3913)

Building package with -PparentDir=/usr/bigtop for zk, 
and installing zookeeper, 
and testing by executing zookeeper-client,zookeeper-server,zookeeper-server-cleanup,zookeeper-server-initialize

### For code changes:
install_zookeeper.sh

Please review @iwasakims thanks

### DIFF
--- a/bigtop-packages/src/common/zookeeper/install_zookeeper.sh
+++ b/bigtop-packages/src/common/zookeeper/install_zookeeper.sh
@@ -170,11 +170,11 @@ cat > $wrapper <<EOF
 # Autodetect JAVA_HOME if not defined
 . /usr/lib/bigtop-utils/bigtop-detect-javahome
 
-export ZOOKEEPER_HOME=\${ZOOKEEPER_CONF:-/usr/lib/zookeeper}
+export ZOOKEEPER_HOME=\${ZOOKEEPER_CONF:-${LIB_DIR}}
 export ZOOKEEPER_CONF=\${ZOOKEEPER_CONF:-/etc/zookeeper/conf}
 export CLASSPATH=\$CLASSPATH:\$ZOOKEEPER_CONF:\$ZOOKEEPER_HOME/*:\$ZOOKEEPER_HOME/lib/*
 export ZOOCFGDIR=\${ZOOCFGDIR:-\$ZOOKEEPER_CONF}
-env CLASSPATH=\$CLASSPATH /usr/lib/zookeeper/bin/zkCli.sh "\$@"
+env CLASSPATH=\$CLASSPATH \${ZOOKEEPER_HOME}/bin/zkCli.sh "\$@"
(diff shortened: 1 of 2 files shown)


iwasakims:
@guluo2016 Could you increment [the release number of the package](https://github.com/apache/bigtop/blob/dc4fe2d6655b3b3f1257f947203fd851a51c1b08/bigtop.bom#L138) to 3? Already released Bigtop 3.2.0 [bundles zookeeper as zookeeper-3.5.9-2](https://issues.apache.org/jira/browse/BIGTOP-3724).

guluo2016:
>

guluo2016:
@iwasakims   Increaseing [the release number of zookeeper](https://github.com/apache/bigtop/blob/dc4fe2d6655b3b3f1257f947203fd851a51c1b08/bigtop.bom#L138) to 3. Does this mean that other components also need to be updated? For example, hadoop, hive, etc.
Because [the release number of hadoop](https://github.com/apache/bigtop/blob/master/bigtop.bom#L154) is not 3

iwasakims:
@guluo2016
> Does this mean that other components also need to be updated? For example, hadoop, hive, etc.

No.

We need to increment the release number if we modify package contents without bumping the version of the product. The version of ZooKeeper packaged by Bigtop 3.2.0 is 3.5.9-2. If we release Bigtop 3.2.1 now, the version will be 3.5.9-3. It implies that the package contents was changed within zookeeper-3.5.9 and we can upgrade installed package by `rpm -U zookeeper` or `yum update zookeeper` (for RPM as example).

We don't need increment the release number of ZooKeeper again before releasing Bigtop 3.2.1 (or Biptop 3.3.0). We can reset the release number to 1 if we bump the ZooKeeper to 3.5.10 or later.

guluo2016:
> @guluo2016
> 
> > Does this mean that other components also need to be updated? For example, hadoop, hive, etc.
> 
> No.
> 
> We need to increment the release number if we modify package contents without bumping the version of the product. The package version of ZooKeeper packaged by Bigtop 3.2.0 is 3.5.9-2. If we release Bigtop 3.2.1 now, the version will be 3.5.9-3. It implies that the package contents was changed within zookeeper-3.5.9 and we can upgrade installed package by `rpm -U zookeeper` or `yum update zookeeper` (for RPM as example).
> 
> We don't need increment the release number of ZooKeeper again before releasing Bigtop 3.2.1 (or Biptop 3.3.0). We can reset the release number to 1 if we bump the ZooKeeper to 3.5.10 or later.

Understanded!  thank you

PR #1091, opened 2023-03-14
BIGTOP-3897: Failed to deploy gpdb on Fedora-36 for arm64

guyuqi:
https://issues.apache.org/jira/browse/BIGTOP-3897

The root cause: Python's cryptography package was upgraded to python3 and a Arm gcc error would occur when attempting to install the cryptography package by puppet.
The patch is to remove unnecessary package installation in puppet except for CentOS. GPDB smoke tests have been tested in Fedora-36, CentOS-7, Rockylinux-8, Debian-10/11 and Ubuntu-20.04.


guyuqi:
I confirmed that we could successfully build GPDB Debian/RPM packages and the somke tests in in Fedora-36, CentOS-7, Rockylinux-8, Debian-10/11 and Ubuntu-20.04 were passed.

PR #1107, opened 2023-04-20
BIGTOP-3931: fix kafka directory permission

JiaLiangC:
### Description of PR
 fix kafka directory permission
details in https://issues.apache.org/jira/browse/BIGTOP-3931

Before the fix, the permission of /usr/bigtop/3.2.0/usr/lib/kafka and its subdirectories was set to kafka, instead of root. This caused a permission error when trying to install Ranger Kafka plugin.
image

image

After the fix, the permission of /usr/bigtop/3.2.0/usr/lib/kafka and its subdirectories were restored to root, and Ranger Kafka plugin could be installed without issues.
image
image

### For code changes:

### DIFF
--- a/bigtop-packages/src/rpm/kafka/SPECS/kafka.spec
+++ b/bigtop-packages/src/rpm/kafka/SPECS/kafka.spec
@@ -180,12 +180,11 @@ fi
 
 %files
 %defattr(-,root,root,755)
+%{usr_lib_kafka}
 %{bin_dir}/*
 %config(noreplace) %{etc_kafka_conf_dist}
 %config(noreplace) %{etc_default}/kafka
 %attr(0755,kafka,kafka) %{np_etc_kafka}
-%attr(0755,kafka,kafka) %{usr_lib_kafka}
-%attr(0755,kafka,kafka) %docdir %{doc_kafka}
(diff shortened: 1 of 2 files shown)


JiaLiangC:
@iwasakims Could you review this patch ?

guluo2016:
I think you need increase [the release number of kafka](https://github.com/apache/bigtop/blob/dc4fe2d6655b3b3f1257f947203fd851a51c1b08/bigtop.bom#L304) to 3 at the same time, because kafka has been updated by you.

You can refer to the explanation here
https://github.com/apache/bigtop/pull/1090#issuecomment-1465206469

JiaLiangC:
> I think you need increase [the release number of kafka](https://github.com/apache/bigtop/blob/dc4fe2d6655b3b3f1257f947203fd851a51c1b08/bigtop.bom#L304) to 3 at the same time, because kafka has been updated by you.
> 
> You can refer to the explanation here [#1090 (comment)](https://github.com/apache/bigtop/pull/1090#issuecomment-1465206469)

@guluo2016  Thank you for your note. The update has been made.

iwasakims (approved):
LGTM. Thanks, @JiaLiangC and @guluo2016.

PR #1109, opened 2023-05-11
BIGTOP-3919: Bump GPDB to 6.23.4

guyuqi:
https://issues.apache.org/jira/browse/BIGTOP-3919
Upgrade GPDB to 6.23.4.

1. Removed the obsoleted patches;
2. Suppressed 'implicit-fallthrough' warning on ubuntu; 
3. Declared python version explicitly to fix gpcheckcat issues of ambiguous python shebang.
4. Refactored puppet scripts.
5. Fixed smoke tests and rpath issues on Fedora.


guyuqi:
Tests:
1. Build gpdb on Fedora-36, Rockylinux-8, CentOS-7, Ubtuntu-20 and Debian-11 (Arm64/x86_64).
2. The related smoke tests are passed.

bigtop-packages/src/common/gpdb/do-component-configure
```diff
@@ -16,4 +16,13 @@
...
 set -ex
 
-./configure --prefix=$1 --disable-orca CFLAGS=-fcommon
+if [ -f /etc/os-release ]; then
+    . /etc/os-release
+fi
+
+OS="$ID"
+if [ "${OS}" = "fedora" ]; then
+  ./configure --prefix=$1 --with-python --with-libxml --with-gssapi --disable-orca --disable-rpath CFLAGS='-fcommon -Wno-implicit-fallthrough'
```
iwasakims, line 25: @guyuqi This looks not the right solution. We should not expect users to set LD_LIBRARY_PATH. If we really need to disable rpath, the RPM should add a configuration file under /etc/ld.so.conf.d/ and run ldconfig on installing.

How about setting `QA_RPATHS=$(( 0x0001|0x0002 ))` on rpmbuild as the message suggest?

```
+ /usr/lib/rpm/check-rpaths
*******************************************************************************
*
* WARNING: 'check-rpaths' detected a broken RPATH OR RUNPATH and will cause
*          'rpmbuild' to fail. To ignore these errors, you can set the
*          '$QA_RPATHS' environment variable which is a bitmask allowing the
*          values below. The current value of QA_RPATHS is 0x0002.
*
*    0x0001 ... standard RPATHs (e.g. /usr/lib); such RPATHs are a minor
*               issue but are introducing redundant searchpaths without
*               providing a benefit. They can also cause errors in multilib
*               environments.
*    0x0002 ... invalid RPATHs; these are RPATHs which are neither absolute
*               nor relative filenames and can therefore be a SECURITY risk
*    0x0004 ... insecure RPATHs; these are relative RPATHs which are a
*               SECURITY risk
*    0x0008 ... the special '$ORIGIN' RPATHs are appearing after other
*               RPATHs; this is just a minor issue but usually unwanted
*    0x0010 ... the RPATH is empty; there is no reason for such RPATHs
*               and they cause unneeded work while loading libraries
*    0x0020 ... an RPATH references '..' of an absolute path; this will break
*               the functionality when the path before '..' is a symlink
*
*
* Examples:
* - to ignore standard and empty RPATHs, execute 'rpmbuild' like
*   $ QA_RPATHS=$(( 0x0001|0x0010 )) rpmbuild my-package.src.rpm
* - to check existing files, set $RPM_BUILD_ROOT and execute check-rpaths like
*   $ RPM_BUILD_ROOT=<top-dir> /usr/lib/rpm/check-rpaths
*
*******************************************************************************
ERROR   0001: file '/usr/lib/gpdb/lib/postgresql/plpython2.so' contains a standard  '/usr/lib64' in [/usr/lib64:/ext/python/lib]
WARNING 0002: file '/usr/lib/gpdb/lib/postgresql/plpython2.so' contains an invalid  '/ext/python/lib' in [/usr/lib64:/ext/python/lib]
error: Bad exit status from /var/tmp/rpm-tmp.4dIXaK (%install)
```

guyuqi: Thanks for the comments and Sorry for the late reply.

Agreed with your suggestion. We shoud not disable rpath check in Gpdb binary building for avoiding unnecessary  `LD_LIBRARY_PATH` setting after users install gpdb rpm.

`QA_RPATHS=$(( 0x0001|0x0002 ))` is to skip the errors of rpath check.

From https://fedoraproject.org/wiki/Changes/Broken_RPATH_will_fail_rpmbuild, what about setting Broken RPATH to null by `%global __brp_check_rpaths %{nil}`  which also performs as skipping  the errors of rpath check.
I also remove unnecessary `LD_LIBRARY_PATH` setting in puppet scripts as shown in 5th commit.  :-)

guyuqi: The building and smoke tests are passed on x86_64 and Arm64.

PR #1116, opened 2023-06-12
BIGTOP-3946. Fix build failure of Ambari due to expired certificate of bower repo.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3946

While upgrading bower should be the right solution, it is not yet done in Ambari upstream. I just added the patch of [AMBARI-25946](https://issues.apache.org/jira/browse/AMBARI-25946).


guyuqi:
LGTM, +1

PR #1121, opened 2023-06-21
BIGTOP-3952. Fix toolchain to download Maven from archive.apache.org instead of dlcdn.apache.org.

sekikn:
### Description of PR

This PR changes the download site of Maven and Ant from dlcdn.apache.org to archive.apache.org in case that old versions are removed from the former site.

By running `./gradlew toolchain` on CentOS 7.

### For code changes:


iwasakims (approved):
LGTM. Thanks, @sekikn.

PR #1160, opened 2023-08-04
BIGTOP-3981. Get rid of building patched leveldbjni in Hadoop packaging on aarch64.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-3981

Since [Hadoop is using org.openlabtesting.leveldbjni on aarh64](https://github.com/apache/hadoop/blob/rel/release-3.3.5/pom.xml#L724-L735) now, we can get rid of patching and building leveldbjni in do-component-build script on aarch64.

Attachments


guyuqi:
+1

guyuqi:
Smoke tests hava been verified on Arm64 Ubuntu-22.04.

guyuqi:
Merged.

PR #1162, opened 2023-08-10
BIGTOP-3985: Make a symlink of hadoop-yarn-server-timelineservice-hba…

JiaLiangC:
…se-coprocessor.jar to hadoop-yarn-server-timelineservice-hbase-coprocessor-version.jar

### Description of PR
When using YARN Timeline Service v2, you need to configure the following property in yarn-site:
yarn.timeline-service.hbase.coprocessor.jar.hdfs.location
hadoop-yarn-server-timelineservice-hbase-coprocessor-3.3.5.jar
This binding to version 3.3.5 can complicate automated configuration and installation. Similar to other components that depend on ZooKeeper's jar, symbolic links are often created to handle version dependencies, generating zookeeper.jar. Both Ambari HDP and Bigtop stacks utilize similar code for configuring this property, making it easier to automatically populate the configuration without needing to consider the specific Hadoop version.

details see https://issues.apache.org/jira/browse/BIGTOP-3985

manual test
after apply this patch and rebuild hadoop rpm, yarn ui2 show logs normally with ambari bigtop stack
image

### For code changes:

### DIFF
--- a/bigtop-packages/src/common/hadoop/install_hadoop.sh
+++ b/bigtop-packages/src/common/hadoop/install_hadoop.sh
@@ -397,6 +397,12 @@ for conf in conf.pseudo ; do
 done
 cp ${BUILD_DIR}/etc/hadoop/log4j.properties $PREFIX/$ETC_HADOOP/conf.pseudo
 
+# Make a symlink of hadoop-yarn-server-timelineservice-hbase-coprocessor.jar to hadoop-yarn-server-timelineservice-hbase-coprocessor-version.jar
+for x in $PREFIX/$YARN_DIR/timelineservice/hadoop-yarn-server-timelineservice-hbase-coprocessor-[[:digit:]]*.jar ; do
+  x=$(basename $x)
+  ln -s $x  $PREFIX/$YARN_DIR/timelineservice/hadoop-yarn-server-timelineservice-hbase-coprocessor.jar
+done
+
 # FIXME: Provide a convenience link for configuration (HADOOP-7939)
 install -d -m 0755 $PREFIX/$HADOOP_DIR/etc
 ln -s $NP_ETC_HADOOP/conf $PREFIX/$HADOOP_DIR/etc/hadoop


iwasakims:
``yarn.timeline-service.hbase.coprocessor.jar.hdfs.location`` implies that the value is a path on HDFS. Uploading the jar to HDFS as hadoop-yarn-server-timelineservice-hbase-coprocessor.jar should be enough?

JiaLiangC:
@iwasakims  Thank you for your suggestion.

The YARN Timeline Service v2 relies on HBase and has three deployment modes:

In the first mode, YARN starts an embedded HBase instance.
In the second mode, YARN uploads YARN files to the resource manager and then launches an HBase instance as a YARN service.
The third mode involves YARN utilizing an external HBase installation.
The property yarn.timeline-service.hbase.coprocessor.jar.hdfs.location isn't restricted to HDFS paths only. It can work with any path that HDFS can resolve, such as file:/// or s3:/. Its purpose is to make it convenient for the HBase JVM to load the JAR file.
yarn FlowRunTableRW.java
This coprocessor JAR is primarily used in the code for creating the FlowRunTable in YARN.
image

hbase HTableDescriptor.java
image
image


In Ambari, a local path is used, following this pattern: yarn_timeline_jar_location = format("file://{stack_root}/{version}/usr/lib/hadoop-yarn/timelineservice/{coprocessor_jar_name}"). This approach is the simplest; it avoids the need to upload the JAR to HDFS. While uploading is still an option, hardcoding the version number in the code can complicate future maintenance. This is why I introduced this symbolic link.

iwasakims (approved):
I think the change itself is harmless. You can proceed with this. I tested the patch on Ubuntu 22.04 x86_64.

> The property yarn.timeline-service.hbase.coprocessor.jar.hdfs.location isn't restricted to HDFS paths only. It can work with any path that HDFS can resolve, such as file:/// or s3:/. Its purpose is to make it convenient for the HBase JVM to load the JAR file.

Yes. `file://` scheme works for only embedded/co-located HBase. Ambari should encourage users to use `hdfs://` which can cover wider scenario?

> While uploading is still an option, hardcoding the version number in the code can complicate future maintenance.

You can specify the destination file name on HDFS upload. Removing the version string can be done on the upload.

JiaLiangC:
@iwasakims 
Thank you for helping with the review and testing. Is there any documentation on how to test in different environments? This way, when I submit a pull request, I will include specific test results.

I appreciate your suggestion. Uploading to HDFS indeed is a better choice. I will add this to the todo list.

iwasakims:
> Is there any documentation on how to test in different environments?

@JiaLiangC I think improving developers docs is our todo.

In short, you can leverage Docker by `*-pkg-ind` and `repo-ind` task.

```
$ ./gradlew hadoop-pkg-ind repo-ind -POS=ubuntu-22.04 -Pprefix=trunk -Dbuildwithdeps=true -Pdocker-run-option="--privileged" -Pmvn-cache-volume=true
```

* `-Dbuildwithdeps=true` kicks packging of products depended by hadoop (such as bigtop-utils and zookeeper).
* `-Pdocker-run-option="--privileged"` is needed on the Fedora-35 and Ubuntu-22.04 now (depending on the version of systemd).
* `-Pmvn-cache-volume=true` attaches docker volume to reuse local repository (~/.m2) to make repeatable build faster.

iwasakims:
You can deploy a cluster and run smoke-tests in container by docker provisioner which requires docker-compose.

```
$ cd provisioner/docker
$ ./docker-hadoop.sh \
    --create 3 \
    --image bigtop/puppet:trunk-ubuntu-22.04 \
    --docker-compose-yml docker-compose-cgroupv2.yml \
    --docker-compose-plugin \
    --memory 8g \
    --disable-gpg-check \
    --stack hdfs,yarn,mapreduce \
    --smoke-tests hdfs,yarn,mapreduce
```

* ``--docker-compose-yml docker-compose-cgroupv2.yml`` is needed on cgroup v2.
* ``--docker-compose-plugin`` is for using `docker compose` instead of `docker-compose`.
* use ``--repo file:///bigtop-home/output`` for RPM.

You can log in to the node and see files if you need.

```
$ ./docker-hadoop.sh --docker-compose-plugin --exec 1 /bin/bash
```

JiaLiangC:
Thanks for your detailed reply

PR #1163, opened 2023-08-14
BIGTOP-3983: Fix the Ranger-2.4 deploying issues

guyuqi:
https://issues.apache.org/jira/browse/BIGTOP-3983

Cherry pick the patch from the commit 2 of [RANGER-3206](https://issues.apache.org/jira/browse/RANGER-3206).
(https://github.com/apache/ranger/commit/f6b69ca5846de9a7dfced70ae8e403388ecb89fe.)

The smoke tests were passed on Arm64 Fedora-36.
```
./docker-hadoop.sh -d --create 3 --image bigtop/puppet:3.2.0-fedora-36 --memory 16g \
--repo file:///bigtop-home/output --disable-gpg-check --enable-local-repo \
--stack bigtop-utils,ranger  \
--smoke-tests ranger 
```
```
...
..
.
> Task :bigtop-tests:smoke-tests:ranger:test
Finished generating test XML results (0.026 secs) into: /bigtop-home/bigtop-tests/smoke-tests/ranger/build/test-results/test
Finished generating test html results (0.058 secs) into: /bigtop-home/bigtop-tests/smoke-tests/ranger/build/reports/tests/test
Now testing...
:bigtop-tests:smoke-tests:ranger:test (Thread[Execution worker for ':' Thread 8,5,main]) completed. Took 4.962 secs.

BUILD SUCCESSFUL in 53s

```

### DIFF
--- a/bigtop-packages/src/common/ranger/patch3-RANGER-3206-commit-2.diff
+++ b/bigtop-packages/src/common/ranger/patch3-RANGER-3206-commit-2.diff
@@ -1,0 +1,13 @@
+diff --git a/security-admin/scripts/db_setup.py b/security-admin/scripts/db_setup.py
+index 10acb5375..24502f4fb 100644
+--- a/security-admin/scripts/db_setup.py
++++ b/security-admin/scripts/db_setup.py
+@@ -149,7 +149,7 @@ def dbversionBasedOnUserName(userName):
+ def set_env_val(command):
+ 	proc = subprocess.Popen(command, stdout = subprocess.PIPE)
+ 	for line in proc.stdout:
+-		(key, _, value) = line.partition("=")
++		(key, _, value) = line.decode('utf8').partition('=')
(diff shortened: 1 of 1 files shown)


JiaLiangC:
@guyuqi 

I am doing an upgrade to Python3 for Ambari and its related components, and I have encountered this problem.
 
Here, the issue is actually because all Python code for Ranger is executed in Python2. Therefore, if the Ranger db_setup.py script was previously run using Python2 in the system, it would not throw an error. 
This is because `subprocess.Popen(command, stdout = subprocess.PIPE)` returns a str in Python2, but returns bytes in Python3.

So, if you use a Python3 script to execute db_setup.py, modifying it to `(key, _, value) = line.decode('utf8').partition('=')` is good. However, there are many places where similar subprocess.Popen code exists, and all have encoding issues. 
All Python scripts for Ranger are based on Python2, so upgrading them all would be a better approach.

PR #1170, opened 2023-08-29
fix install groovy unzip bug

jasonchrion:
Fix install groovy unzip bug


guyuqi:
Thanks for working on it. 
Could you please file a Jira ticket for Bigtop first? Then, rename the PR title to `BIGTOP-xxxx: Fix ....`.
It would be appreciated if you could provide instructions on how to reproduce the issue on the relevant Jira.
@jasonchrion

PR #1177, opened 2023-09-11
BIGTOP-4005: Bump Kafka version to 2.8.2

timyuer:
### Description of PR
Remove [patch1-PR11554-set-version-of-jgit.diff](https://github.com/apache/bigtop/compare/master...timyuer:BIGTOP-4005?expand=1#diff-bada2f9e5f713102f78a0fac0e6a76cf834ab05472b6a910b717afb84c4b32b8)
as https://github.com/apache/kafka/pull/11561 said.

Successfully tested on Centos7 (x86_64).

```bash
./docker-hadoop.sh \
    --create 1 \
    --image bigtop/puppet:trunk-centos-7 \
    --memory 8g \
    --enable-local-repo \
    --disable-gpg-check \
    --stack bigtop-utils,zookeeper,kafka \
    --smoke-tests kafka
```

### For code changes:


bigtop.bom
```diff
@@ -301,7 +301,7 @@ bigtop {
       name    = 'kafka'
       rpm_pkg_suffix = "_" + bigtop.base_version.replace(".", "_")
       relNotes = 'Apache Kafka'
-      version { base = '2.8.1'; pkg = base; release = 3 }
+      version { base = '2.8.2'; pkg = base; release = 4 }
```
iwasakims, line 304: @timyuer We should reset the release number to 1 if we bump the base version.

timyuer: > @timyuer We should reset the release number to 1 if we bump the base version.

Thank you for your reminder. I have updated the release version.

iwasakims (approved):
+1. smoke-tests passed on Ubuntu 22.04 aarch64 too.

PR #1214, opened 2023-12-15
BIGTOP-4046: Enable Parallel Compilation for Multiple Modules in SPARK

JiaLiangC:
### Description of PR
Spark 3.3 itself does support parallel compilation. However, Spark 3.3 uses Maven Shade Plugin version 3.2.4, which triggers an infinite loop bug during parallel compilation. You can find more details about this issue at [MSHADE-413](https://issues.apache.org/jira/browse/MSHADE-413).

A PR has been submitted for Spark 3.3 to upgrade the Maven Shade Plugin to version 3.5.0. This upgrade has already been applied in Spark 3.5,  [SPARK-44257](https://issues.apache.org/jira/browse/SPARK-44257).

Therefore, this PR primarily introduces the Maven patch and also includes the upgrade of the mvn-shade-plugin as part of [SPARK-44257](https://issues.apache.org/jira/browse/SPARK-44257). It's worth noting that [SPARK-44257](https://issues.apache.org/jira/browse/SPARK-44257) contains some code that doesn't exist in Spark 3.3 and has been ignored.

**depend on [BIGTOP-4044](https://issues.apache.org/jira/browse/BIGTOP-4044). Therefore, when testing [BIGTOP-4046](https://issues.apache.org/jira/browse/BIGTOP-4046) , it's necessary to include [BIGTOP-4044]**(https://issues.apache.org/jira/browse/BIGTOP-4044) in the merge as well.
manual test
image
./gradlew spark-clean spark-pkg -PcompileThreads=2C
### For code changes:

### DIFF
--- a/bigtop-packages/src/common/spark/patch0-SPARK-44257.diff
+++ b/bigtop-packages/src/common/spark/patch0-SPARK-44257.diff
@@ -1,0 +1,22 @@
+diff --git a/pom.xml b/pom.xml
+index 7189fd2e093..fa4a8c1c987 100644
+--- a/pom.xml
++++ b/pom.xml
+@@ -3023,7 +3023,7 @@
+         <plugin>
+           <groupId>org.apache.maven.plugins</groupId>
+           <artifactId>maven-clean-plugin</artifactId>
+-          <version>3.1.0</version>
++          <version>3.3.1</version>
+           <configuration>
+             <filesets>
+               <fileset>
+@@ -3103,7 +3103,7 @@
+         <plugin>
+           <groupId>org.apache.maven.plugins</groupId>
+           <artifactId>maven-shade-plugin</artifactId>
+-          <version>3.2.4</version>
++          <version>3.5.0</version>
+           <dependencies>
+             <dependency>
+               <groupId>org.ow2.asm</groupId>
--- a/bigtop.bom
+++ b/bigtop.bom
@@ -219,6 +219,7 @@ bigtop {
       url     { download_path = "/$name/$name-${version.base}"
                 site = "${apache.APACHE_MIRROR}/${download_path}"
                 archive = "${apache.APACHE_ARCHIVE}/${download_path}" }
+      maven_parallel_build = true
     }
 
     'flink' {


iwasakims:
@JiaLiangC I got following error.

```                                                                                                                                                                                                                                                                                                         
$ ./gradlew spark-pkg                                                                                                                                                                                                                                                                                       
...                                                                                                                                                                                                                                                                                                         
++ mvn help:evaluate -Dexpression=project.version -Divy.home=/home/iwasakims/.ivy2 -Dsbt.ivy.home=/home/iwasakims/.ivy2 -Duser.home=/home/iwasakims -Drepo.maven.org= -Dreactor.repo=file:///home/iwasakims/.m2/repository -Dhadoop.version=3.3.6 -Dkafka.version=2.8.2 -Dzookeeper.version=3.7.2 -Dyarn.ve\
rsion=3.3.6 -Pyarn -Phadoop-3 -Phive -Phive-thriftserver -Psparkr -Pkubernetes -Pscala-2.12 -Dguava.version=27.0-jre -DskipTests                                                                                                                                                                            
++ grep -v INFO                                                                                                                                                                                                                                                                                             
++ grep -v WARNING                                                                                                                                                                                                                                                                                          
++ tail -n 1                                                                                                                                                                                                                                                                                                
+ VERSION='[ERROR] Invalid threads value: '\''@CORES@'\''. Supported are int values.'                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                            
```

bigtop-packages/src/common/spark/patch-maven-parallel-build.diff
```diff
@@ -0,0 +1,10 @@
+diff --git a/.mvn/maven.config b/.mvn/maven.config
+new file mode 100644
+index 00000000..e69de29b
+diff --git a/.mvn/maven.config b/.mvn/maven.config
+index e69de29b..0bea21c9 100644
+--- a/.mvn/maven.config
++++ b/.mvn/maven.config
+@@ -0,0 +1 @@
++-T@CORES@
+\ No newline at end of file
```
iwasakims, line 10: Usually patches are prefixed with numbers (`patch[0-9]+`) as done for following patch0-SPARK-44257.diff. If it is needed to be applied before all such patches, this one should follow the convention too.

JiaLiangC:
@iwasakims 
I'm sorry for not providing a clear PR description initially. Both BIGTOP-4046 and BIGTOP-4045 are based on BIGTOP-4044, and they both depend on BIGTOP-4044. Therefore, when testing BIGTOP-4046 and BIGTOP-4045, it's necessary to include BIGTOP-4044 in the merge as well.

Regarding the patch numbering convention you mentioned, I've also noticed it. However, what has been confusing is that this is a shared patch. If we add a patch number, it means that the names of this patch will be different for all components because the patch numbers will vary.

I have also considered another approach, which involves placing this patch in bigtop-packages/src/templates and then copying this patch to the respective build directory after adding the corresponding compilation options. However, there are two drawbacks to this approach:

It would require an additional property to control whether the patch needs to be copied to this project because some projects are not Java projects, and some projects do not yet support parallel compilation. This would mean having two properties to control parallel compilation, adding complexity and potentially reducing usability.
Considering this, it seems that adding a patch number might be a better choice compared to the second approach.

JiaLiangC:
@iwasakims 
I have grouped the parallel compilation PRs as subtasks under the BIGTOP-4044 issue for better progress management. I will update the patches for BIGTOP-4044, BIGTOP-4045, and BIGTOP-4046 to include patch numbers. Thank you for your suggestion.

PR #1216, opened 2023-12-19
BIGTOP-4051: Add python3 dependency for Phoenix deb

JiaLiangC:
### Description of PR
In the minimal Python code of phoenix 5.1.3, there is a handling for Python that allows it to support both Python 3 and Python 2. Therefore, a dependency on Python 3 is added in the Phoenix deb package because many systems only have Python 3 by default. Not adding this dependency would result in users being unable to install Phoenix properly.

The code related to Python 3 compatibility can be found at:
https://github.com/apache/phoenix/blob/5.1.3/bin/daemon.py#L66

### For code changes:

### DIFF
--- a/bigtop-packages/src/deb/phoenix/control
+++ b/bigtop-packages/src/deb/phoenix/control
@@ -23,7 +23,7 @@ Homepage: http://phoenix.apache.org
 
 Package: phoenix
 Architecture: all
-Depends: bigtop-utils (>= 0.7),  python (>= 2.6) | python2 (>= 2.6)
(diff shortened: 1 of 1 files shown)


iwasakims:
@JiaLiangC How about removing dependency on Python for aligning with other activities in BIGTOP-4052? The manifests of bigtop-deploy must be fixed if you need tweak the environment for running smoke-tests.

JiaLiangC:
@iwasakims 
I agree with removing the Python dependency. I will close this PR.

PR #1220, opened 2023-12-23
BIGTOP-3978. Drop debian-10 support

vivostar:
### Description of PR

Drop debian-10 support

We can use following comand to test it
```shell
./gradlew bigtop-puppet -POS=debian-10 -Pprefix=3.0.0 bigtop-puppet
```
and it will be get a failure

Screenshot from 2023-12-23 23-32-21
that's what we expected.

### For code changes:


build.gradle
```diff
@@ -267,7 +267,7 @@ Properties:
   -Pmemory=[4g|8g|...]
   -Pnum_instances=[NUM_INSTANCES]
   -Pnexus=[NEXUS_URL] (NEXUS_URL is optional)
-  -POS=[centos-7|fedora-35|debian-10|ubuntu-18.04|opensuse-42.3|openeuler-22.03]
+  -POS=[centos-7|fedora-35|ubuntu-18.04|opensuse-42.3|openeuler-22.03]
```
iwasakims, line 270: Replacing it with `debian-11` could be better than removing it. Since the description is outdated and not exhaustive, we can fix it in dedicated PR later.

iwasakims (approved):
LGTM. Thanks, @vivostar.

build.gradle
vivostar: @iwasakims thanks for your advice, I will fix it

PR #1229, opened 2024-01-03
BIGTOP-3896. Failed to deploy Alluxio on Arm64/ppc64le Fedora-38

MacChen01:
https://issues.apache.org/jira/browse/BIGTOP-3896

The so of alluxio is not compiled by default, and uses the x86 version, so we need to compile libjnifuse*.so and copy it to the corresponding directory of alluxio when building adaptation

Modify the bigtop-release-3.2.0/bigtop-packages/src/common/alluxio/do-component-build file to adapt Arm64 architecture code for fedora OS.
Modify the bigtop-release-3.2.0/bigtop-packages/src/common/alluxio/ install_alluxio.sh file to copy the libjnifuse*.so file.

- [ ] Does the title or this PR starts with the corresponding JIRA issue id (BIGTOP-3896)

### DIFF
--- a/bigtop-packages/src/common/alluxio/do-component-build
+++ b/bigtop-packages/src/common/alluxio/do-component-build
@@ -45,7 +45,7 @@ if [ $HOSTTYPE = "powerpc64le" ] ; then
 
 else
   #need to manually compile the libjnifuse*.so in openEuler
-  if [ "${OS}" = "openEuler" ] ; then
+  if [ "${OS}" = "openEuler" ] || [ "${OS}" = "fedora" ]; then
     sed -i "s|<activeByDefault>false</activeByDefault>|<activeByDefault>true</activeByDefault>|g" integration/jnifuse/native/pom.xml
   fi
 
--- a/bigtop-packages/src/common/alluxio/do-component-build
+++ b/bigtop-packages/src/common/alluxio/do-component-build
@@ -44,7 +44,9 @@ if [ $HOSTTYPE = "powerpc64le" ] ; then
   mvn clean install -DskipTests -Dhadoop.version=${HADOOP_VERSION} -Dgrpc.version=1.28.0 -Dmaven.buildNumber.revisionOnScmFailure=v${ALLUXIO_VERSION} -Phadoop-3 -Pyarn "$@"
 
 else
-  #need to manually compile the libjnifuse*.so in openEuler
+  #need to manually compile the libjnifuse*.so in openEuler and fedora.
+  #The libjnifuse*.so files which alluxio need is not compiled by default, and uses the x86 ARCH version
+  #so we need to compile libjnifuse*.so and copy it to the corresponding directory of alluxio
   if [ "${OS}" = "openEuler" ] || [ "${OS}" = "fedora" ]; then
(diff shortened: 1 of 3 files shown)


iwasakims:
@MacChen01 [BIGTOP-4014](https://issues.apache.org/jira/browse/BIGTOP-4014) bumped the version of Fedora to 38. Is the issue still reproducible on Fedora 38 and the patch is applicable?

MacChen01:
> @MacChen01 [BIGTOP-4014](https://issues.apache.org/jira/browse/BIGTOP-4014) bumped the version of Fedora to 38. Is the issue still reproducible on Fedora 38 and the patch is applicable?

@iwasakims 

The issue can reproduced in Fedora 38,  and the patch compiled and smoke test in Fedora 38 too.
The compile log:
image
The smoke test log:
image

bigtop-packages/src/common/alluxio/do-component-build
```diff
@@ -45,7 +45,7 @@ if [ $HOSTTYPE = "powerpc64le" ] ; then
 
 else
   #need to manually compile the libjnifuse*.so in openEuler
```
iwasakims, line 47: The comment should be updated too. It would be nice if you can describe the reason why this is needed.

bigtop-packages/src/common/alluxio/install_alluxio.sh
```diff
@@ -131,7 +134,7 @@ cp -a integration/* $PREFIX/$LIB_DIR/integration
 cp integration/fuse/target/alluxio-integration-fuse-*-jar-with-dependencies.jar $PREFIX/$LIB_DIR/integration/fuse
 
 # replace the original libjnifuse*.so file with the manually compiled in openEuler
```
iwasakims, line 136: The comment should be updated too.

bigtop_toolchain/manifests/packages.pp
```diff
@@ -29,6 +29,8 @@
         "gcc",
         "gcc-c++",
         "fuse",
+        "fuse3",
+        "fuse3-devel",
```
iwasakims, line 33: While this affects all Red Hat clones, I'm ok here since fuse3 and fuse3-devel is available in CentOS 7 and Rocky Linux 8 at least.

iwasakims:
smoke-tests of alluxio passed on my side (using pkg-ind and docker provisioner on aarch64 host). LGTM pending trivial comments.

bigtop-packages/src/common/alluxio/do-component-build
MacChen01: @iwasakims Thanks, the comments has updated. https://github.com/apache/bigtop/pull/1229/commits/9f0695218f5a73315101dfac29beef07a80526f5

bigtop-packages/src/common/alluxio/install_alluxio.sh
MacChen01: @iwasakims Thanks, the comments has updated. https://github.com/apache/bigtop/commit/9f0695218f5a73315101dfac29beef07a80526f5

MacChen01:
> smoke-tests of alluxio passed on my side (using pkg-ind and docker provisioner on aarch64 host). LGTM pending trivial comments.

@iwasakims Thanks for you review and check it, the comments has updated.

iwasakims (approved):
LGTM. Thanks, @MacChen01.

PR #1235, opened 2024-02-15
BIGTOP-3991 add rockylinux 9 support

cwseys:
### Description of PR
This change allows the creation of Rocky Linux 9 packages.  (Presumably compatible with AlmaLinux 9, Centos 9, and RHEL 9 also.)

### How was this patch tested?
After patching packages for Centos 7, Rocky Linux 8, and Rocky Linux 9 were built in the following ways:
(container only builds)
./gradlew clean -POS=$DISTRO bigtop-groovy-pkg-ind bigtop-jsvc-pkg-ind bigtop-utils-pkg-ind zookeeper-pkg-ind hadoop-pkg-ind
(also on bare metal)
./gradlew clean -POS=$DISTRO hadoop-pkg

Building for other distros and task were not tested.  (Let me know if there is a command to test the full gamut.  I will be happy to test it.)

PR #1238, opened 2024-02-25
BIGTOP-4064. Fix Bigtop toolchain to work with Puppet 3.x.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4064

Fix Puppet manifest syntax unusable in Puppet 3 used by CentOS 7. (broken by https://github.com/apache/bigtop/pull/1235)

### DIFF
--- a/bigtop_toolchain/manifests/packages.pp
+++ b/bigtop_toolchain/manifests/packages.pp
@@ -70,7 +70,7 @@ class bigtop_toolchain::packages {
         "yasm"
       ]
 
-      if (/redhat|centos|rocky/ in downcase($operatingsystem) and Integer($operatingsystemmajrelease) >= 9) {
+      if ($osfamily == 'RedHat' and 0 <= versioncmp($operatingsystemmajrelease, '9')) {
         $pkgs = $_pkgs + ['cmake']
       } elsif ($operatingsystem == 'Fedora' or $operatingsystemmajrelease !~ /^[0-7]$/) {
         $pkgs = concat($_pkgs, ["python2-devel", "libtirpc-devel", "cmake"])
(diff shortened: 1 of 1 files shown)


bigtop_toolchain/manifests/packages.pp
```diff
@@ -70,7 +70,7 @@
         "yasm"
       ]
 
-      if (/redhat|centos|rocky/ in downcase($operatingsystem) and Integer($operatingsystemmajrelease) >= 9) {
+      if (0 == versioncmp($operatingsystemmajrelease, '9')) {
```
sekikn, line 73: `<=` may be better than `==` considering the future release of new version.

bigtop_toolchain/manifests/packages.pp
```diff
@@ -401,7 +401,7 @@
   # download python 2.7.14 for openEuler docker slaves
   # and RHEL9 based distros
   if $operatingsystem == 'openEuler'
-     or (/redhat|centos|rocky/ in $operatingsystem and Integer($operatingsystemmajrelease) >= 9) {
+     or ($osfamily == 'RedHat' and 0 == versioncmp($operatingsystemmajrelease, '9')) {
```
sekikn, line 404: Same as above

sekikn (changes requested):

bigtop_toolchain/manifests/packages.pp
iwasakims: I refactored the conditional a bit.

bigtop_toolchain/manifests/packages.pp
iwasakims: I updated the condition based on your suggestion. I needed additional guard for Fedora since its `osfamily` is `RedHat` and  `operatingsystemmajrelease` is `38` (or similar). python2-devel is still available on Fedora.

iwasakims: On openEuler, `osfamily` is `Openeuler`.

MacChen01:
I met the same problem using the master branch to create the docker slaves of centos 7.
image

sekikn:
+1, ensured that no syntactic error occurred when running `./gradlew toolchain` on CentOS 7, Rocky 8 and 9 and Fedora 38. Thanks @iwasakims.

PR #1239, opened 2024-02-26
BIGTOP-4065: Fixing BUG caused by the lack of preun logic in Flink spec

JiaLiangC:
### Description of PR

The Flink RPM lacks preun logic, causing the alternative symbolic link to persist even after uninstalling Flink.

For example, if Flink was initially installed under /usr/bigtop/3.2.0 and then installed again with a different version under /usr/bigtop/3.3.0, the alternative symbolic link /etc/alternatives/flink-conf -> /usr/bigtop/3.3.0/etc/flink/conf.dist will not be correctly created due to the old alternatives not being uninstalled.

Upon checking other components, they all include preun uninstallation logic, which is why they do not have this issue.

 
image

### For code changes:


JiaLiangC:
@iwasakims @sekikn @guyuqi 
Could you help review this PR?

iwasakims (approved):
LGTM. Thanks, @JiaLiangC.

PR #1243, opened 2024-03-05
BIGTOP-4068: fix the alluxio in rockylinux9 bug

MacChen01:
https://issues.apache.org/jira/browse/BIGTOP-4068

The bug is alluxio smoketest failed in rockylinux 9, the test cmd
     ./docker-hadoop.sh -d -c 3 -C config_rockylinux-9.yaml

 **redhat-lsb** should be installed by dnf when alluxio somketest in  rockylinux 9

- [ ] Does the title or this PR starts with the corresponding JIRA issue id (BIGTOP-4068)

### DIFF
--- a/bigtop_toolchain/bin/puppetize.sh
+++ b/bigtop_toolchain/bin/puppetize.sh
@@ -70,7 +70,7 @@ case ${ID}-${VERSION_ID} in
         # Enabling the PowerTools and EPEL repositories via Puppet doesn't seem to work in some cases.
         # As a workaround for that, enable the former here in advance of running the Puppet manifests.
         dnf config-manager --set-enabled crb devel
-        dnf install -y redhat-lsb
+        dnf install -y redhat-lsb initscripts
         ;;
     rhel-8*)
         rpm -Uvh https://yum.puppet.com/puppet5-release-el-8.noarch.rpm
--- a/bigtop_toolchain/bin/puppetize.sh
+++ b/bigtop_toolchain/bin/puppetize.sh
@@ -70,7 +70,6 @@ case ${ID}-${VERSION_ID} in
         # Enabling the PowerTools and EPEL repositories via Puppet doesn't seem to work in some cases.
         # As a workaround for that, enable the former here in advance of running the Puppet manifests.
         dnf config-manager --set-enabled crb devel
-        dnf install -y redhat-lsb initscripts
         ;;
     rhel-8*)
         rpm -Uvh https://yum.puppet.com/puppet5-release-el-8.noarch.rpm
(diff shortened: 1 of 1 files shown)


MacChen01:
The successful log is:
image

MacChen01:
@guyuqi @iwasakims @sekikn 
Please help review this code, thanks.

bigtop_toolchain/bin/puppetize.sh
```diff
@@ -69,7 +69,8 @@ case ${ID}-${VERSION_ID} in
         puppet module install puppetlabs-stdlib --version 4.12.0
         # Enabling the PowerTools and EPEL repositories via Puppet doesn't seem to work in some cases.
         # As a workaround for that, enable the former here in advance of running the Puppet manifests.
-        dnf config-manager --set-enabled crb
+        dnf config-manager --set-enabled crb devel
```
iwasakims, line 72: This should be done in the [Puppet manifest](https://github.com/apache/bigtop/blob/master/bigtop-deploy/puppet/modules/alluxio/manifests/init.pp)? We can not expect that the puppetize.sh is applied in all deployment.

bigtop_toolchain/bin/puppetize.sh
```diff
@@ -69,7 +69,8 @@ case ${ID}-${VERSION_ID} in
...
+        dnf install -y redhat-lsb initscripts
```
iwasakims, line 73: Package [dependency on initscripts](https://github.com/apache/bigtop/blob/36a3b44905e062e0b1c657dba6bc82dc84accbb4/bigtop-packages/src/rpm/alluxio/SPECS/alluxio.spec#L75) seems to be already exist. How about adding dependency on redhat-lsb too? We can not expect that the puppetize.sh is applied in all deployment.

uscxiexing: Thanks.  @iwasakims 
I run alluxio smoke test test, hadoop installation error,the log is :
image
1. hadoop installed need /lib/lsb/init-functions in bigto/puppet:trunk-rockylinux-9 ,But it didn't.
2. /lib/lsb/init-functions  should pass dnf install redhat-lsb and initscripts.
3. I think this is the base environment and should be installed in puppet.  If you don't want to installed it in puppet,can I move  provisioner/utils/smoke-test.sh ,and add rockylinux9 branch ?

uscxiexing: Thanks for the advice. @iwasakims 
And I will delete “_dnf install -y redhat-lsb initscripts_”。Only increase ”_dnf config-manager --set-enabled crb devel”。
it succeed too，hadoop can also start。
image
image

MacChen01: Thanks  @iwasakims @uscxiexing 
The  “dnf install -y redhat-lsb initscripts” code has deleted, it's work fine in my env.

MacChen01:
@iwasakims 
Please help check the code from your advice, thanks.

iwasakims (approved):
I'm merging this as temporary fix.

iwasakims:
Filed [BIGTOP-4080](https://issues.apache.org/jira/browse/BIGTOP-4080) as a follow-up.

PR #1148, opened 2023-07-28
BIGTOP-3967: add support for spark

MacChen01:
### Description of PR
Add support for spark component, include compile command and problem.
1. install R by dnf in openEuler
2. add pandoc by wget and install in openEuler

### How was this patch tested?
build command
docker run --rm -v pwd:/home/bigtop --workdir /home/bigtop bigtop/slaves:trunk-openeuler-22.03-aarch64 bash -c '. /etc/profile.d/bigtop.sh; ./gradlew spark-pkg'

smoke test command
./docker-hadoop.sh -C config_openeuler-22.03.yaml -c 3 -s

### For code changes:

- [ ] Does the title or this PR starts with the corresponding JIRA issue id (BIGTOP-3967)
- [ ] Make sure that newly added files do not have any licensing issues. When in doubt refer to https://www.apache.org/licenses/

PR #1251, opened 2024-04-09
BIGTOP-4082: Fix installation failure or R by bigtop_toolchain on openEuler

MacChen01:
The cmd are:
    gradlew  bigtop-puppet  -POS=openeuler-22.03-lts-sp1 -Pprefix=3.3.0
    gradlew  bigtop-slaves  -POS=openeuler-22.03-lts-sp1 -Pprefix=3.3.0

- [ ] Does the title or this PR starts with the corresponding JIRA issue id (BIGTOP-4082)

### DIFF
--- a/docker/bigtop-puppet/build.sh
+++ b/docker/bigtop-puppet/build.sh
@@ -61,7 +61,8 @@ case "${OS}-${VERSION}" in
     ;;
   openeuler-22.03*)
     OPENEULER_OS="${OS}/${OS}"
-    sed -i -e "s|${OS}:${VERSION}|$OPENEULER_OS:${VERSION}|" ./Dockerfile
+    # BIGTOP-4082 specify the openEuler OS version is : 22.03-lts-sp1
+    sed -i -e "s|${OS}:${VERSION}|$OPENEULER_OS:${VERSION}-lts-sp1|" ./Dockerfile
     ;;
   *)
     ;;
--- a/docker/bigtop-puppet/build.sh
+++ b/docker/bigtop-puppet/build.sh
@@ -29,7 +29,14 @@ fi
 
 PREFIX=$(echo "$1" | cut -d '-' -f 1)
 OS=$(echo "$1" | cut -d '-' -f 2)
+
+# BIGTOP-4082 get the openEuler OS version
+if [ "${OS}" == "openeuler" ];then
+VERSION=$(echo "$1" | cut -d '-' -f 3-5)
+else
 VERSION=$(echo "$1" | cut -d '-' -f 3)
+fi
+
 ARCH=$(uname -m)
 if [ "${ARCH}" != "x86_64" ];then
 ARCH="-${ARCH}"
@@ -61,8 +68,7 @@ case "${OS}-${VERSION}" in
     ;;
   openeuler-22.03*)
     OPENEULER_OS="${OS}/${OS}"
-    # BIGTOP-4082 specify the openEuler OS version is : 22.03-lts-sp1
-    sed -i -e "s|${OS}:${VERSION}|$OPENEULER_OS:${VERSION}-lts-sp1|" ./Dockerfile
+    sed -i -e "s|${OS}:${VERSION}|$OPENEULER_OS:${VERSION}|" ./Dockerfile
     ;;
   *)
     ;;
--- a/docker/bigtop-puppet/build.sh
+++ b/docker/bigtop-puppet/build.sh
@@ -29,14 +29,7 @@ fi
 
 PREFIX=$(echo "$1" | cut -d '-' -f 1)
 OS=$(echo "$1" | cut -d '-' -f 2)
-
-# BIGTOP-4082 get the openEuler OS version
-if [ "${OS}" == "openeuler" ];then
-VERSION=$(echo "$1" | cut -d '-' -f 3-5)
-else
-VERSION=$(echo "$1" | cut -d '-' -f 3)
-fi
-
+VERSION=$(echo "$1" | cut -d '-' -f 3-)
 ARCH=$(uname -m)
 if [ "${ARCH}" != "x86_64" ];then
 ARCH="-${ARCH}"
--- a/build.gradle
+++ b/build.gradle
@@ -267,7 +267,7 @@ Properties:
   -Pmemory=[4g|8g|...]
   -Pnum_instances=[NUM_INSTANCES]
   -Pnexus=[NEXUS_URL] (NEXUS_URL is optional)
-  -POS=[centos-7|fedora-35|ubuntu-18.04|opensuse-42.3|openeuler-22.03]
+  -POS=[centos-7|fedora-35|ubuntu-18.04|opensuse-42.3|openeuler-22.03-lts-sp1]
   -Pprefix=[trunk|1.2.1|1.2.0|1.1.0|...]
   -Prepository=[REPO_URL]
   -Prun_smoke_tests (run test components defined in config file)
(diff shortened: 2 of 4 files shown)


MacChen01:
@iwasakims @guyuqi 
please help review the code, thanks.

docker/bigtop-puppet/build.sh
```diff
@@ -61,7 +61,8 @@ case "${OS}-${VERSION}" in
     ;;
   openeuler-22.03*)
     OPENEULER_OS="${OS}/${OS}"
-    sed -i -e "s|${OS}:${VERSION}|$OPENEULER_OS:${VERSION}|" ./Dockerfile
+    # BIGTOP-4082 specify the openEuler OS version is : 22.03-lts-sp1
+    sed -i -e "s|${OS}:${VERSION}|$OPENEULER_OS:${VERSION}-lts-sp1|" ./Dockerfile
```
iwasakims, line 65: @MacChen01 How about setting the VERSION by `VERSION=$(echo "$1" | cut -d '-' -f 3-)` in line 32? We can avoid hardcoding the version into the script by that. I'm expecting we can specify the version like

```
$ gradlew bigtop-puppet -POS=openeuler-22.03-lts-sp1 -Pprefix=3.3.0`
```

iwasakims:
> gradlew bigtop-slaves -POS=openeuler-22.03 -Pprefix=3.3.0

We need to fix `docker/bigtop-slaves/build.sh` too?

docker/bigtop-puppet/build.sh
MacChen01: @iwasakims Thanks for your review.
The version of openEuler 22.03 has many different subversions, we chose the 22.03-lts-sp1 version for openeuler-22.03.
So we hardcoding the version into the script by that.

iwasakims: I don't like to update the code every time we change the version used (e.g. 22.03-lts-sp2 in the future).

iwasakims: In addition, we should publish the image as openeuler-22.03-lts-sp1 rather than openeuler-22.03 if the base image is different.

MacChen01: ok, i will modify the version follow your advice.

iwasakims:
I could not reproduce the build failure reported in BIGTOP-4082 today.

iwasakims:
@MacChen01 I could not reproduce the build failure reported on [BIGTOP-4082](https://issues.apache.org/jira/browse/BIGTOP-4082). Could you try the build after removing cached old images?

iwasakims:
@MacChen01 Sorry, I can reproduce the issue now. Please update the patch to make tags including `-` available.

iwasakims:
[provisioner/docker/config_openeuler-22.03.yaml](https://github.com/apache/bigtop/blob/0fbaaa90859a889dbd86b869895224fb7126f677/provisioner/docker/config_openeuler-22.03.yaml) should be renamed to config_openeuler-22.03-lts-sp1.yaml and it should refer to `bigtop/puppet:trunk-openeuler-22.03-lts-sp1` as required image.

docker/bigtop-puppet/build.sh
```diff
@@ -29,7 +29,14 @@ fi
...
 PREFIX=$(echo "$1" | cut -d '-' -f 1)
 OS=$(echo "$1" | cut -d '-' -f 2)
+
+# BIGTOP-4082 get the openEuler OS version
+if [ "${OS}" == "openeuler" ];then
+VERSION=$(echo "$1" | cut -d '-' -f 3-5)
+else
 VERSION=$(echo "$1" | cut -d '-' -f 3)
+fi
+
```
iwasakims, line 39: How about always using `VERSION=$(echo "$1" | cut -d '-' -f 3-)` ? Since the version is the last part of the argument, we can cover all the case of `22.03` and `22.03-lts-sp1` and `22.03-lts-sp1-some-more-addendum`.

iwasakims:
@MacChen01 The originally reported issue is conflict between R and openblas-0.18.5.oe2203sp3. Is it not bug of openEuler? We can not expected it to be fixed? How about [installing R from source code as done for old Ubuntu](https://github.com/apache/bigtop/blob/0fbaaa90859a889dbd86b869895224fb7126f677/bigtop_toolchain/manifests/renv.pp#L87-L117) rather than sticking with old openEuler?

docker/bigtop-puppet/build.sh
MacChen01: After communicatted with the person who in charge of the openeuler OS versoin, we got the all version list of 22.03 are:
22.03-lts-sp1/22.03-lts-sp2/22.03-lts-sp3/22.03-lts-sp4, no other version number.

MacChen01:
> [provisioner/docker/config_openeuler-22.03.yaml](https://github.com/apache/bigtop/blob/0fbaaa90859a889dbd86b869895224fb7126f677/provisioner/docker/config_openeuler-22.03.yaml) should be renamed to config_openeuler-22.03-lts-sp1.yaml and it should refer to `bigtop/puppet:trunk-openeuler-22.03-lts-sp1` as required image.

The code has been modified, thanks.

docker/bigtop-puppet/build.sh
iwasakims: The scope is not limited to openEuler. Please remove the conditional. @MacChen01

MacChen01: @iwasakims 
ok, code has been modified, thanks.

MacChen01:
> @MacChen01 The originally reported issue is conflict between R and openblas-0.18.5.oe2203sp3. Is it not bug of openEuler? We can not expected it to be fixed? How about [installing R from source code as done for old Ubuntu](https://github.com/apache/bigtop/blob/0fbaaa90859a889dbd86b869895224fb7126f677/bigtop_toolchain/manifests/renv.pp#L87-L117) rather than sticking with old openEuler?

when we adapted bigtop in openeuler, using the version of openEuler docker img is 22.03-lts-sp1, other versions have not been released yet, including sp2 and sp3. So the BIGTOP-4082 we didn't met.

MacChen01:
@iwasakims 
The reason of BIGTOP-4082 is install conflict R and openblas in openeuler 22.03-lts-sp3.
we are looking for help from the openEuler community, and analyze the bug by myself.

iwasakims (approved):
I'm merging this. Thanks, @MacChen01.

R-core and openblas conflicts even on openeuler-22.03-lst-sp1. If there is no dependency between them, the issue seems to be caused by something pulling openblas in the upstream of 22.03. I will file another JIRA if we can purge openblas in the upstream.

```
[root@a9e272ddcfd6 /]# repoquery -q -l openblas | grep libRblas
/usr/lib64/R/lib/libRblas.so

[root@a9e272ddcfd6 /]# repoquery -q -l R-core | grep libRblas
/usr/lib64/R/lib/libRblas.so
```

MacChen01:
> I'm merging this. Thanks, @MacChen01.
> 
> R-core and openblas conflicts even on openeuler-22.03-lst-sp1. If there is no dependency between them, the issue seems to be caused by something pulling openblas in the upstream of 22.03. I will file another JIRA if we can purge openblas in the upstream.
> 
> ```
> [root@a9e272ddcfd6 /]# repoquery -q -l openblas | grep libRblas
> /usr/lib64/R/lib/libRblas.so
> 
> [root@a9e272ddcfd6 /]# repoquery -q -l R-core | grep libRblas
> /usr/lib64/R/lib/libRblas.so
> ```

@iwasakims 
Thanks, if you create new JIRA to track conflicts between  R and openblas, i will also track the JIRA.

iwasakims:
> How about [installing R from source code as done for old Ubuntu](https://github.com/apache/bigtop/blob/0fbaaa90859a889dbd86b869895224fb7126f677/bigtop_toolchain/manifests/renv.pp#L87-L117) rather than sticking with old openEuler?

self reply:  BIGTOP-3967 (#1148) replaced R installed from source with R packaged by thirdpatry (oepkgs). Reverting the related code is an reasonable option.

PR #1256, opened 2024-04-10
BIGTOP-4073: spark support python3 in rockylinux 9

MacChen01:
Spark smoketest need python3 in rockylinux9.

- [ ] Does the title or this PR starts with the corresponding JIRA issue id (BIGTOP-4073)

### DIFF
--- a/bigtop-packages/src/rpm/spark/SPECS/spark.spec
+++ b/bigtop-packages/src/rpm/spark/SPECS/spark.spec
@@ -116,8 +116,9 @@ Server for Spark worker
 %package -n %{spark_pkg_name}-python
 Summary: Python client for Spark
 Group: Development/Libraries
+# BIGTOP-4073 rockylinux 8 use python2, rockylinux 9 use python3
 %if 0%{?rhel} >= 8 || 0%{?openEuler}
-%if 0%{?rhel} >= 8
+%if 0%{?rhel} == 8
 Requires: %{spark_pkg_name}-core = %{version}-%{release}, python2
 %else
 Requires: %{spark_pkg_name}-core = %{version}-%{release}, python3


MacChen01:
The query successful log:
image

The smoketest successful log:
image

JiaLiangC:
@MacChen01 
Spark 3.3.4 only supports Python 3.7+, meaning that the default version of Python on many operating systems does not meet this requirement. Simply specifying Python 3 or Python 2 for installation might work, but it won't utilize the system-provided Python. Therefore, the Python requirement in the Spark spec doesn't make sense semantically; a dependency is established, allowing for installation, but it doesn't guarantee usability.

I believe we can eliminate the Python dependency since Spark's operation does not rely on Python. Only specific functionalities require Python, and users can manually install an appropriate version if needed. Otherwise, managing Spark's Python dependencies within Bigtop would be a disaster.

MacChen01:
> @MacChen01 Spark 3.3.4 only supports Python 3.7+, meaning that the default version of Python on many operating systems does not meet this requirement. Simply specifying Python 3 or Python 2 for installation might work, but it won't utilize the system-provided Python. Therefore, the Python requirement in the Spark spec doesn't make sense semantically; a dependency is established, allowing for installation, but it doesn't guarantee usability.
> 
> I believe we can eliminate the Python dependency since Spark's operation does not rely on Python. Only specific functionalities require Python, and users can manually install an appropriate version if needed. Otherwise, managing Spark's Python dependencies within Bigtop would be a disaster.

@JiaLiangC 
Spark has the smoketest about python, and the spark-python is part of spark. 
_Spark's operation does not rely on Python_ ??

MacChen01:
@iwasakims @guyuqi 
please help review the code, thanks.

JiaLiangC:
@MacChen01 
Yes, logically speaking, testSparkPythonExample should not be able to run successfully on a machine that does not have Python 3 and has a Python version less than 3.7.

JiaLiangC:
Aside from PySpark tasks, other types of Spark tasks do not depend on Python, so I think it would be better to remove the Python dependency in Spark.

MacChen01:
@JiaLiangC
 Thanks for your advice, removing the python from spark is a complicated thing. I think it should not be handled in bigtop.
You can create a discuss or JIRA for it.

@iwasakims @guyuqi
please help review the code, thanks.

iwasakims:
@MacChen01 @JiaLiangC I'm +1 on dropping package dependency on python. BIGTOP-4053 looks filed for this.

If we need /usr/bin/python pointing to python3 for smoke-tests on Rocky Linux 9, python-unversioned-command package should be installed by the manifest of bigtop-deploy.

iwasakims (approved):
I'm merging this while it might be temporal fix.

PR #1257, opened 2024-04-16
BIGTOP-4090. Remove obsolete Python installation from puppet.sh.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4090

Installing of python2 is already covered by [the Puppet manifest of bigtop_toolchain](https://github.com/apache/bigtop/blob/2cc7b4042d9c359f4c248b2639dea777e91668e7/bigtop_toolchain/manifests/packages.pp#L405-L439).

While this can be part of BIGTOP-4092, I'm submitting this PR first for reducing build time of bigtop/puppet docker image of openEuler.


iwasakims:
#1262 will cover this.

PR #1258, opened 2024-04-17
BIGTOP-4093. Avoid oepkgs for installing R on openEuler.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4093

This PR replace R packaged by oepkgs with R installed from source code. We do not need to be bound to 22.03 LTS SP1.

Since the manifest before BIGTOP-3967 did not work due to conflicting `$url` variables for pandoc an R, this PR fixed the part.

bigtop/slaves:openeuler-22.03 was successfully built with R installation and packaging of Spark succeeded in the image.

```
$ ./gradlew spark-pkg-ind -Pprefix=trunk -POS=openeuler-22.03 -Pdocker-run-option="--privileged"
```


sekikn:
+1, tested with the following commands. Thanks @iwasakims.

```
$ curl -sL https://patch-diff.githubusercontent.com/raw/apache/bigtop/pull/1258.diff | git apply
$ (cd docker/bigtop-puppet && ./build.sh trunk-openeuler-22.03)
$ (cd docker/bigtop-slaves && ./build.sh trunk-openeuler-22.03)
```

iwasakims:

PR #1260, opened 2024-04-23
BIGTOP-4097. Fix missing R package resource in toolchain manifest.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4097

R can not be installed on OS distributions other than openEuler due to missing package resource definition unintentionally removed by [BIGTOP-4093](https://issues.apache.org/jira/browse/BIGTOP-4093).

### DIFF
--- a/bigtop_toolchain/manifests/renv.pp
+++ b/bigtop_toolchain/manifests/renv.pp
@@ -51,6 +51,11 @@ class bigtop_toolchain::renv {
     }
   }
 
+  package { $pkgs:
+    ensure => installed,
+    before => [Exec["install_r_packages"]]
+  }
+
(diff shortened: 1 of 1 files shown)


sekikn:
+1, tested with another distro than openeuler.

```
$ curl -sL https://github.com/apache/bigtop/pull/1260.diff | git apply
$ (cd docker/bigtop-puppet && ./build.sh trunk-ubuntu-22.04)
$ (cd docker/bigtop-slaves && ./build.sh trunk-ubuntu-22.04)
$ docker run -it bigtop/slaves:trunk-ubuntu-22.04 R --version
```

iwasakims:

PR #1263, opened 2024-04-26
BIGTOP-4100: Set default python to python3 in bigtop deploy

JiaLiangC:
### Description of PR
Set default python to python3 in bigtop deploy

Since CentOS 7's yum depends on Python 2, it's necessary to replace the Python interpreter in yum with Python 2.7.

Rocky 8 comes with /usr/libexec/no-python, so only an update-alternatives --set is needed.

Starting from Red Hat 9, the python3-unversioned-command package is provided, thus only the installation of this package is required.

Setting Python3 as the default Python version across various operating systems:

1. **CentOS 7:**
   - Install Python 3 development package:
     ```
     yum install python3-devel
     ```
   - Set Python 3 as the default:
     ```
     update-alternatives --install /usr/bin/python python /usr/bin/python3 1
     ```

2. **Rocky Linux 8:**
   - Installation:
     ```
     yum install python3-devel
     ```
   - Due to the presence of `/etc/alternatives/python -> /usr/libexec/no-python`, the command to set Python 3 as the default is:
     ```
     update-alternatives --set python /usr/bin/python3
     ```

3. **Rocky Linux 9:**
   - Installation commands:
     ```
     yum install python3-devel 
     yum install python-unversioned-command
     ```

4. **Fedora 35 to Fedora 38:**
   - For all mentioned Fedora versions, the installation commands are:
     ```
     yum install python3-devel
     yum install python-unversioned-command
     ```

5. **Debian 10 (Buster):**
   - Installation commands:
     ```
     apt install python3-dev
     apt install python-is-python3
     ```

6. **Debian 11 (Bullseye):**
   - Installation commands:
     ```
     apt install python3-dev 
     apt install python-is-python3
     ```

7. **Ubuntu 20.04 (Focal Fossa)** and **Ubuntu 22.04 (Jammy Jellyfish):**
   - Installation commands for both versions:
     ```
     apt install python3-dev 
     apt install python-is-python3
     ```

**Results of Default Python Version Setting Across Operating Systems:**
- For CentOS 7 and 8, manual setting is required.
- CentOS 7 relies on Python 2 for yum, so it's necessary to replace Python within yum to Python 2.7.
- Rocky Linux 8 comes pre-configured with `/usr/libexec/no-python`, thus only requiring the `update-alternatives --set` command.
- Starting from Red Hat 9, the package `python-unversioned-command` is available, negating the need for additional settings beyond package installation.
  
- Testing revealed that Fedora versions 35 to 38 all include the `python-unversioned-command` package.
- Debian 10, Debian 11, Ubuntu 20.04, and Ubuntu 22.04 all include the `python-is-python3` package, facilitating the setting of Python 3 as the default version.

manual test by running puppet script
 
image

### For code changes:

### DIFF
--- a/bigtop-deploy/puppet/manifests/python.pp
+++ b/bigtop-deploy/puppet/manifests/python.pp
@@ -29,7 +29,7 @@ class python {
                     '8': {
                         exec { 'set-python3':
                           command => '/usr/sbin/update-alternatives --set python /usr/bin/python3',
-                          unless  => "/usr/sbin/update-alternatives --display python | grep  '/usr/bin/python3'",
+                          unless  => "/usr/sbin/update-alternatives --display python | grep  'link currently points to /usr/bin/python3'",
                           path    => ['/bin', '/usr/bin', '/sbin', '/usr/sbin'],
                         }
                     }
--- a/bigtop-deploy/puppet/manifests/python.pp
+++ b/bigtop-deploy/puppet/manifests/python.pp
@@ -14,66 +14,46 @@
 # limitations under the License.
 
 class python {
-
-    define set_python_alternatives($command) {
-        exec { $name:
-          command => $command,
-          unless  => "/usr/sbin/update-alternatives --display python | grep '/usr/bin/python3'",
-          path    => ['/bin', '/usr/bin', '/sbin', '/usr/sbin'],
-        }
-    }
-
-    define change_python_interpreter($file, $interpreter) {
-        file_line { "change_interpreter_in_${name}":
-          path  => $file,
-          line  => "#!${interpreter}",
-          match => '^#!.*python.*$',
-        }
-    }
-
     case $operatingsystem {
         /(?i:(centos|fedora|redhat|rocky))/: {
+            package { 'python3-devel':
+              ensure => 'present',
+            }
             if ($operatingsystem != 'Fedora') {
                 case $operatingsystemmajrelease {
                     '9': {
-                        package { 'python3-unversioned-command':
+                        package { 'python-unversioned-command':
                           ensure => 'present',
                         }
                     }
                     '8': {
-                        set_python_alternatives { 'set-python3':
+                        exec { 'set-python3':
                           command => '/usr/sbin/update-alternatives --set python /usr/bin/python3',
-                        }
-                    }
-                    '7': {
-                        change_python_interpreter { 'yum':
-                          file        => '/usr/bin/yum',
-                          interpreter => '/usr/bin/python2.7',
-                        }
-                        change_python_interpreter { 'urlgrabber-ext-down':
-                          file        => '/usr/libexec/urlgrabber-ext-down',
-                          interpreter => '/usr/bin/python2.7',
-                        }
-
-                        set_python_alternatives { 'install-python3':
-                          command => '/usr/sbin/update-alternatives --install /usr/bin/python python /usr/bin/python3 1',
+                          unless  => "/usr/sbin/update-alternatives --display python | grep  '/usr/bin/python3'",
+                          path    => ['/bin', '/usr/bin', '/sbin', '/usr/sbin'],
                         }
                     }
                 }
             }else {
-                set_python_alternatives { 'install-python3':
-                  command => '/usr/sbin/update-alternatives --install /usr/bin/python python /usr/bin/python3 1',
+                package { 'python-unversioned-command':
+                  ensure => 'present',
                 }
             }
         }
 
         /(Ubuntu|Debian)/: {
-            set_python_alternatives { 'install-python3-debian':
-              command => '/usr/sbin/update-alternatives --install /usr/bin/python python /usr/bin/python3 1',
+            package { 'python3-dev':
+              ensure => 'present',
+            }
+            package { 'python-is-python3':
+              ensure => 'present',
             }
         }
 
         /openEuler/: {
+            package { 'python3-devel':
+              ensure => 'present',
+            }
             package { 'python3-unversioned-command':
               ensure => 'present',
             }
--- a/bigtop-deploy/puppet/manifests/python.pp
+++ b/bigtop-deploy/puppet/manifests/python.pp
@@ -14,16 +14,70 @@
 # limitations under the License.
 
 class python {
-    if (($operatingsystem == 'Ubuntu' and 0 <= versioncmp($operatingsystemmajrelease, '22.04'))) {
-        file { '/usr/bin/python':
-          ensure => 'link',
-          target => '/usr/bin/python2',
+
+    define set_python_alternatives($command) {
+        exec { $name:
+          command => $command,
+          unless  => "/usr/sbin/update-alternatives --display python | grep '/usr/bin/python3'",
+          path    => ['/bin', '/usr/bin', '/sbin', '/usr/sbin'],
         }
     }
-    
-    if ($operatingsystem == 'openEuler') {
-        package { 'python3-unversioned-command':
-          ensure => 'present'
+
+    define change_python_interpreter($file, $interpreter) {
+        file_line { "change_interpreter_in_${name}":
+          path  => $file,
+          line  => "#!${interpreter}",
+          match => '^#!.*python.*$',
+        }
+    }
+
+    case $operatingsystem {
+        /(?i:(centos|fedora|redhat|rocky))/: {
+            if ($operatingsystem != 'Fedora') {
+                case $operatingsystemmajrelease {
+                    '9': {
+                        package { 'python3-unversioned-command':
+                          ensure => 'present',
+                        }
+                    }
+                    '8': {
+                        set_python_alternatives { 'set-python3':
+                          command => '/usr/sbin/update-alternatives --set python /usr/bin/python3',
+                        }
+                    }
+                    '7': {
+                        change_python_interpreter { 'yum':
+                          file        => '/usr/bin/yum',
+                          interpreter => '/usr/bin/python2.7',
+                        }
+                        change_python_interpreter { 'urlgrabber-ext-down':
+                          file        => '/usr/libexec/urlgrabber-ext-down',
+                          interpreter => '/usr/bin/python2.7',
+                        }
+
+                        set_python_alternatives { 'install-python3':
+                          command => '/usr/sbin/update-alternatives --install /usr/bin/python python /usr/bin/python3 1',
+                        }
+                    }
+                }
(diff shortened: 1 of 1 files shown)


JiaLiangC:
@iwasakims Could you help review this pr?

JiaLiangC:
I've tried to avoid directly modifying operating system files as much as possible. However, in Red Hat 7 series operating systems, once the default Python is set to Python3, it results in Yum becoming unusable. The only solution seems to be modifying Yum . I haven't thought of a more elegant method. If there are any better suggestions, I welcome them.

pan3793:
A naive quesiton: seems all OS supports `/usr/bin/python3` directly? Insteaf of enforced to use `python`, why set `#!/usr/bin/env python3` or explicitly call `python3 xxx`? I saw Spark already did in this way.

JiaLiangC:
@pan3793 
Some components, such as Spark and Ambari, have already explicitly set specific versions of the interpreter. However, in many components like Ranger and gpdb, the interpreter is defined in the Python scripts as #!/usr/bin/env python, so it's necessary to set the default Python to accommodate this scenario.

bigtop-deploy/puppet/manifests/python.pp
```diff
@@ -14,16 +14,70 @@
...
+                          interpreter => '/usr/bin/python2.7',
+                        }
+                        change_python_interpreter { 'urlgrabber-ext-down':
+                          file        => '/usr/libexec/urlgrabber-ext-down',
+                          interpreter => '/usr/bin/python2.7',
+                        }
+
+                        set_python_alternatives { 'install-python3':
+                          command => '/usr/sbin/update-alternatives --install /usr/bin/python python /usr/bin/python3 1',
+                        }
```
iwasakims, line 60: python3 must be installed before this (by package resource) in order to make this functional in the bigtop/puppet image.

```
Notice: Roles to deploy: [bigtop-utils]
Notice: Compiled catalog for 670e444169cb.bigtop.apache.org in environment production in 0.85 seconds
Notice: /Stage[main]/Hadoop_cluster_node/User[testuser]/ensure: created
Notice: /Stage[main]/Hadoop_cluster_node/User[jenkins]/ensure: created
Notice: /Stage[main]/Bigtop_repo/Yumrepo[Bigtop_0]/ensure: created
Notice: /Stage[main]/Python/Python::Set_python_alternatives[install-python3]/Exec[install-python3]/returns: executed successfully
Notice: /Stage[main]/Python/Python::Change_python_interpreter[urlgrabber-ext-down]/File_line[change_interpreter_in_urlgrabber-ext-down]/ensure: created
Notice: /Stage[main]/Python/Python::Change_python_interpreter[yum]/File_line[change_interpreter_in_yum]/ensure: created
Notice: Baseurl: file:///bigtop-home/output
Notice: /Stage[main]/Bigtop_repo/Notify[Baseurl: file:///bigtop-home/output]/message: defined 'message' as 'Baseurl: file:///bigtop-home/output'
Notice: /Stage[main]/Hadoop_cluster_node/User[hudson]/ensure: created
Error: /Stage[main]/Bigtop_utils::Client/Package[bigtop-utils]: Provider yum is not functional on this host
Error: /Stage[main]/Jdk/Package[jdk]: Provider yum is not functional on this host

```

bigtop-deploy/puppet/manifests/python.pp
```diff
@@ -14,16 +14,70 @@
...
+                    }
+                    '7': {
+                        change_python_interpreter { 'yum':
+                          file        => '/usr/bin/yum',
+                          interpreter => '/usr/bin/python2.7',
+                        }
+                        change_python_interpreter { 'urlgrabber-ext-down':
+                          file        => '/usr/libexec/urlgrabber-ext-down',
+                          interpreter => '/usr/bin/python2.7',
+                        }
```
iwasakims, line 56: We seem not be able to cover all scenario by this hack.

```
Debug: Executing '/usr/bin/python /usr/local/share/gems/gems/puppet-3.6.2/lib/puppet/provider/package/yumhelper.py'
Error: Could not get latest version: Execution of '/usr/bin/python /usr/local/share/gems/gems/puppet-3.6.2/lib/puppet/provider/package/yumhelper.py' returned 1: File "/usr/local/share/gems/gems/puppet-3.6.2/lib/puppet/provider/package/yumhelper.py", line 40
    raise ValueError, a
                    ^
SyntaxError: invalid syntax
Error: /Stage[main]/Bigtop_utils::Client/Package[bigtop-utils]/ensure: change from 3.3.0-1.el7 to latest failed: Could not get latest version: Execution of '/usr/bin/python /usr/local/share/gems/gems/puppet-3.6.2/lib/puppet/provider/package/yumhelper.py' returned 1: File "/usr/local/share/gems/gems\
/puppet-3.6.2/lib/puppet/provider/package/yumhelper.py", line 40
    raise ValueError, a
                    ^
SyntaxError: invalid syntax
```

iwasakims:
@JiaLiangC How about leave CentOS 7 as is (i.e. python2 as default) and commit other part first?

JiaLiangC:
> @JiaLiangC How about leave CentOS 7 as is (i.e. python2 as default) and commit other part first?
@iwasakims  Setting the default Python version to Python 3 on CentOS 7 is quite challenging. Even after changing the default to Python 3, you may encounter issues where other programs that still rely on Python 2 stop working. Therefore, it is advisable to address these other situations first.

bigtop-deploy/puppet/manifests/python.pp
JiaLiangC: Alright, we'll have to temporarily ignore dealing with CentOS 7 for now.

iwasakims:
> Even after changing the default to Python 3, you may encounter issues where other programs that still rely on Python 2 stop working.

yeah. I know as I commented in the patch. I'm happy we can just cover the coverage of current smoke-tests. We can put python setup into user's hand.

bigtop-deploy/puppet/manifests/python.pp
JiaLiangC: Okay, I did overlook the fact that the deploy image doesn't have Python 3 installed by default. I'll add that part.

iwasakims:
I'm ok to declare some limitation on CentOS 7 in the release. CentOS 7 will be EOL on June 30, 2024.

JiaLiangC:
> I'm ok to declare some limitation on CentOS 7 in the release. CentOS 7 will be EOL on June 30, 2024.

Agreed, I'll skip setting the default Python on CentOS 7. Users can manage multiple Python versions on CentOS 7 themselves.

bigtop-deploy/puppet/manifests/python.pp
```diff
@@ -14,16 +14,57 @@
...
+                case $operatingsystemmajrelease {
+                    '9': {
+                        package { 'python-unversioned-command':
+                          ensure => 'present',
+                        }
+                    }
+                    '8': {
+                        set_python_alternatives { 'set-python3':
+                          command => '/usr/sbin/update-alternatives --set python /usr/bin/python3',
+                        }
```
iwasakims, line 41: Thanks for the update, @JiaLiangC. The user defined resource can not be resolved without quolifying with the class name like `python::set_python_alternatives` on Puppet 6 used in Rocky Linux 8. I think we can use plain `exec` if it is not used in other places.

```
Error: Evaluation Error: Error while evaluating a Resource Statement, Unknown resource type: 'set_python_alternatives' (file: /bigtop-home/bigtop-deploy/puppet/manifests/python.pp, line: 39, column: 25) on node 70361087e339.bigtop.apache.org
```

In addition, evaluation of the resource is skipped because the unless conditional is true even before updating alternatives.

```
[root@1ef2ea9dd4a9 /]# alternatives --display python
python - status is auto.
 link currently points to /usr/libexec/no-python
/usr/libexec/no-python - priority 404
 slave unversioned-python: (null)
 slave unversioned-python-man: /usr/share/man/man1/unversioned-python.1.gz
/usr/bin/python3 - priority 300
 slave unversioned-python: /usr/bin/python3
 slave unversioned-python-man: /usr/share/man/man1/python3.1.gz
Current `best' version is /usr/libexec/no-python.

[root@1ef2ea9dd4a9 /]# /usr/sbin/update-alternatives --display python | grep '/usr/bin/python3' && echo true
/usr/bin/python3 - priority 300
 slave unversioned-python: /usr/bin/python3
true
```

JiaLiangC: Thanks for the review help. I also noticed this issue during testing. Still testing locally~

JiaLiangC:
image
image
@iwasakims  The issue has been fixed and it has been successfully tested on Rocky 8. The previous unless logic had a problem. Could you please help review it again?

bigtop-deploy/puppet/manifests/python.pp
```diff
@@ -14,16 +14,50 @@
...
+                case $operatingsystemmajrelease {
+                    '9': {
+                        package { 'python-unversioned-command':
+                          ensure => 'present',
+                        }
+                    }
+                    '8': {
+                        exec { 'set-python3':
+                          command => '/usr/sbin/update-alternatives --set python /usr/bin/python3',
+                          unless  => "/usr/sbin/update-alternatives --query python | grep -q 'Value: /usr/bin/python3'",
```
iwasakims, line 32: Hmm. `--query` seems not to be valid option of `update-alternatives` of rockylinux-8. @JiaLiangC 

```
[root@4ff53a675cd3 /]# /usr/sbin/update-alternatives --query python
alternatives version 1.19.1 - Copyright (C) 2001 Red Hat, Inc.
This may be freely redistributed under the terms of the GNU Public License.

usage: alternatives --install <link> <name> <path> <priority>
                    [--initscript <service>]
                    [--family <family>]
                    [--slave <slave_link> <slave_name> <slave_path>]*
       alternatives --remove <name> <path>
       alternatives --auto <name>
       alternatives --config <name>
       alternatives --display <name>
       alternatives --set <name> <path>
       alternatives --list
       alternatives --remove-all <name>
       alternatives --add-slave <name> <path> <slave_link> <slave_name> <slave_path>
       alternatives --remove-slave <name> <path> <slave_name>

common options: --verbose --test --help --usage --version --keep-missing --keep-foreign
                --altdir <directory> --admindir <directory>
```

JiaLiangC: already fixed.

iwasakims (approved):
LGTM. Thanks, @JiaLiangC.

PR #1266, opened 2024-05-01
BIGTOP-4102. Fix failure of toolchain manifest on Fedora 38 due to unresolved facter variable

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4102

```
[91m[1;31mError: Evaluation Error: Unknown variable: '::operatingsystem'. (file: /etc/puppet/code/modules/bigtop_toolchain/manifests/jdk11.pp, line: 17, column: 8) on node bigtop-slave-arm5.novalocal[0m
[0mThe command '/bin/sh -c if [ -f ~/.bash_profile ]; then . ~/.bash_profile; fi &&     dnf clean all && dnf updateinfo &&     puppet apply -e "include bigtop_toolchain::installer" || if [ $? -ne 2 ]; then exit 1; fi' returned a non-zero code: 1
```

[Legacy facts are disabled by default in Puppet 8](https://www.puppet.com/docs/pe/2023.5/osp/upgrading-from-puppet7-to-puppet8.html#upgrading-from-puppet7-to-puppet8-legacy-facts-deprecation). include_legacy_facts=true in puppet.conf will re-enable it.

This PR is quick fix only targeting fedora-38. I filed [BIGTOP-4103](https://issues.apache.org/jira/browse/BIGTOP-4103) as a follow-up.

### DIFF
--- a/bigtop-deploy/puppet/hiera.yaml
+++ b/bigtop-deploy/puppet/hiera.yaml
@@ -3,6 +3,5 @@
   :datadir: /etc/puppet/hieradata
 :hierarchy:
   - site
-  - "bigtop/%{hadoop_hiera_ha_path}"
   - bigtop/cluster
   - bigtop/repo
--- a/bigtop-deploy/puppet/hieradata/bigtop/ha.yaml
+++ b/bigtop-deploy/puppet/hieradata/bigtop/ha.yaml
@@ -1,7 +1,0 @@
----
-hadoop::common_hdfs::ha: "manual"
-hadoop::common_hdfs::hadoop_namenode_host:
-  - "%{hiera('bigtop::hadoop_head_node')}"
-  - "%{hiera('bigtop::standby_head_node')}"
-hadoop::common_hdfs::hadoop_ha_nameservice_id: "ha-nn-uri"
-hadoop_cluster_node::hadoop_namenode_uri: "hdfs://%{hiera('hadoop_ha_nameservice_id')}:8020"
--- a/bigtop-deploy/puppet/hieradata/bigtop/noha.yaml
+++ b/bigtop-deploy/puppet/hieradata/bigtop/noha.yaml
@@ -1,2 +1,0 @@
----
-# all done via defaults
--- a/bigtop-deploy/puppet/hieradata/site.yaml
+++ b/bigtop-deploy/puppet/hieradata/site.yaml
@@ -11,6 +11,12 @@ hadoop::hadoop_storage_dirs:
   - /data/3
   - /data/4
 
+#hadoop::common_hdfs::ha: "auto"
+#hadoop::common_hdfs::hadoop_namenode_host:
+#  - "%{hiera('bigtop::hadoop_head_node')}"
+#  - "%{hiera('bigtop::standby_head_node')}"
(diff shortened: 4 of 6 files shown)


yoda-mon:
@iwasakims It seems work well but I have one concern about the Puppet version.

After applying this patch
```sh
./gradlew -Pprefix=trunk -POS=fedora-38 bigtop-puppet
./gradlew -Pprefix=trunk -POS=fedora-38 bigtop-slaves
```
Both builds succeeded. but
```
docker run -it --rm bigtop/puppet:trunk-fedora-38-aarch64  puppet --version
```
showed `8.3.1`. The images that uploads to dockerhub shows `7.27.0` today, and the difference might affect the smoke tests.

This may not be caused by this PR, but we might have to fix the version.

iwasakims:
@yoda-mon The images on Docker Hub will be updated by running Jenkins jobs. I will run them after merging this PR.
* https://ci.bigtop.apache.org/job/Docker-Puppet-Trunk/
* https://ci.bigtop.apache.org/job/Docker-Toolchain-Trunk/

If the manifest of bigtop-deploy contains code incompatible with Puppet 8, it should be fixed in follow-up JIRAs.

iwasakims:
Actually [deprecation of Hiera 3](https://www.puppet.com/docs/puppet/8/upgrading-from-puppet7-to-puppet8.html#upgrading-from-puppet7-to-puppet8-heira-3-deprecation) affects bigtop-deploy. I will add commit to fix this.

```
Error: Undefined variable 'hadoop_hiera_ha_path' (file: /etc/puppet/hiera.yaml, line: 8) on node 1ac306408ff6.bigtop.apache.org
```

iwasakims:
puppet 8 packaged by fedora-38 is coupled with hiera 3. I guess the error is caused by just variable lookup settings.

```
[root@495d41ec7cd8 /]# rpm -q hiera
hiera-3.11.0-2.fc38.noarch

[root@495d41ec7cd8 /]# rpm -q --whatrequires hiera
puppet-8.3.1-1.fc38.noarc
```

iwasakims:
The current manifest [tries to conditionaly switch hiera file](https://github.com/apache/bigtop/blob/3808ba7d1b8d6c2504c08ed9079a9d5673ae83de/bigtop-deploy/puppet/hiera.yaml#L6) based on [the variable in class](https://github.com/apache/bigtop/blob/3808ba7d1b8d6c2504c08ed9079a9d5673ae83de/bigtop-deploy/puppet/manifests/cluster.pp#L158-L163). We can not assume that the class local variable is visible. We should switch ha/non-ha by explicit declaration in site.yaml.

iwasakims:
The cause of the error turned out to be bug of the manifest rather than Puppet version. smoke-tests of Hadoop passed on both fedora-38 and centos-7 with the updated manifest.

iwasakims:
@yoda-mon Could you take a look again? This issue blocks release of Bigtop 3.3.0.

yoda-mon:
@iwasakims (Sorry for the late, I was on vacation) 
Thank you for the information, I understood the situations related to the deprecation of Hiera 3.
On that case, the build itself was suceeded so let's merge this.

yoda-mon (approved):

iwasakims:
Thanks, @yoda-mon.

PR #1267, opened 2024-05-07
BIGTOP-4104: Introduce Apache Doris into Bigtop

timyuer:
### For code changes:


timyuer:
@sekikn 
Sorry to bother, I found that `snappy-devel` causes doris to fail to compile during the compilation process, so I can ask which components are used by this package?

PR #1274, opened 2024-05-31
BIGTOP-4114. Fix build failure of Hadoop and Alluxio due to node-gyp error on rockylinux-8 and fedora-38 aarch64.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4114

Unfortunately we seem to need python2 for rebuilding native libraries required for (old?) node-spss by node-gyp. `/usr/bin/python` pointing to `/usr/bin/pytyon3` did not fix this. `/usr/bin/python` is not needed if `/usr/bin/python2` exists. Building Hadoop (with yarn-ui enabled) and Alluxio on rockylinux-8 and fedora-38 aarch \
64 is affected by this. While Alluxio on openeuler-22.03 on aarch64 seems to be affected too, I'm going to declare it as unsupported in Bigtop 3.3.0 since openeuler-22.03 does not provide python2 package.

By applying current bigtop_toolchain manifest, CentOS 7 has python2 since OS core packages depend on it. Debian 11 and Ubuntu 22.04 have python2 as dependency of installed packages.

I moved python related staffs to dedicated python.pp for ease of maintenance.

### DIFF
--- a/bigtop_toolchain/manifests/installer.pp
+++ b/bigtop_toolchain/manifests/installer.pp
@@ -21,6 +21,7 @@ class bigtop_toolchain::installer {
   include bigtop_toolchain::gradle
   include bigtop_toolchain::protobuf
   include bigtop_toolchain::packages
+  include bigtop_toolchain::python
   include bigtop_toolchain::env
   include bigtop_toolchain::user
   include bigtop_toolchain::renv
--- a/bigtop_toolchain/manifests/packages.pp
+++ b/bigtop_toolchain/manifests/packages.pp
@@ -68,7 +68,6 @@ class bigtop_toolchain::packages {
         "perl-Digest-SHA",
         "nasm",
         "yasm",
-        "python3-devel"
       ]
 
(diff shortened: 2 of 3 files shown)


sekikn:
+1, ensured that both Rocky 8 and Fedora 38 have python2 on aarch64.

```
$ curl -sL https://github.com/apache/bigtop/pull/1274.diff | git apply
$ cd docker/bigtop-puppet
$ for i in centos-7 rockylinux-8 fedora-38; do ./build.sh trunk-$i; done
$ cd -
$ cd docker/bigtop-slaves
$ for i in centos-7 rockylinux-8 fedora-38; do ./build.sh trunk-$i; done
$ for i in centos-7 rockylinux-8 fedora-38; do docker run -it bigtop/slaves:trunk-$i-aarch64 python2 -V; done
Python 2.7.5
Python 2.7.18
Python 2.7.18
```

iwasakims:
Thanks, @sekikn. I merged this to master and branch-3.3.

PR #1294, opened 2024-09-09
BIGTOP-4215. Upgrade ZooKeeper to 3.8.4

masatana:
### Description of PR
The latest Hadoop uses 3.8.4 so we can update to that version.

https://github.com/apache/hadoop/blob/trunk/hadoop-project/pom.xml#L106

Built & test with Ubuntu-22.04
```
$ ./gradlew allclean zookeeper-pkg repo -Dbuildwithdeps=true
$ ./docker-hadoop.sh -c 1 --image bigtop/puppet:trunk-ubuntu-22.04 --docker-compose-yml docker-compose-cgroupv2.yml -dcp --memory 8g --repo file:///bigtop-home/output/apt --disable-gpg-check --stack zookeeper --smoke-tests zookeeper 
```

### For code changes:

### DIFF
--- a/bigtop.bom
+++ b/bigtop.bom
@@ -137,7 +137,7 @@ bigtop {
       pkg     = name
       rpm_pkg_suffix = "_" + bigtop.base_version.replace(".", "_")
       version {
-        base  = '3.7.2'
+        base  = '3.8.4'
         pkg   = base
         release = 1
       }


masatana:
We might need to include this patch https://github.com/apache/hadoop/pull/6675/files to build hadoop 3.3.6 against zookeeper 3.8.4.

iwasakims (approved):
LGTM. smoke-tests passed on rockylinux-8-aarch64 too. We should address the issue of downstream products in follow-up JIRAs. Thanks, @masatana.

PR #1296, opened 2024-09-14
BIGTOP-4157. Drop CentOS 7

yamasakisua:
### Description of PR
https://issues.apache.org/jira/browse/BIGTOP-4157

Replace CentOS-7 to RockyLinux-8 in docs.

In some cases, default OS is setting to CentOS-7.
I set the default to Ubuntu, one of the most used operating systems in Bigtop supported OS.

But some concern still exists.
e.g.  `-Pdocker-run-option="--privileged"` must need if `-POS` option is not set.

If anyone have any good ideas, I'd like to hear comments.

```
./gradlew zookeeper-pkg-ind -Pdocker-run-option="--privileged"
...

BUILD SUCCESSFUL in 21s
1 actionable task: 1 executed
```

```
$ ./docker-hadoop.sh -dcp --docker-compose-yml docker-compose-cgroupv2.yml -k zookeeper -c 1
...
Notice: Baseurl: http://repos.bigtop.apache.org/releases/3.3.0/ubuntu/22.04/$(ARCH)
...
Notice: /Stage[main]/Hadoop_zookeeper::Server/Service[zookeeper-server]/ensure: ensure changed 'stopped' to 'running'
Notice: Applied catalog in 58.65 seconds
$
```

### For code changes:


packages.gradle
```diff
@@ -745,14 +745,14 @@ def genTasks = { target ->
   }
   task "$target-pkg-ind" (
           description: "Invoking a native binary packaging for $target in Docker. Usage: \$ ./gradlew " +
-                  "-POS=[centos-7|fedora-38|debian-11|ubuntu-18.04] " +
```
iwasakims, line 748: We can update the mention about ubuntu-18.04 in another JIRA.

iwasakims (approved):
LGTM. Thanks, @yamasakisua.

> -Pdocker-run-option="--privileged" must need if -POS option is not set.

Maybe we can automate it based on the distro in follow-up JIRA.

PR #1302, opened 2024-11-14
BIGTOP-4083. Upgrade Hive to 4.0.1.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4083

* bumped hive to 4.0.1
* removed obsolete patches.
* added excludes to rat check for files placed on deb packaging.
* added JDBC URL param since hive CLI is wrapper of beeline now.

I want to merge this first then address further nits in follow-up JIRAs.


iwasakims:
smoke-tests passed on Ubuntu 22.04 x86_64.

sekikn:
+1 also tested with Debian 11 on x86_64. Thanks @iwasakims.

PR #1320, opened 2024-12-20
BIGTOP-4309. Add Fedora 40 support to toolchain.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4309

as a candidate for Bigtop 3.4.0 release.

Issues related to packaging will be addressed in follow-up JIRAs.


sekikn:
+1, thanks @iwasakims.

PR #1321, opened 2024-12-24
BIGTOP-4310: Executing `{hadoop_mapreduce_home}/bin/mapred` would occur ERROR that the mapred-config.sh doesn't exist

guluo2016:
### Description of PR

Details see: [BIGTOP-4310](https://issues.apache.org/jira/browse/BIGTOP-4310)

Fix the issue that executing `{hadoop_mapreduce_home}/bin/mapred` would occur ERROR.

Applying this changes，
And executing the command: `{hadoop_mapreduce_home}/bin/mapred --help`, would print messages, as fallow.
```
Usage: mapred [OPTIONS] SUBCOMMAND [SUBCOMMAND OPTIONS]
 or    mapred [OPTIONS] CLASSNAME [CLASSNAME OPTIONS]
  where CLASSNAME is a user-provided Java class

  OPTIONS is none or any of:
......
```

### For code changes:


iwasakims:
Bigtop provides `/usr/bin/mapred` as a wrapper script for setting up required environment such as HADOOP_LIBEXEC_DIR and JAVA_HOME. If users want to directly invoke `/usr/lib/hadoop-mapreduce/bin/mapred`, they should set required environment on their own. Same as /usr/bin/hadoop and /usr/bin/yarn. @guluo2016

guluo2016:
Understand, thanks for your reply.

PR #1329, opened 2025-02-15
BIGTOP-4364. Fix Zeppelin's default settings.

sekikn:
### Description of PR

This PR fixes the Puppet manifest of Zeppelin and its related files so that Zeppelin works just after deployed using them.

Created a cluster with the revised files and ensured Spark example notebooks worked by default.

### For code changes:


iwasakims (approved):
+1. Thanks, @sekikn.

PR #1337, opened 2025-03-13
BIGTOP-4388. Upgrade Ranger to 2.6.0.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4388

We need to upgrade Ranger before releasing Bigtop 3.4.0.

- Ranger 2.4.0 is not compatible with HBase 2.6. ([RANGER-4862](https://issues.apache.org/jira/browse/RANGER-4862))
- Build fails due to unresolvable dependency. ([RANGER-4919](https://issues.apache.org/jira/browse/RANGER-4919))
- We need backport of [RANGER-4952](https://issues.apache.org/jira/browse/RANGER-4952) for Hive 4.
- `credValidator.uexe` is not available after [RANGER-5075](https://issues.apache.org/jira/browse/RANGER-5075) removed unixauthnative from modules list.


sekikn:
+1, tested with Debian 12 and Ubuntu 24.04 on x86_64, as follows:

```
$ curl -sL https://github.com/apache/bigtop/pull/1337.diff | git apply
$ ./gradlew ranger-clean ranger-pkg repo -Dbuildwithdeps=true
$ cd provisioner/docker
$ ./docker-hadoop.sh -d -dcp -C config_debian-12.yaml -F docker-compose-cgroupv2.yml -G -L -k bigtop-utils,hdfs,ranger -s ranger -c 1
```

iwasakims:
Thanks, @sekikn. I merged this to branch-3.4 and master.

PR #1345, opened 2025-05-08
BIGTOP-4418. Fix installation failure of packages on Fedora 40 due to lack of /lib/lsb/init-functions.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4418

While redhat-lsb provides the [/lib/lsb/init-functions](https://github.com/apache/bigtop/blob/master/bigtop-packages/src/rpm/zookeeper/SPECS/zookeeper.spec#L138), it can not be pulled by package dependency nor Puppet manifest.

```
Error: /Stage[main]/Hadoop_zookeeper::Server/Package[zookeeper-server]/ensure: change from 'purged' to 'latest' failed: Could not update: Execution of '/usr/bin/dnf -d 0 -e 1 -y install zookeeper-server' returned 1: Error:
 Problem: conflicting requests
  - nothing provides /lib/lsb/init-functions needed by zookeeper-server-3.8.4-1.fc40.x86_64 from Bigtop_0
```

This PR adds Fedora specific conditionals to the .spec files. While I think conditionals for SUSE and Megaia are obsolete and useless, I left them as-is. It should be removed by follow-up JIRAs.

### DIFF
--- a/bigtop-deploy/puppet/modules/hadoop/manifests/init.pp
+++ b/bigtop-deploy/puppet/modules/hadoop/manifests/init.pp
@@ -29,7 +29,7 @@ class hadoop ($hadoop_security_authentication = "simple",
   $hadoop_http_authentication_type = undef,
   $hadoop_http_authentication_signature_secret = undef,
   $hadoop_http_authentication_signature_secret_file = "/etc/hadoop/conf/hadoop-http-authentication-signature-secret",
-  $hadoop_http_authentication_cookie_domain = $fqdn.split("\\.")[1, -1].join("."),
+  $hadoop_http_authentication_cookie_domain = regsubst($fqdn, "^[^\\.]+\\.", ""),
 ) {
 
   include stdlib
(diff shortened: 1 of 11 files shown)


sekikn:
@iwasakims Livy also seems to have init-functions in its requirements. Should it be fixed too?

iwasakims:
> Livy also seems to have init-functions in its requirements. Should it be fixed too?

Yes. I updated the PR. Thanks.

sekikn:
+1, tested against ZK on Fedora 40. Thanks @iwasakims.

iwasakims:

laixiong:
Hi, thanks for the patch!

I noticed that this PR replaces the hardcoded path dependency on /lib/lsb/init-functions with a conditional requirement on redhat-lsb-core for Fedora ≥ 40 and EL9 systems.

However, on Rocky Linux 9 (and generally RHEL 9-based systems), the redhat-lsb-core package is no longer available in the official repositories. As a result, even though the RPM no longer directly depends on the path /lib/lsb/init-functions, it may still fail to install due to the missing package dependency.

Could you please confirm whether, with this change, the resulting RPMs can now be installed successfully on Rocky Linux 9 without additional manual steps (like creating symlinks or building dummy packages)? Or would it still require further compatibility handling?

Thanks again for your work on this!

iwasakims:
@laixiong redhat-lsb-core is provided by `devel` repository which is disabled by default on Rocky Linux 9. You need to enable it by `dnf config-manager --set-enabled devel` (or `dnf install --enablerepo=devel` on-the-fly).
https://github.com/apache/bigtop/blob/master/bigtop-deploy/puppet/manifests/yumrepo.pp#L40-L43

PR #1347, opened 2025-05-10
BIGTOP-4420. Fix packaging failure of bigtop-select using container due to undefined Gradle tasks.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4420

```
[bigtop-select] $ /bin/bash -ex /tmp/jenkins1810671388621150548.sh
+ JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-arm64
+ ./gradlew -POS=rockylinux-9 -Pprefix=3.4.0 -Pdocker-run-option=--privileged -Pmvn-cache-volume=true bigtop-select-clean bigtop-select-pkg-ind
Starting a Gradle Daemon, 3 busy Daemons could not be reused, use --status for details

* What went wrong:
Task 'bigtop-select-clean' not found in root project 'bigtop'.
```

`bigtop-select-*` tasks are not defined if the host OS is Debian family after [BIGTOP-4420](https://issues.apache.org/jira/browse/BIGTOP-4420). The intent was to avoid failure `./gradlew pkgs`. bigtop-select-pkg-ind and bigtop-select-clean should be defined for RPM packaging on Debian family host using container. I updated to exclude only bigtop-select-[s]pkg on Debian family host here.

### DIFF
--- a/packages.gradle
+++ b/packages.gradle
@@ -221,7 +221,7 @@ task "all-components" (description: "List the components of the stack") doLast {
   }
 }
 
-def genTasks = { target ->
+def genTasks = { target, packaging ->
   Task t = task "${target}-download" (dependsOn: "${target}_vardefines",
       description: "Download $target artifacts",
       group: PACKAGES_GROUP) doLast {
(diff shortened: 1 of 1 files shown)


sekikn:
+1, tested that `./gradlew -POS=rockylinux-9 -Pdocker-run-option=--privileged -Pmvn-cache-volume=true bigtop-select-clean bigtop-select-pkg-ind` worked on the Debian 12 host after building the Docker image for Rocky 9 and applying the PR. Thanks @iwasakims.

iwasakims:

sekikn:
@iwasakims after merging this, `./gradlew pkgs` doesn't seem to work. Would you take a look into it?

iwasakims:
@sekikn I added [addendum commit](https://github.com/apache/bigtop/commit/a2fec82dd73b0e47cdc5c545873457af74670f71). Could you see if this fixed the issue?

PR #1348, opened 2025-05-11
BIGTOP-4421. Fix build failure of Hadoop and Alluxio due to node-gyp error.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4421

We need python2 for rebuilding native libraries required for old node-spss by node-gyp on the architecture for which pre-built native libraries are not provided. Since recent OS distributions do not provide python2, we need another solution here.

```
[INFO] gyp ERR! configure error
[INFO] gyp ERR! stack Error: Can't find Python executable "python", you can set the PYTHON env variable.
[INFO] gyp ERR! stack     at PythonFinder.failNoPython (/bigtop-home/build/hadoop/rpm/BUILD/hadoop-3.3.6-src/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/target/webapp/node_modules/node-sass/node_modules/node-gyp/lib/configure.js:484:19)
[INFO] gyp ERR! stack     at PythonFinder.<anonymous> (/bigtop-home/build/hadoop/rpm/BUILD/hadoop-3.3.6-src/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/target/webapp/node_modules/node-sass/node_modules/node-gyp/lib/configure.js:406:16)
[INFO] gyp ERR! stack     at F (/bigtop-home/build/hadoop/rpm/BUILD/hadoop-3.3.6-src/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/target/webapp/node_modules/which/which.js:68:16)
[INFO] gyp ERR! stack     at E (/bigtop-home/build/hadoop/rpm/BUILD/hadoop-3.3.6-src/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/target/webapp/node_modules/which/which.js:80:29)
[INFO] gyp ERR! stack     at /bigtop-home/build/hadoop/rpm/BUILD/hadoop-3.3.6-src/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/target/webapp/node_modules/which/which.js:89:16
[INFO] gyp ERR! stack     at /bigtop-home/build/hadoop/rpm/BUILD/hadoop-3.3.6-src/hadoop-yarn-project/hadoop-yarn/hadoop-yarn-ui/target/webapp/node_modules/isexe/index.js:42:5
[...]

PR #1349, opened 2025-05-12
Update do-component-build

MacChen01:
Fix the alluxio compile failed bug.

openEuler use python3 default, node-gyp support python3 from 6.0.0.

### For code changes:


sekikn:
@MacChen01 Thank you for the PR. Is #1348 enough to resolve the compile issue, or should we still merge this PR [for the consistency between Hadoop](https://github.com/apache/bigtop/blob/rel/3.3.0/bigtop-packages/src/common/hadoop/do-component-build#L78-L80)? Either is fine with me.

iwasakims:
All OS distributions are affected on aarch64 (and ppc64le).

iwasakims:
> All OS distributions are affected on aarch64 (and ppc64le).

We should [add patch files (patch*.diff)](https://github.com/apache/bigtop/tree/master/bigtop-packages/src/common/alluxio) in this case. While the patch is applied on x86_64 too, it should be good for consistency.

PR #1366, opened 2025-07-22
BIGTOP-4471. Upgrade the Puppet manifest for Airflow to adopt PostgreSQL and LocalExecutor.

sekikn:
### Description of PR

This PR upgrades Airflow's Puppet manifest to use PostgreSQL as its backend and enable LocalExecutor, instead of SQLite and SequentialExecutor respectively, which are not recommended for production.

Ran smoke test against Debian 12, Fedora 40, Rocky 9 and Ubuntu 24.04 on x86_64.

### For code changes:


iwasakims:
@sekikn Could you rebase the PR?

iwasakims (approved):
LGTM. smoke-tests passed on openeuler-22.03-aarch64 too. Thanks, @sekikn.

PR #1377, opened 2026-02-05
BIGTOP-4516. Upgrade Tez to 0.10.5.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4516

smoke-tests of tez passed on Ubuntu 24.04 x86_64 and Rocky Linux 9 aarch64.


sekikn:
+1, tested with Debian 12, Rocky 9 and Ubuntu 24.04. Thanks @iwasakims.

iwasakims:

PR #1381, opened 2026-03-03
BIGTOP-4519. Upgrade Groovy to 3.0.25

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4519

We need to upgrade Groovy to address JDK upgrade in the stack. Since Hadoop still requires JDK 8 for release build, Groovy 3.0.25 can be good candidate since it supports all of Java 8, 11 and 17. Since Hadoop 3.4 required JDK 8 for release build, we can not drop the OpenJDK 8 now.

following commands used for release succeeded with the patch.

```
$ mvn clean install
$ mvn clean install -f bigtop-test-framework/pom.xml
$ mvn clean install -f bigtop-tests/test-artifacts/pom.xml
$ HADOOP_CONF_DIR=/etc/hadoop/conf HADOOP_HOME=/usr/lib/hadoop mvn clean install -DskipITs -f bigtop-tests/test-execution/pom.xml
```

smoke-tets of hdfs, yarn and mapreduce passed with the upgraded bigtop-groovy package on Ubuntu 24.04 x86_64.

### DIFF
--- a/bigtop-tests/smoke-tests/build.gradle
+++ b/bigtop-tests/smoke-tests/build.gradle
@@ -27,7 +27,7 @@ subprojects {
     }
   }
 
-  ext.groovyVersion = '2.5.4'
+  ext.groovyVersion = '3.0.25'
   ext.hadoopVersion = '2.7.4'
   ext.hbaseVersion = '1.1.9'
   ext.solrVersion = '8.11.4'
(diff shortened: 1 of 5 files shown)


sekikn:
+1, tested with Debian 12 and Ubuntu 24.04 on x86_64. Thanks @iwasakims.

iwasakims:

PR #1382, opened 2026-03-09
BIGTOP-4520. Upgrade Hadoop to 3.4.3.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4520

Hadoop 3.4.3 was released. Since it contains patches required for us to fix problems around building, we can remove patches applied on-the-fly on packaging 3.4.2.


iwasakims:
smoke-tests of hdfs, yarn and mapreduce passed on ubuntu-24.04-x86_64 and rockylinux-9-aarch64.

sekikn:
+1, tested with Debian 12, Rocky 9 and Ubuntu 24.04. Thanks @iwasakims.

iwasakims:

PR #1384, opened 2026-04-20
BIGTOP-4523. Make container tagging consistent between Linux and Mac OS on ARM

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4523

On ARM, uname -m returns arm64 on Mac while we are assuming aarch64 which is the value on Linux distributions. We should get consistent container image name even if it is built on Mac OS.


sekikn:
+1, tested on x86_64 as a regression. Thanks @iwasakims.

iwasakims:
Thanks, @sekikn. I merged this to master and branch-3.6.

PR #1386, opened 2026-04-30
BIGTOP-4525. Upgrade Ranger to 2.8.0.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4525

We need to upgrade Ranger to 2.8.0 for compatibility with Hadoop 3.4.

smoke-tests of ranger passed on ubuntu-24.04 x86_64.


sekikn:
+1, tested on openEuler 22.03 (x86_64). Thanks @iwasakims.

iwasakims:
Thanks, @sekikn. I merged this to master and branch-3.6.

PR #1392, opened 2026-06-03
BIGTOP-4531. Fix build failure of Airflow due to compatibility of google-re2.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4531

Building Airflow fails on rockylinux-9 and openeuler-22.03. The cause is version conflict between google-re2 and Python.

```
Collecting google-re2>=1.0 (from apache-airflow==2.10.5)
  Using cached google_re2-1.1.20251105.tar.gz (11 kB)
  ...(snip)
  building '_re2' extension
  ...(snip)
      g++ -Wno-unused-result -Wsign-compare -DDYNAMIC_ANNOTATIONS_ENABLED=1 -DNDEBUG -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -march=x86-64-v2 -mtune=generic -fasynchronous-unwind-ta\
bles -fstack-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -march=x86-64-v2 -mtune=generic -fasynchronous-unwind-tables -fstac\
k-clash-protection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fstack-protector-strong -m64 -march=x86-64-v2 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-pro\
tection -fcf-protection -D_GNU_SOURCE -fPIC -fwrapv -fPIC -I/bigtop-home/build/airflow/rpm/BUILD/apache_airflow-2.10.5/build/include -I/usr/include/python3.9 -c _re2.cc -o build/temp.linux-x86_64-cpython-39/_re2.o -fvisibility=hidden
      _re2.cc:15:10: fatal error: absl/strings/string_view.h: No such file or directory
         15 | #include "absl/strings/string_view.h"
[...]

PR #1393, opened 2026-06-03
BIGTOP-4533. Upgrade Airflow to 2.11.0.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4533

Since some providers require Airflow 2.11.0+ now, upgrading Airflow to 2.11.0 should resolve issues like [BIGTOP-4532](https://issues.apache.org/jira/browse/BIGTOP-4532).

We can not adopt Airflow 2.11.2 and 2.11.1 since [Python 3.9 support was dropped](https://airflow.apache.org/docs/apache-airflow/2.11.2/release_notes.html#python-3-9-support-removed). After upgrading the supported version of Rocky Linux and openEuler, we can upgrade Airflow further. The target version should be Airflow 3 rather than 2 in that case.

### DIFF
--- a/bigtop.bom
+++ b/bigtop.bom
@@ -349,7 +349,7 @@ bigtop {
     'airflow' {
       name    = "airflow"
       relNotes = "Apache Airflow"
-      version { base = '2.10.5'; pkg = base; release = 1 }
+      version { base = '2.11.0'; pkg = base; release = 1 }
       tarball { source      = "apache_airflow-${version.base}.tar.gz"
                 destination = source }
       url     { download_path = "/${name}/${version.base}/"


iwasakims:
smoke-tests of airflow passed on fedora-40-x86_64.

iwasakims:
We need the patch of BIGTOP-4531 (#1392) for rockylinux-9 and openeuler-22.03. The added [patch1-version-constraint-of-google-re2.diff‎](https://github.com/apache/bigtop/pull/1392/changes#diff-cd33f51ee1ffe27a086cdc9be7efe26519ccc914b0e814e60c06a90d00cd3f9f) is applicable even after upgrading AIrflow to 2.11.0.

sekikn:
+1, tested on Ubuntu 24.04 (x86_64) with #1392. Thanks @iwasakims.

iwasakims:
I merged this to branch-3.6 and master. Thanks, @sekikn.

PR #1396, opened 2026-06-14
BIGTOP-4535. Fix installation failure of R packages in toolchain.

iwasakims:
https://issues.apache.org/jira/browse/BIGTOP-4535

We need additional libraries and header files for installing R packages. We need to fix this for fixing build failure of Spark on openEuler 22.03 and Fedora 40.

Since the added packages are also available on Rocky Linux 9, I added them to manifests for Red Hat variants instead of Fedora specific conditional.

### DIFF
--- a/bigtop_toolchain/manifests/packages.pp
+++ b/bigtop_toolchain/manifests/packages.pp
@@ -68,6 +68,14 @@ class bigtop_toolchain::packages {
         "perl-Digest-SHA",
         "nasm",
         "yasm",
+        "harfbuzz-devel",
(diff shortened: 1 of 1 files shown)


sekikn:
+1. I made sure that `./gradlew spark-pkg-ind` was successful on Debian 12, Fedora 40, OpenEuler 22.03 and Rocky 9 (x86_64) with the docker image which was built with this PR. Thanks @iwasakims.

iwasakims:
I merged this to branch-3.6 and trunk. Thanks, @sekikn.

